PR #20, opened 2019-01-21
*: add safe mode test

amyangfei:
ref: https://internal.pingcap.net/jira/browse/TOOL-790

we use failpoint to simulate DM-worker exits in two scene:
* DM-worker exits when it is in sharding re-sync after sharding group synced
* DM-worker exits when waiting for sharding group synced
* test `safe mode` value for update binlog event, if the update binlog event is added during sharding DDL not synced, the `safe mode` should be true, otherwise the `safe mode` is false

### Check List 

Tests 

 - Integration test


csuzhangxc:
LGTM

GregoryIan:
I think we should check `safe mode` value in test case

amyangfei:
check `safe mode` value with the help of tracing result in test case
PTAL @GregoryIan @csuzhangxc

cmd/dm-tracer/main_test.go
```diff
@@ -0,0 +1,37 @@
...
+	"strings"
+	"testing"
+)
+
+func TestRunMain(t *testing.T) {
+	var args []string
+	for _, arg := range os.Args {
+		switch {
+		case arg == "DEVEL":
+		case strings.HasPrefix(arg, "-test."):
```
GregoryIan: I have never understood this logic, a `-test.` string prefix?

amyangfei: The parameter prefixed with -test is processed by go test and auto added coverage file

syncer/syncer.go
```diff
@@ -1639,7 +1644,12 @@ func (s *Syncer) Run(ctx context.Context) (err error) {
 					log.Info("[syncer] cancel to add DDL to job because of canceled from external")
 					return nil
 				}
+
 				if ddlExecItem.req.Exec {
+					failpoint.Inject("WaitShardingSyncExit", func() {
```
GregoryIan: better to rename it to `ShardSyncedExecutionExit`, 😄

cmd/dm-tracer/main_test.go
GregoryIan: ok I known

GregoryIan:
Well Done！ LGTM

tests/safe_mode/check_safe_mode.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+class OpType(object):
+    UPDATE = 2
+    DDL = 4
+
+
+class ReplicationEvent(object):
+    QUERY_EVENT = 2
+    UPDATE_ROWS_EVENTv0 = 21
```
GregoryIan: hard code is not a good habit

tests/safe_mode/check_safe_mode.py
```diff
@@ -0,0 +1,115 @@
...
+    UPDATE_ROWS_EVENTv1 = 24
+    UPDATE_ROWS_EVENTv2 = 31
+
+    UPDATE = [UPDATE_ROWS_EVENTv0, UPDATE_ROWS_EVENTv1, UPDATE_ROWS_EVENTv2]
+
+
+def scan_events():
+    url = 'http://127.0.0.1:8264/events/scan?limit=1000'
```
GregoryIan: is 1000 enough? should we not set limit argument?

tests/safe_mode/check_safe_mode.py
amyangfei: I have changed the test script using golang, which can reuse constant in dm pkg as much as possible. But still we have to hard code some unexported variables, such as job type defined in DM sync unit.

tests/safe_mode/check_safe_mode.py
amyangfei: added new scan iterator in test scripts

tests/_utils/check_port_offline
```diff
@@ -0,0 +1,24 @@
...
+
+port=$1
+retry=$2
+
+i=0
+while [ $i -lt $retry ]
+do
+    check_port 127.0.0.1 $port > /dev/null 2>&1
+    ret=$?
+    if [ "$ret" != 0 ]; then
```
csuzhangxc:
```suggestion
    if [ "$ret" != "0" ]; then
```
?

amyangfei: comparison here with or without quote works same, added to keep unified

tests/safe_mode/check_safe_mode.go
```diff
@@ -0,0 +1,176 @@
...
+			if pos.Compare(endPos) != 0 {
+				latestDDL[worker][0] = events[i]
+				break
+			}
+		}
+	}
+
+	// check safe mode of each `update` binlog event is reasonable
+	for worker, evs := range latestDDL {
+		startPos := binlogPosition(evs[0])
```
csuzhangxc: is `evs[0]` possible invalid if L153 not executed?

amyangfei: added ddl events validation

csuzhangxc:
LGTM

GregoryIan:
LGTM

PR #32, opened 2019-01-26
hide coveralls token

zhouqiang-cl:
Make coveralls token hide

hide token in makefile

### Check List 

Tests 

 - No code


GregoryIan:
LGTM

PR #35, opened 2019-01-27
*: fix ci name

amyangfei:
Currently we don't pass JobID of coverage test on coveralls.io, it is not easy for us to distinguish coverage report of each commit

rename ci service name to `jenkins`, pass `BUILD_NUMBER` from Jenkins (def BUILD_NUMBER = "${env.BUILD_NUMBER}")

### Check List 

Tests 

 - Integration test


amyangfei:
name `jenkins-ci` is ok, CI is already fixed

Issue #34, opened 2019-01-27
support generate MySQL/MariaDB binlog events for testing


csuzhangxc:
## Feature Request

**Describe the feature you'd like:**

Adding a pkg or some functions to generate MySQL/MariaDB binlog events. These events can be used for unit testing or integration testing.

**Teachability, Documentation, Adoption, Migration Strategy:**

- both MySQL and MariaDB need to be supported
- at least the following events type need to be supported
	- FORMAT_DESCRIPTION_EVENT
	- GTID_EVENT
	- QUERY_EVENT
	- TABLE_MAP_EVENT
	- WRITE_ROWS_EVENTv*, UPDATE_ROWS_EVENTv*, DELETE_ROWS_EVENTv*
	- XID_EVENT

PR #37, opened 2019-01-28
*: support generate MySQL/MariaDB binlog events for testing

csuzhangxc:

ref #34 

- add functions to support generate most important binlog events
- add a wrapper to support generate random binlog events/files

---

- 3 levels to generate binlog events (lowest to hightest)
	1. generate specified binlog events (`event.go`)
	2. generate binlog events for file header (`common.go`), DDL (`ddl.go`) and DML (`dml.go`)
	3. generate binlog events for file (`generator.go`)
- for the usage, see unit test cases

### Check List 

Tests 

 - Unit test


csuzhangxc:
@GregoryIan @amyangfei PTAL

GregoryIan:
Good Job！

pkg/binlog/event/common.go
```diff
@@ -0,0 +1,206 @@
...
+	)
+
+	switch flavor {
+	case gmysql.MySQLFlavor:
+		uuidSet := singleGTID.(*gmysql.UUIDSet)
+		interval := uuidSet.Intervals[0]
+		gtidEv, err = GenGTIDEvent(header, latestPos, defaultGTIDFlags, uuidSet.SID.String(), interval.Start, defaultLastCommitted, defaultSequenceNumber)
+	case gmysql.MariaDBFlavor:
+		mariaGTID := singleGTID.(*gmysql.MariadbGTID)
+		if mariaGTID.ServerID != header.ServerID {
```
GregoryIan: lol

pkg/binlog/event/common.go
```diff
@@ -0,0 +1,206 @@
...
+	case gmysql.MySQLFlavor:
+		mysqlGTIDs, ok := origin.(*gmysql.MysqlGTIDSet)
+		if !ok {
+			return nil, errors.NotValidf("GTID set string %s for MySQL", gSet)
+		}
+		if len(mysqlGTIDs.Sets) != 1 {
+			return nil, errors.Errorf("only one GTID in set is supported, but got %d (%s)", len(mysqlGTIDs.Sets), gSet)
+		}
+		var uuidSet *gmysql.UUIDSet
+		for _, uuidSet = range mysqlGTIDs.Sets {
```
GregoryIan: remove it？

pkg/binlog/event/common.go
csuzhangxc: for MariaDB, GTID including the `server-id`. :relieved:

pkg/binlog/event/event.go
```diff
@@ -0,0 +1,759 @@
...
+
+// GenEventHeader generates a EventHeader's raw data according to a passed-in EventHeader struct.
+// ref: https://dev.mysql.com/doc/internals/en/binlog-event-header.html
+func GenEventHeader(header *replication.EventHeader) ([]byte, error) {
+	buf := new(bytes.Buffer)
+
+	// timestamp, 4 bytes
+	err := binary.Write(buf, binary.LittleEndian, header.Timestamp)
+	if err != nil {
+		return nil, errors.Annotatef(err, "write timestamp %d", header.Timestamp)
```
GregoryIan: do you want add "[generate event header]" information in the error?

pkg/binlog/event/util.go
```diff
@@ -0,0 +1,270 @@
...
+	if err != nil {
+		return nil, errors.Annotatef(err, "decode % X", buf.Bytes())
+	}
+
+	return &replication.BinlogEvent{RawData: buf.Bytes(), Header: &header, Event: event}, nil
+}
+
+// combineHeaderPayload combines header, postHeader and payload together.
+func combineHeaderPayload(buf *bytes.Buffer, header, postHeader, payload []byte) error {
+	err := binary.Write(buf, binary.LittleEndian, header)
```
GregoryIan: do you want check `its length == eventHeaderLen`?

pkg/binlog/event/event.go
csuzhangxc: I tend to:

if `A` call `B`, `B` call `C`, and `C` return an error.
then
1. annotate `err` in `B` with something like `calling C fail`
2. annotate `err` in `A` with something like `calling B fail`

use this way, if `B` also call `D` and `D` return an error, then only need annotate `err` with something like `calling D` fail.

so, in this L66, I only annotate the error with information about L64, and annotate something like `generate event header` in the outer function.

pkg/binlog/event/common.go
```diff
@@ -0,0 +1,206 @@
...
+//   -. MariadbBinlogCheckPointEvent, not added yet
+func GenCommonFileHeader(flavor string, serverID uint32, gSet gtid.Set) ([]*replication.BinlogEvent, []byte, error) {
+	var (
+		header = &replication.EventHeader{
+			Timestamp: uint32(time.Now().Unix()),
+			ServerID:  serverID,
+			Flags:     defaultHeaderFlags,
+		}
+		latestPos       = uint32(len(replication.BinLogFileHeader))
+		prevGTIDsEv     *replication.BinlogEvent // for MySQL, this will be nil
```
GregoryIan: not implemented?

pkg/binlog/event/common.go
```diff
@@ -0,0 +1,206 @@
...
+			Flags:     defaultHeaderFlags,
+		}
+		gtidEv *replication.BinlogEvent
+	)
+
+	switch flavor {
+	case gmysql.MySQLFlavor:
+		uuidSet := singleGTID.(*gmysql.UUIDSet)
+		interval := uuidSet.Intervals[0]
+		gtidEv, err = GenGTIDEvent(header, latestPos, defaultGTIDFlags, uuidSet.SID.String(), interval.Start, defaultLastCommitted, defaultSequenceNumber)
```
GregoryIan: use `interval.Start`?

pkg/binlog/event/common.go
```diff
@@ -0,0 +1,206 @@
...
+// for MySQL:
+//   1. BinLogFileHeader, [ fe `bin` ]
+//   2. FormatDescriptionEvent
+//   3. PreviousGTIDsEvent
+// for MariaDB:
+//   1. BinLogFileHeader, [ fe `bin` ]
+//   2. FormatDescriptionEvent
+//   3. MariadbGTIDListEvent
+//   -. MariadbBinlogCheckPointEvent, not added yet
+func GenCommonFileHeader(flavor string, serverID uint32, gSet gtid.Set) ([]*replication.BinlogEvent, []byte, error) {
```
GregoryIan: for `Genxxx` functions,  I suggest to add `latestPos` for all them

pkg/binlog/event/common.go
GregoryIan: ok, I know  but still confused first time

pkg/binlog/event/common.go
```diff
@@ -0,0 +1,206 @@
...
+	singleGTID, err := verifySingleGTID(flavor, gSet)
+	if err != nil {
+		return nil, errors.Annotatef(err, "verify single GTID in set")
+	}
+	clone := gSet.Clone()
+
+	switch flavor {
+	case gmysql.MySQLFlavor:
+		uuidSet := singleGTID.(*gmysql.UUIDSet)
+		uuidSet.Intervals[0].Start++
```
GregoryIan: clone `uuidSet`

pkg/binlog/event/common.go
```diff
@@ -0,0 +1,206 @@
...
+		uuidSet.Intervals[0].Stop++
+		gtidSet := new(gmysql.MysqlGTIDSet)
+		gtidSet.Sets = map[string]*gmysql.UUIDSet{uuidSet.SID.String(): uuidSet}
+		err = clone.Set(gtidSet)
+	case gmysql.MariaDBFlavor:
+		mariaGTID := singleGTID.(*gmysql.MariadbGTID)
+		mariaGTID.SequenceNumber++
```
GregoryIan: ditto

pkg/binlog/event/common.go
csuzhangxc: the header has no meaningful pos/latestPos
1. 4 bytes magic num
2. (123-4) bytes FormatDescriptionEvent (currently with CRC32 checksum enabled)
3. variable length PreviousGTIDsEvent/MariadbGTIDListEvent

pkg/binlog/event/common.go
csuzhangxc: already cloned in `gSet.Origin()` call in `verifySingleGTID`.

pkg/binlog/event/common.go
csuzhangxc: same above

pkg/binlog/event/common.go
GregoryIan: while I saw this code, I had to confirm the position mechanism, although it is correct, but not uniform

pkg/binlog/event/util.go
csuzhangxc: added

pkg/binlog/event/common.go
csuzhangxc: yeh, but if add a `latestPos`, then do not know which value should be passed.:worried:

GregoryIan:
Rest LGTM

pkg/binlog/event/event.go
GregoryIan: ok. Just be unified

pkg/binlog/event/common.go
GregoryIan: we would use like this way
```
ev = GenXX(latestPos)
updateLatePos(ev)
```
so it's not a problem, expect diving into it

but it's not a big problem

pkg/binlog/event/generator.go
```diff
@@ -0,0 +1,134 @@
...
+	if err != nil {
+		return nil, nil, errors.Trace(err)
+	}
+	g.updateLatestPosGTID(result.LatestPos, result.LatestGTID)
+	return result.Events, result.Data, nil
+}
+
+// GenDDLEvents generates binlog events for DDL statements.
+// events: [GTIDEvent, QueryEvent]
+func (g *Generator) GenDDLEvents(schema string, query string) ([]*replication.BinlogEvent, []byte, error) {
```
GregoryIan: I perfer it

csuzhangxc: yeh. In fact, only this is enough for DDL.

pkg/binlog/event/common.go
csuzhangxc: check latest GTID whether is one of the last GTID in previous GTIDs?

GregoryIan: yep

GregoryIan:
LGTM

pkg/binlog/event/event.go
```diff
@@ -0,0 +1,759 @@
...
+	eventHeaderLen         = uint8(replication.EventHeaderSize) // always 19
+	crc32Len        uint32 = 4                                  // CRC32-length
+	tableMapFlags   uint16 = 1                                  // flags in TableMapEvent's post-header, not used yet
+)
+
+var (
+	// A array indexed by `Binlog-Event-Type - 1` to extract the length of the event specific header.
+	// It is copied from a binlog file generated by MySQL 5.7.22-log.
+	// The doc at https://dev.mysql.com/doc/internals/en/format-description-event.html does not include all of them.
+	eventTypeHeaderLen = []byte{0x38, 0x0d, 0x00, 0x08, 0x00, 0x12, 0x00, 0x04, 0x04, 0x04, 0x04, 0x12, 0x00, 0x00, 0x5f, 0x00, 0x04, 0x1a, 0x08, 0x00, 0x00, 0x00, 0x08, 0x08, 0x08, 0x02, 0x00, 0x00, 0x00, 0x0a, 0x0a, 0x0a, 0x2a, 0x2a, 0x00, 0x12, 0x34, 0x00}
```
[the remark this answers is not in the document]
csuzhangxc: split to more lines (every line with 16 bytes)

pkg/binlog/event/util.go
```diff
@@ -0,0 +1,274 @@
...
+	return int(columnCount+7) / 8
+}
+
+// nullBytes returns a n-length null bytes slice
+func nullBytes(n int) []byte {
+	buf := new(bytes.Buffer)
+	for i := 0; i < n; i++ {
+		buf.WriteByte(0x00)
+	}
+	return buf.Bytes()
```
amyangfei: we can get a n-length null bytes slice by `make([]byte, n)` directly?

pkg/binlog/event/common.go
```diff
@@ -0,0 +1,206 @@
...
+	case gmysql.MariaDBFlavor:
+		prevGTIDsEv, err = GenMariaDBGTIDListEvent(header, latestPos, gSet)
+		if err == nil {
+			prevGTIDsEvData = prevGTIDsEv.RawData
+		}
+	default:
+		return nil, nil, errors.NotSupportedf("flavor %s", flavor)
+	}
+	if err != nil {
+		return nil, nil, errors.Annotatef(err, "generate PreviousGTIDsEvent/MariadbGTIDListEvent")
```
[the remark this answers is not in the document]
csuzhangxc: ❤️ addressed.

pkg/binlog/event/util.go
csuzhangxc: yeh, we can

PR #42, opened 2019-02-12
*: support generated column in loader and binlog replication

amyangfei:
support synchronization data with generated column

1. auto filter generated column when generating DDL or DML
2. auto add column name list when generating insert SQL with generated column in loader

### Check List 

Tests 

 - Unit test
 - Integration test

need to upgrade test MySQL to 5.7

Side effects

 - Increased code complexity

Related changes


amyangfei:
PTAL @GregoryIan @csuzhangxc

loader/convert_data.go
```diff
@@ -56,6 +67,10 @@ func parseInsertStmt(sql []byte, table *tableInfo, columnMapping *cm.Mapping) ([
 			if sql[e] == '\n' && (sql[e-1] == ',' || sql[e-1] == ';') && sql[e-2] == ')' {
 				break
 			}
+			if sql[e] == '\n' && e-6 > s && bytes.Compare(sql[e-6:e], VALUES) == 0 {
```
csuzhangxc: we have `e = s + 3` in L66, so we can use `sql[e-2]` in L67.
is it safe to use `sql[e-6:e]`?

amyangfei: we have `e-6 > s` in this if condition, so it is safe to use `sql[e-6:e]`

GregoryIan:
deal with the ci coverage decline @amyangfei

amyangfei:
> deal with the ci coverage decline @amyangfei

I find the instability of coverage rate relates to some cleanup staff, for example we can compare difference between https://coveralls.io/builds/21662660/source?filename=syncer/db.go#L280 and https://coveralls.io/builds/21662523/source?filename=syncer/db.go#L280 .
All new code in this commit are tested in integration test or unit test. I will make another pr to try to make coverage more stable

loader/convert_data.go
```diff
@@ -236,10 +251,25 @@ func parseTable(r *router.Table, schema, table, file string) (*tableInfo, error)
...
 	}
 
-	columns := make([]string, 0, len(ct.Cols))
+	var (
+		columns          = make([]string, 0, len(ct.Cols))
+		hasGeneragedCols = false
+		columnNameFields = ""
+	)
 	for _, col := range ct.Cols {
+		if len(col.Options) > 0 && col.Options[0].Tp == ast.ColumnOptionGenerated {
```
csuzhangxc: `ColumnOptionGenerated` is always the first option in `col.Options`?

loader/convert_data_test.go
```diff
@@ -88,6 +88,50 @@ func (t *testConvertDataSuite) TestReassemble(c *C) {
...
+
+	sql := `INSERT INTO t1 (id,t_json) VALUES
+(10,'{}'),
+(9,NULL);
+(8,'{"a":123}');
+`
+
+	expected := []string{
+		"INSERT INTO t (`id`,`t_json`) VALUES(585520728116297738,'{}'),(585520728116297737,NULL),(585520728116297736,'{\"a\":123}');",
+		"INSERT INTO t (`id`,`t_json`) VALUES(10,'{}'),(9,NULL),(8,'{\"a\":123}');",
```
csuzhangxc: is this line needed?

loader/convert_data_test.go
```diff
@@ -88,6 +88,50 @@ func (t *testConvertDataSuite) TestReassemble(c *C) {
...
+		{
+			PatternSchema: "test*",
+			PatternTable:  "t*",
+			TargetColumn:  "id",
+			Expression:    cm.PartitionID,
+			Arguments:     []string{"1", "test", "t"},
+		},
+	}
+
+	for i, r := range rules {
```
GregoryIan: `rules` only contains one rule，but `expected` contains two case

loader/convert_data.go
amyangfei: changed to iterate all column options, which is more robust

syncer/ast.go
```diff
@@ -122,6 +122,14 @@ func columnOptionsToSQL(options []*ast.ColumnOption) string {
 			sql += fmt.Sprintf(" COMMENT '%s'", comment)
 		case ast.ColumnOptionOnUpdate: // For Timestamp and Datetime only.
 			sql += " ON UPDATE CURRENT_TIMESTAMP"
+		case ast.ColumnOptionGenerated:
```
GregoryIan: I would use restore to implement it later

syncer/db.go
```diff
@@ -390,6 +391,7 @@ func getTableColumns(db *Conn, table *table, maxRetry int) error {
 		column.idx = idx
 		column.name = string(data[0])
 		column.tp = string(data[1])
+		column.extra = string(data[5])
```
GregoryIan: update example？

syncer/ddl_test.go
```diff
@@ -338,3 +349,46 @@ func (s *testSyncerSuite) TestIgnoreDMLInQuery(c *C) {
 		c.Assert(pr.isDDL, Equals, cs.isDDL)
 	}
 }
+
+func (s *testSyncerSuite) TestResolveSQL(c *C) {
```
GregoryIan: is `TestResolveGeneratedColumnDDL` better?

syncer/dml.go
```diff
@@ -423,3 +423,43 @@ func (s *Syncer) mappingDML(schema, table string, columns []string, data [][]int
...
+		rows = append(rows, value)
+	}
+	return rows
+}
+
+// generatedColumnFilter iterates column list and returns
+// a bool indicates where one or more generated column exists
+// a bool slice indicates whether the i-th column is generated column
+// a new column slice without generated columns
+func generatedColumnFilter(columns []*column) (bool, []bool, []*column) {
```
GregoryIan: generate column would be one of unique columns？

syncer/syncer_test.go
```diff
@@ -135,6 +135,32 @@ func (s *testSyncerSuite) resetMaster() {
...
+			}
+			ch <- struct{}{}
+		}
+	}()
+
+	for {
+		t := time.NewTimer(10 * time.Millisecond)
+		select {
+		case <-ch:
+			t.Stop()
```
csuzhangxc: whether needing to call `cancel` here? and whether needing to *wait* the above goroutine to return?

syncer/syncer_test.go
```diff
@@ -135,6 +135,32 @@ func (s *testSyncerSuite) resetMaster() {
...
+			ev, _ := s.streamer.GetEvent(ctx)
+			if ev == nil {
+				return
+			}
+			ch <- struct{}{}
+		}
+	}()
+
+	for {
+		t := time.NewTimer(10 * time.Millisecond)
```
csuzhangxc: in `for` loop, can replace this with `time.After`?

syncer/dml.go
amyangfei: only secondary index is allowed on generated columns.

syncer/syncer_test.go
GregoryIan: is `10 millisecond` short?

amyangfei: 1. `BinlogStreamer` is reading relay log from background, which should be fast enough than this time interval.
2. This `s.catchUpBinlog` is commonly used when some tests meet error and leave some unread binlog out.
So I think 10ms is enough here

syncer/syncer_test.go
```diff
@@ -135,6 +135,31 @@ func (s *testSyncerSuite) resetMaster() {
...
+				return
+			}
+			ch <- struct{}{}
+		}
+	}()
+
+	for {
+		select {
+		case <-ch:
+			// do nothing
```
csuzhangxc: Can we remove the usage of `ch`?

csuzhangxc: needed, my fault.

syncer/dml.go
```diff
@@ -423,3 +423,64 @@ func (s *Syncer) mappingDML(schema, table string, columns []string, data [][]int
...
+		genColumnNames  = make(map[string]bool)
+	)
+
+	for _, c := range columns {
+		isGenColumn := c.isGeneratedColumn()
+		colIndexfilters = append(colIndexfilters, isGenColumn)
+		if isGenColumn {
+			needPrune = true
+			genColumnNames[c.name] = true
+			continue
```
csuzhangxc: useless.

syncer/dml.go
```diff
@@ -423,3 +423,64 @@ func (s *Syncer) mappingDML(schema, table string, columns []string, data [][]int
...
+	}
+
+	if !needPrune {
+		return columns, data, index
+	}
+
+	var (
+		cols  = make([]*column, 0, len(columns))
+		rows  = make([][]interface{}, 0, len(data))
+		idxes = make(map[string][]*column)
```
csuzhangxc: maybe we can cache `col` and `idxes` later? just like `syncer.cacheColumns`?

tests/sharding/data/db1.increment.sql
```diff
@@ -4,4 +4,19 @@ update t1 set name = 'Gabriel José de la Concordia García Márquez' where `uid
...
+insert into t2 (uid, name, info) values (20006, 'Buenos Aires', '{"age": 1000}');
+alter table t1 add column info_json json GENERATED ALWAYS AS (`info`) VIRTUAL;
+alter table t1 add column id_gen int as (uid + 1);
+alter table t2 add column id_gen int as (uid + 1);
+alter table t1 add unique (id_gen);
+alter table t2 add unique (id_gen);
+insert into t1 (uid, name, info) values (10005, 'Buenos Aires', '{"age": 100}');
+insert into t2 (uid, name, info) values (20007, 'Buenos Aires', '{"age": 200}');
+alter table t1 add key multi_col_idx(uid, id_gen);
+alter table t2 add key multi_col_idx(uid, id_gen);
```
csuzhangxc: Can it work if no primary key, but only multi_col_idx including generated column?

amyangfei:
PTAL @GregoryIan @csuzhangxc

syncer/dml.go
```diff
@@ -423,3 +484,94 @@ func (s *Syncer) mappingDML(schema, table string, columns []string, data [][]int
...
+		cacheKey    = dbutil.TableName(schema, table)
+		cacheStatus = cache.status(cacheKey)
+	)
+
+	if cacheStatus == noGenColumn {
+		return columns, data, index, nil
+	}
+	if cacheStatus == hasGenColumn {
+		rows := make([][]interface{}, 0, len(data))
+		filters, ok := cache.isGenColumn[cacheKey]
```
csuzhangxc: also test `cache.columns[cacheKey]` and `rows, cache.indexes[cacheKey]` here? like `ok1`, `ok2`, `ok3`.
but this maybe not a problem.

syncer/db.go
```diff
@@ -456,3 +459,7 @@ func getBinaryLogs(db *sql.DB) ([]binlogSize, error) {
 	}
 	return files, nil
 }
+
+func (c *column) isGeneratedColumn() bool {
+	return strings.Contains(c.extra, "VIRTUAL GENERATED") || strings.Contains(c.extra, "STORED GENERATED")
```
GregoryIan: must `extra`  be uppercase?

amyangfei: according to MySQL doc, `VIRTUAL GENERATED` and `STORED GENERATED` in `Extra` field are always uppercase

amyangfei: `strings.Contains` is used because there may exist other data in `Extra`?

GregoryIan: I'm afraid some comment options are in it

amyangfei: https://dev.mysql.com/doc/refman/5.7/en/show-columns.html
In MySQL doc, there are only four Extra fields: `auto_increment`, `on update`, `VIRTUAL GENERATED` or `VIRTUAL STORED`

GregoryIan:
Rest LGTM

GregoryIan:
LGTM

Issue #49, opened 2019-02-18
make coverage in CI more stable


amyangfei:
## Feature Request

**Is your feature request related to a problem? Please describe:**
Currently the coverage rate from the same code could be different, event have 2% decrease or increase. For example we have different coverage in 
```
https://coveralls.io/builds/21662660/source?filename=syncer/db.go#L280
https://coveralls.io/builds/21662523/source?filename=syncer/db.go#L280
```
```
https://coveralls.io/builds/21700081/source?filename=loader/db.go
https://coveralls.io/builds/21662523/source?filename=loader/db.go
```

**Describe the feature you'd like:**
We should get the same coverage rate after we run test case on the same commit/code.

PR #50, opened 2019-02-19
*: DM server graceful exit in integration test

amyangfei:
fix https://github.com/pingcap/dm/issues/49

* add a channel to control `DM-worker`/`DM-master` server close
* wait all test main processes exit in each integration test case before next one
* ~fix some race condition~ fix in https://github.com/pingcap/dm/pull/53

Aflter change the coverage rate of DM increases from 52% to 57%

### Check List 

Tests 

 - Integration test
   run CI multiple times and observe coverage rate

Code changes


amyangfei:
PTAL @GregoryIan @csuzhangxc

GregoryIan:
good catch！

GregoryIan:
LGTM

cmd/dm-master/main.go
```diff
@@ -71,3 +77,8 @@ func main() {
 
 	log.Info("dm-master exit")
 }
+
+func main() {
```
csuzhangxc: not used anymore?

amyangfei: main function is still the execution entrance of `dm-master`, `dm-worker`

cmd/dm-master/main_test.go
```diff
@@ -34,19 +41,20 @@ func TestRunMain(t *testing.T) {
...
 
 	signalCh := make(chan os.Signal, 1)
 	signal.Notify(signalCh, syscall.SIGINT, syscall.SIGQUIT, syscall.SIGTERM, syscall.SIGHUP)
 	select {
-	case <-signalCh:
-		return
 	case <-waitCh:
-		return
+	case <-signalCh:
+		close(exitCh)
```
csuzhangxc: our purpose is to control
1. exit the process when received signals
2. wait for the process exit
?

seems we call `signal.Notify` both in `main.go` and `main_test.go`. Is it needed?

tests/_utils/wait_process_exit
```diff
@@ -0,0 +1,16 @@
...
+while true
+do
+    pgrep $process > /dev/null 2>&1
+    ret=$?
+    if [ "$ret" != "0" ]; then
+        echo "process $process already exit"
+        break
+    fi
+    sleep 0.1
+    echo "wait process $process exit..."
```
csuzhangxc: `echo` every `0.1` second, too much log?

amyangfei: changed to `0.2s`, not a lot of logs, processes often exit within 1s in these integration test

cmd/dm-master/main_test.go
amyangfei: well the test starts `main` in a single goroutine, so it catches singal in `main_test.go` too, which is not necessary. I simplify the code by call `main()` directly

PR #54, opened 2019-02-20
use restore feature to refine codes releated ddl

GregoryIan:
use restore feature to refine codes releated ddl

use restore feature to implement rename and split ddl

### Check List 

Tests 

 - Unit test


csuzhangxc:
@GregoryIan resolve the conflict.

GregoryIan:
I would do it after #50 merged

GregoryIan:
@csuzhangxc @amyangfei PTAL

syncer/ddl_test.go
```diff
@@ -373,22 +349,25 @@ func (s *testSyncerSuite) TestResolveGeneratedColumnSQL(c *C) {
...
+		sqls, _, err := syncer.resolveDDLSQL(parser, ast1, "test")
 		c.Assert(err, IsNil)
 
 		c.Assert(len(sqls), Equals, 1)
 		getSQL := sqls[0]
 		c.Assert(getSQL, Equals, tc.expected)
 
-		ast2, err := parser.ParseOneStmt(getSQL, "", "")
-		c.Assert(err, IsNil)
+		// FIXME: remove me after implement generated function in restore
```
amyangfei: latest parser has fixed this bug, we can remove these comments now

test.go
```diff
@@ -0,0 +1,119 @@
+package main
```
amyangfei: is this file used for testing only, can remove now?

GregoryIan: yep

syncer/ddl_test.go
GregoryIan: it's useless, because of some very tricky code,  I decide remove these code

syncer/filter.go
```diff
@@ -128,7 +110,7 @@ func (s *Syncer) skipQuery(tables []*filter.Table, stmt ast.StmtNode, sql string
 
 	if len(tables) > 0 {
 		tbs := s.bwList.ApplyOn(tables)
-		if len(tbs) == 0 {
+		if len(tbs) != len(tables) {
```
amyangfei: suppose we rename a table from name `old` to name `new`, and table `old` is in `do-tables` and `new` is new `ignore-tables`.
should we filter this rename DDL?

GregoryIan: yep

GregoryIan:
@csuzhangxc @july2993 PTAL

pkg/parser/common.go
```diff
@@ -0,0 +1,309 @@
...
+	return bf.String(), nil
+}
+
+// SplitDDL splits multiple operations in one DDL statement into multiple DDL statements
+// if fail to restore, it would not restore the value of `stmt` (it changes it's values if `stmt` is one of  DropTableStmt, RenameTableStmt, AlterTableStmt)
+func SplitDDL(stmt ast.StmtNode, schema string) (sqls []string, err error) {
+	var (
+		b          []byte
+		schemaName = model.NewCIStr(schema) // fill schema name
+		bf         = bytes.NewBuffer(b)
```
csuzhangxc: `bf         buffer.Buffer` or `bf         = new(bytes.Buffer)` is clearer?

pkg/parser/common.go
```diff
@@ -0,0 +1,309 @@
...
+		for _, t := range tables {
+			if t.Schema.O == "" {
+				t.Schema = schemaName
+			}
+
+			v.Tables = []*ast.TableName{t}
+			bf.Reset()
+			err = stmt.Restore(ctx)
+			if err != nil {
+				return nil, errors.Annotate(err, "restore ast node")
```
csuzhangxc: is it needed to `v.Tables = tables` before this `return`? similar work for other cases.

pkg/parser/common.go
```diff
@@ -0,0 +1,309 @@
...
+			}
+			sqls = append(sqls, bf.String())
+
+			if spec.Tp == ast.AlterTableRenameTable {
+				v.Table = spec.NewTable
+			}
+		}
+		v.Specs = specs
```
csuzhangxc: whether needing to reset `v.Table` here?

pkg/parser/common.go
```diff
@@ -0,0 +1,309 @@
...
+
+		return sqls, nil
+
+	default:
+		return nil, errors.Errorf("unkown type ddl %+v", stmt)
```
csuzhangxc: s/unkown/unknown

pkg/parser/common.go
GregoryIan: according to comment in L179: if fail to restore, it would not restore the value of `stmt`.  but we can do it, let me see

PR #57, opened 2019-02-25
syncer: do not skip all queries begin with '#'

csuzhangxc:

Previously, we skipped all queries begin with `#` as a comment.
But, a query starts with `#` may be a valid DDL or DML, like
```
# this is a comment
CREATE TABLE do_db.do_table (c1 INT)
```
So, we can not simply skip it.

- remove `#` from the `builtInSkipDDLs`
- let the caller decide whether needing to skip the query

### Check List 

Tests 

 - Unit test

---

In our current integration test framework, comments start with `#` line will be ignored by the *mysql* client, then no comments will stay in binlog events. So, add such a test case has no meaning.

### DIFF
--- a/syncer/ddl_test.go
+++ b/syncer/ddl_test.go
@@ -326,6 +326,34 @@ func (s *testSyncerSuite) TestIgnoreDMLInQuery(c *C) {
 			isDDL:    false,
 			hasError: true,
 		},
+		{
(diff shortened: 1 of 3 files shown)


GregoryIan:
LGTM

csuzhangxc:
@amyangfei PTAL

PR #59, opened 2019-02-26
fix typo

ddatsh:
fix typo


csuzhangxc:
@ddatsh Thanks for your contribution. Is this PR ready for review? If so, you can add a `status/PTAL` label.

kennytm:
(I don't think non-members can add labels themselves.)

csuzhangxc:
Sorry. They can't. :joy:

GregoryIan:
LGTM！thanks @ddatsh

loader/status.go
```diff
@@ -63,7 +63,7 @@ func (l *Loader) PrintStatus(ctx context.Context) {
 
 		finishedSize := l.finishedDataSize.Get()
 		totalSize := l.totalDataSize.Get()
-		log.Infof("[loader] finished_bytes = %d, total_bytes = GetAllRestoringFiles%d, progress = %s", finishedSize, totalSize, percent(finishedSize, totalSize))
+		log.Infof("[loader] finished_bytes = %d, total_bytes = GetAllRestoringFiles = %d, progress = %s", finishedSize, totalSize, percent(finishedSize, totalSize))
```
amyangfei:
```suggestion
		# get filesCount from l.db2Tables maybe
		log.Infof("[loader] finished_bytes = %d, total_bytes = %d, GetAllRestoringFiles = %d, progress = %s", finishedSize, totalSize, filesCount, percent(finishedSize, totalSize))
```
here we miss the format string for `total_bytes` and real count for `GetAllRestoringFiles`. Maybe we can get `GetAllRestoringFiles` from `l.db2Tables` or adding another counter.

loader/status.go
```diff
@@ -63,7 +63,8 @@ func (l *Loader) PrintStatus(ctx context.Context) {
 
 		finishedSize := l.finishedDataSize.Get()
 		totalSize := l.totalDataSize.Get()
-		log.Infof("[loader] finished_bytes = %d, total_bytes = GetAllRestoringFiles%d, progress = %s", finishedSize, totalSize, percent(finishedSize, totalSize))
+		filesCount := l.db2Tables
```
amyangfei: `l.db2Tables` is a map from db -> tables and table -> data files, we can't get AllRestoringFiles count directly.

We have two options to get this count:
1. count db, table, and data files via `l.db2Tables`
2. update this count during `l.prepare()`

GregoryIan: what do you think？ @ddatsh

ddatsh: first, we just print with  l.db2Tables only?

amyangfei: yep, `l.db2Tables` contains all db schema, table schema, table table files that we concern (concern here means we have filtered system schema and schema/tables in black list). Count files stored in `l.db2Tables` is enough

PR #60, opened 2019-02-26
syncer: fix generated column in where condition

amyangfei:

Currently we remove all generated column for DML, but generated column in where condition is valid.

- keep generated column in index searching
- keep generated column in where condition when generating `update`/`delete` DML

### Check List 

Tests 

 - Integration test

Side effects

 - Increased code complexity

### DIFF
--- a/syncer/dml.go
+++ b/syncer/dml.go
@@ -92,6 +92,14 @@ func (c *GenColCache) reset() {
 	c.isGenColumn = make(map[string][]bool)
 }
 
+func extractValueFromData(data []interface{}, columns []*column) []interface{} {
+	value := make([]interface{}, 0, len(data))
+	for i := range data {
+		value = append(value, castUnsigned(data[i], columns[i].unsigned, columns[i].tp))
+	}
+	return value
+}
+
 func genInsertSQLs(param *genDMLParam) ([]string, [][]string, [][]interface{}, error) {
 	var (
 		schema               = param.schema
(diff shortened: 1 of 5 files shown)


amyangfei:
PTAL @GregoryIan @csuzhangxc

syncer/dml.go
```diff
@@ -139,17 +140,21 @@ func genUpdateSQLs(schema string, table string, data [][]interface{}, columns []
...
+		for i := range oriOldData {
+			oriOldValues = append(oriOldValues, castUnsigned(oriOldData[i], originalColumns[i].unsigned, originalColumns[i].tp))
+		}
 
 		if len(defaultIndexColumns) == 0 {
-			defaultIndexColumns = getAvailableIndexColumn(indexColumns, oldValues)
+			defaultIndexColumns = getAvailableIndexColumn(originalIndexColumns, oriOldValues)
 		}
 
 		ks := genMultipleKeys(columns, oldValues, indexColumns)
```
GregoryIan: should we use `oriOldValues`?

GregoryIan:
Rest LGTM

amyangfei:
PTAL @GregoryIan

syncer/dml.go
```diff
@@ -74,24 +83,31 @@ func (c *GenColCache) clearTable(schema, table string) {
...
+		originalIndexColumns = param.originalIndexColumns
+	)
 	sqls := make([]string, 0, len(dataSeq))
 	keys := make([][]string, 0, len(dataSeq))
 	values := make([][]interface{}, 0, len(dataSeq))
 	columnList := genColumnList(columns)
 	columnPlaceholders := genColumnPlaceholders(len(columns))
-	for _, data := range dataSeq {
+	for dataIdx, data := range dataSeq {
 		if len(data) != len(columns) {
```
GregoryIan: also check `originalColumns ` and `originalDataSeq `?

syncer/dml.go
```diff
@@ -140,16 +177,30 @@ func genUpdateSQLs(schema string, table string, data [][]interface{}, columns []
 			changedValues = append(changedValues, castUnsigned(changedData[i], columns[i].unsigned, columns[i].tp))
 		}
 
+		oriOldValues := make([]interface{}, 0, len(oriOldData))
```
GregoryIan: `genUpdateSQLs`, `genInsertSQLs`` has these codes in same pattern:

```
...values
for i := range data {
    values = append(values,....)
}

if len(columns) == len(originalColumns) {
  originalVals = values
} else {
    for {....}
}
```

can we extract it a function?

syncer/dml.go
amyangfei: not needed, data/columns and origianlData/originalColumns have the same length distance ?

amyangfei: add column and data length check in generated column prune

GregoryIan:
LGTM

PR #61, opened 2019-02-26
syncer: take the first 2 fields of show binary logs

ddatsh:
for to mysql 8 compatibility

mysql 8 , show binary logs,One more field

Log_name,File_size,Encrypted

in order to avoid

```
sql: expected 3 destination arguments in Scan, not 2
```

### DIFF
--- a/syncer/db.go
+++ b/syncer/db.go
@@ -444,11 +444,20 @@ func getBinaryLogs(db *sql.DB) ([]binlogSize, error) {
 	}
 	defer rows.Close()
 
+	rowColumns, err := rows.Columns()
+	if err != nil {
+		return nil, errors.Trace(err)
+	}
 	files := make([]binlogSize, 0, 10)
 	for rows.Next() {
 		var file string
 		var pos int64
-		err = rows.Scan(&file, &pos)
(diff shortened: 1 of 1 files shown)


kennytm:
Thanks for the contribution!

It is simpler to just check `len(rows.Columns())` and call `rows.Scan(&files, &pos, &dummy)` if the length is 3. This is in fact how we resolved the compatibility issue with the `SHOW MASTER STATUS` query

https://github.com/pingcap/dm/blob/3479b737986e7ab3345da93161bea53c50d1f302/pkg/utils/db.go#L68-L72

ddatsh:
@kennytm

GregoryIan:
LGTM

syncer/db.go
```diff
@@ -444,11 +444,17 @@ func getBinaryLogs(db *sql.DB) ([]binlogSize, error) {
 	}
 	defer rows.Close()
 
+	rowColumns, err := rows.Columns()
```
amyangfei: better adding a check to this `err`

kennytm:
LGTM after the review comment above is addressed.

PR #62, opened 2019-02-27
Retry sql if commit error says it's retryable

sunxiaoguang:
Signed-off-by: Xiaoguang Sun <sunxiaoguang@zhihu.com>

Load task may fail and requires manual intervention. This PR try to restart query if the error was retryable. 

Error message returned by TiDB has a marker [try again later] if the error is retryable. Keep running the sql until it's either succeeded or it can not be retried.

### DIFF
--- a/loader/db.go
+++ b/loader/db.go
@@ -16,6 +16,7 @@ package loader
 import (
 	"database/sql"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/go-sql-driver/mysql"
@@ -24,6 +25,8 @@ import (
 	"github.com/pingcap/errors"
 	tmysql "github.com/pingcap/parser/mysql"
 )
+
+const txnRetryableMark = "[try again later]"
 
 // Conn represents a live DB connection
 type Conn struct {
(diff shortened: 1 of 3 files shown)


loader/db.go
```diff
@@ -119,48 +122,54 @@ func (conn *Conn) executeSQLCustomRetry(sqls []string, enableRetry bool, isRetry
...
 			}
 		}
-	}
 
-	err = txn.Commit()
-	if err != nil {
-		return errors.Trace(err)
+		err = txn.Commit()
+		if err != nil {
+			if !strings.Contains(err.Error(), txnRetryableMark) {
```
WangXiangUSTC: how about add a function to judge whether this err is retryable?

sunxiaoguang:
> `executeSQLImp` is called in `executeSQLCustomRetry`, which already supports retry for most errors.
> Have you meet any case that loader fails without retry?

TiDB cluster load is pretty heavy while loading huge amount of data. It would eventually fail too many times and stop load task. This [try again later] mark is a clear indicator that retrying will eventually succeed and keep trying it is better than having it failed and let user to resume task.

loader/db.go
sunxiaoguang: Still loading data. So is there other place that should test this and retry indefinitely?

sunxiaoguang: Found there is similar code in syncer. Let me move it to pkg/utils/db.go so it can be reused.

sunxiaoguang: Done, and I found there is error code checking implementation in pkg/utils/db.go. So it's testing code instead of tricky pattern matching.

syncer/db.go
```diff
@@ -151,32 +152,38 @@ func (conn *Conn) executeSQLImp(sqls []string, args [][]interface{}) error {
...
 
-	txn, err := conn.db.Begin()
-	if err != nil {
-		log.Errorf("exec sqls[%v] begin failed %v", sqls, errors.ErrorStack(err))
-		return errors.Trace(err)
-	}
-
-	for i := range sqls {
-		log.Debugf("[exec][sql]%s[args]%v", sqls[i], args[i])
+	for {
```
amyangfei: `syncer` unit also has retry mechanism in `executeSQLJob`, and user can configure the max-retry count. So I think retry is not needed here.

sunxiaoguang: It' ok then. Let me close this pr

PR #63, opened 2019-02-28
*: fileSizeUpdated return value minor fix

amyangfei:
`fileSizeUpdated` returns the file size change status and error, but we have different error semantics, sometime it is unexpected execution error, sometime it is file size unexpected change error.

use the right return value

### Check List 

Tests 

 - Unit test
 - Integration test


amyangfei:
PTAL @GregoryIan @csuzhangxc

amyangfei:
> LGTM
> but if `err` is not `nil`, we did not use `cmp`, right?

yep

GregoryIan:
LGTM

PR #66, opened 2019-03-05
poor: resume task if sync unit exits with `invalid connection` error

GregoryIan:
try to resume task if sync unit exits with `invalid connection` error

it's a poor and very rough retry feature, the main reason is that the concurrency control of the sub task module is very confusing and needs to be optimized. After improve its state transition and concurrency control, I will optimize the implementation of retry feature.

### Check List 

Tests 
   - set up a dm cluster
   - start a task
   - interfere network to let connetion timeout
   - check whether dm-worker would try to resume task

### DIFF
--- a/dm/worker/subtask.go
+++ b/dm/worker/subtask.go
@@ -14,6 +14,7 @@
 package worker
 
 import (
+	"strings"
 	"sync"
 	"time"
 
@@ -166,6 +167,7 @@ func (st *SubTask) Run() {
 func (st *SubTask) fetchResult(ctx context.Context, cancel context.CancelFunc, pr chan pb.ProcessResult) {
 	defer st.wg.Done()
 
+retry:
 	select {
 	case <-ctx.Done():
 		return
@@ -177,7 +179,10 @@ func (st *SubTask) fetchResult(ctx context.Context, cancel context.CancelFunc, p
 			return // paused by external request
 		}
 
-		var stage pb.Stage
+		var (
+			cu    = st.CurrUnit()
+			stage pb.Stage
+		)
 		if len(result.Errors) == 0 {
 			if result.IsCanceled {
 				stage = pb.Stage_Stopped // canceled by user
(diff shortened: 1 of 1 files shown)


dm/worker/subtask.go
```diff
@@ -163,34 +164,50 @@ func (st *SubTask) Run() {
...
 		if len(result.Errors) == 0 {
 			if result.IsCanceled {
 				stage = pb.Stage_Stopped // canceled by user
 			} else {
 				stage = pb.Stage_Finished // process finished with no error
 			}
 		} else {
+			/* TODO
+			it's a poor and very rough retry feature, the main reason is that
+			the concurrency control of the sub task module is very confusing and needs to be optimized.
```
amyangfei: Is the `concurrency control optimization` including the following scenario?
all of DM-worker's connections to downstream TiDB suddenly reset by downstream(often caused by network issue), then we must recover each connection in syncer and if the worker-count config of syncer is N, we will wait (N + 1) * 10 seconds at most.

dm/worker/subtask.go
```diff
@@ -163,34 +164,50 @@ func (st *SubTask) Run() {
...
+			After improving its state transition and concurrency control,
+			I will optimize the implementation of retry feature.
+			*/
+			if st.retryErrors(result.Errors, cu) {
+				log.Warnf("[subtask] %s (%s) retry on error %v, waiting 10 second!", st.cfg.Name, cu.Type(), result.Errors)
```
amyangfei:
```suggestion
				log.Warnf("[subtask] %s (%s) retry on error %v, waiting 10 seconds!", st.cfg.Name, cu.Type(), result.Errors)
```

dm/worker/subtask.go
GregoryIan: first of all it is concurrent, I don't understand why we need to wait (N + 1) * 10 seconds at most.
For the refactoring, more is the reconstruction of the the structure, making the moudle and funcion more reasone, of course we can increase the backoff property

amyangfei: I mean for each subtask, `(N + 1) * 10 seconds` may be not continuous, but each DB connection must be recovered. So the worst case is `(N + 1) * 10 seconds` sums up.

GregoryIan: but it's real concurrent, right?

GregoryIan: it's the logic of unit, we can refine while refactoring  sync unit

dm/worker/subtask.go
```diff
@@ -621,3 +638,20 @@ func (st *SubTask) unitTransWaitCondition() error {
 	}
 	return nil
 }
+
+func (st *SubTask) retryErrors(errors []*pb.ProcessError, current unit.Unit) bool {
+	retry := true
+	switch current.Type() {
+	case pb.UnitType_Sync:
+		for _, err := range errors {
+			if strings.Contains(err.Msg, "invalid connection") {
```
csuzhangxc: should we need to close all DB connections and re-open them in sync unit?

dm/worker/subtask.go
csuzhangxc: we can refactor units one by one and then the outer state transition later.

csuzhangxc:
LGTM

PR #73, opened 2019-03-12
*: release resources acquired when Init fail in the half-way

csuzhangxc:

- release some resources (like DB connection) when `Init` fail

- record some functions to rollback `Init` operations
- call recorded functions if `Init` fail

### Check List 

Tests 

 	- modify the code to inject some errors in `Init`
 	- observe the log whether rollback functions are called

Side effects

 - Increased code complexity


csuzhangxc:
@GregoryIan @amyangfei PTAL

amyangfei:
after `NewRollbackHolder` is used, is the `needCloseUnits` logic (https://github.com/pingcap/dm/blob/master/dm/worker/subtask.go#L98-L119) still needed?

checker/checker.go
```diff
@@ -258,21 +267,27 @@ func (c *Checker) Close() {
...
+	c.closed.Set(true)
+}
+
+func (c *Checker) closeDBs() {
 	for _, instance := range c.instances {
 		if instance.sourceDB != nil {
 			if err := dbutil.CloseDB(instance.sourceDB); err != nil {
 				log.Errorf("close source db %+v error %v", instance.sourceDBinfo, err)
 			}
+			instance.sourceDB = nil
```
amyangfei: is it necessary to set to `nil`? the `subtask` will be always recycled if `Init` failed

dm/unit/rollback.go
```diff
@@ -0,0 +1,60 @@
...
+type RollbackFuncHolder struct {
+	mu   sync.Mutex
+	name string // used to make log clearer
+	fns  []RollbackFunc
+}
+
+// NewRollbackHolder creates a new RollbackFuncHolder instance.
+func NewRollbackHolder(name string) *RollbackFuncHolder {
+	return &RollbackFuncHolder{
+		name: name,
```
GregoryIan: is `owner` better?

dm/unit/rollback.go
```diff
@@ -0,0 +1,60 @@
...
+
+// Add adds a func to the holder.
+func (h *RollbackFuncHolder) Add(fn RollbackFunc) {
+	h.mu.Lock()
+	defer h.mu.Unlock()
+	h.fns = append(h.fns, fn)
+}
+
+// RollbackReverseOrder executes rollback functions in reverse order
+func (h *RollbackFuncHolder) RollbackReverseOrder() {
```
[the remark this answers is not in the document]
GregoryIan: add a simple test

dm/unit/unit.go
```diff
@@ -24,6 +24,8 @@ type Unit interface {
 	// Init initializes the dm process unit
 	// every unit does base initialization in `Init`, and this must pass before start running the sub task
 	// other setups can be done in `Process`, but this should be treated carefully, let it's compatible with Pause / Resume
+	// if Init successfully, the outer caller should call `Close` when the unit (or the task) finished, stopped or canceled (because other units Init fail).
```
GregoryIan:
```suggestion
	// if initialing successfully, the outer caller should call `Close` when the unit (or the task) finished, stopped or canceled (because other units Init fail).
```

GregoryIan:
is this a long-term solution? would we change it to initial unit only while we need to do it later?

checker/checker.go
csuzhangxc: not necessary. only want to do the pattern (release then reset) and not rely on other code.

csuzhangxc:
@GregoryIan @amyangfei PTAL.
I still chose to add a mechanism of `rollback`.

> would we change it to initial unit only while we need to do it later?

we can do this in `subtask` (outer of `Init`) later

GregoryIan:
LGTM

relay/relay.go
```diff
@@ -129,14 +130,22 @@ func NewRelay(cfg *Config) *Relay {
 }
 
 // Init implements the dm.Unit interface.
-func (r *Relay) Init() error {
+func (r *Relay) Init() (err error) {
```
amyangfei: Is it a little dangerous to use `err` in this way? For example if error happens in LINE#159, the rollback will not fire. We must ensure no `shadows declaration err` after `FuncRollback` is registered.

csuzhangxc: `return` will **set** this `err`, then `defer` can detect it.

amyangfei: well I see, it is the returned value

PR #74, opened 2019-03-12
loader: add db retry in checkpoint operation

amyangfei:
loader checkpoint init sometimes failed because of concurrent DDL

add retry to loader checkpoint db operation

### Check List 

Tests 

 - Unit test
 - Integration test


amyangfei:
PTAL @GregoryIan @csuzhangxc

amyangfei:
yep, currently db operation also can't be canceled from outside. I will propose an issue for it

GregoryIan:
LGTM

PR #75, opened 2019-03-12
*: replace golang.org/x/net/context with standard context

amyangfei:

`golang.org/x/net/context` is type alias with standard `context` since go 1.9. It is safe and convenient to replace it with standard library

replace `golang.org/x/net/context` with `context` and change some import sequence

### Check List 

Tests 

 - Unit test
 - Integration test

### DIFF
--- a/checker/checker.go
+++ b/checker/checker.go
@@ -15,6 +15,7 @@ package checker
 
 import (
 	"bytes"
+	"context"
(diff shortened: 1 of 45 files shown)


amyangfei:
PTAL @csuzhangxc  @july2993

GregoryIan:
LGTM

PR #76, opened 2019-03-13
*: fix go vet test in CI

amyangfei:

go module may be downloaded during `go vet` test and these download/extract information should be ignored in fail detect

remove stdout, check stderr only in `go vet` bash pipeline.
ref: https://stackoverflow.com/a/8535532/1115857

### Check List 

Tests 

 - Unit test
 - Integration test


GregoryIan:
LGTM

PR #78, opened 2019-03-13
*: fix go vet test

amyangfei:
go module download information writes to tty directly, so we use tee
to redirect stderr for fail detect

fix issue in https://github.com/pingcap/dm/pull/76.
because go module download information does not go to stdout/stderr directly, we always get these messages from bash pipeline

use /dev/stderr to filter real error messages.

### Check List 

Tests 

 - Unit test
 - Integration test


amyangfei:
PTAL @GregoryIan @csuzhangxc

GregoryIan:
LGTM

PR #79, opened 2019-03-13
test: add parameter to run single integration_test

amyangfei:

There will be more and more integration test cases, we should add an easy way for developer to run single integration test case.

add a parameter to integration_test script

### Check List 

Tests 

 - Integration test


amyangfei:
PTAL @GregoryIan @csuzhangxc

PR #81, opened 2019-03-13
test: fix vet test in CI

amyangfei:
fix go vet in CI.
unexpected error happened in JenkinsCI, however I can't reproduce this error in both my local environment and online pod in Jenkins. It may relate to different tty between CI system and other environments.

rethink the design, I found some unnecessary bash pipeline and redirect.
1. we can check all default vet options with `go vet` and let `go vet` complains if failed
2. we use a standalone test tool to detect `shadows declaration`, we can always return true in this check

so, make it simple, just run `go vet` and let go vet tool itself decides whether to exit.

Besides, in order to test CI works right, I update `tidb`, `tidb parser` and `go-mysql` to the latest version in `go.mod`. This is not harmful, and needed in some way.

### Check List 

Tests 

 - Unit test
 - Integration test


amyangfei:
PTAL @GregoryIan @csuzhangxc

Issue #77, opened 2019-03-13
task and subtask would be restarted automatically after we restart dm-worker


GregoryIan:
## Feature Request

**Is your feature request related to a problem? Please describe:**

I’m always frustrated when we must start task or subtask manually after we restart dm-worker. 
Essential cause is that both dm-master and dm-worker don’t store task/subtask configuration in a remote storage or local disk

**Describe the feature you'd like:**

After we restart dm-worker, task and subtask would be restarted automatically. 
DM should store task/subtask information in a remote storage or local disk

**Describe alternatives you've considered:**

Declare in advance that these solution is irrelevant about HA of dm-master/dm-worker processes

S1: store task configuration in dm-master
 -  need a remote storage
 -  need hearbeat mechanism between task and dm-master

S2: store subtask configuration in dm-worker’s disk
-  implement a local file task management component

if keeping dm-master simple,  we would like to choose S2, there are benefits:

- Don’t add a remote storage process in DM cluster
- Implementing HA of dm processes using a third  party program, like k8s operator, dm-worker only need to store relay log, meta, task configuration in to a NFS in this solution

csuzhangxc:
I think S2 is enough currently and easier to be implemented.

other things need to consider:

1. handle `stop-task` and `update-task`
2. maybe need a flag to disable this feature for debugging or other purposes?

GregoryIan:
yep, we must handle `stop-task` and `update-task`. But I don't quite understand this special debug purpose

csuzhangxc:
I haven’t thought very well about this debug purpose, maybe you can just ignore it.

PR #88, opened 2019-03-21
restore task after dm-workers restart

GregoryIan:
* refer issue #77 
* refine encrypted password, dm only decrypt password 
   * start a task
   * start relay/check unit
   * update worker config

store subtask in local disk of dm-worker. 

In all request `start/stop/update task`, dm-worker would change meta firstly, then handle the request,  we should explain it in document.

And I would modify ansible and add integrate test later

### Check List 

Tests 

 - Unit test
- integration test
 - Manual test
    - deploy a dm cluster
    - start a task
    - check dm worker meta file content
    - restart dm-worker
    - check whether task is running
    - add another task
    - check dm worker meta file content
    - stop one task
    - check dm worker meta file content
    - update task
    - check dm worker meta file content


checker/checker.go
```diff
@@ -118,7 +118,13 @@ func (c *Checker) Init() (err error) {
 			User:     instance.cfg.From.User,
 			Password: instance.cfg.From.Password,
 		}
-
+		if len(instance.sourceDBinfo.Password) > 0 {
+			pswd, err2 := utils.Decrypt(instance.sourceDBinfo.Password)
+			if err2 != nil {
+				return errors.Annotatef(err2, "can not decrypt password %s", instance.sourceDBinfo.Password)
```
csuzhangxc: maybe we need to do a trick for one of:

1. a new error rather than this `err` (returned by `utils.Decrypt`) 
2. modify `utils.Decrypt` itself 
3. modfiy `errors.Cause(err).Error()` in _dm/master/server.go_ L1721. :sweat:

currently, if check task fails when starting a task, the returned message to dmctl will be something like `illegal base64 data at input byte` because we returned the original error message, but maybe `invalid encrypted password` is better.

dm/worker/meta.go
```diff
@@ -0,0 +1,157 @@
...
+// Close closes meta DB
+func (metaDB *FileMetaDB) Close() error {
+	metaDB.lock.Lock()
+	defer metaDB.lock.Unlock()
+
+	return errors.Trace(metaDB.save())
+}
+
+func (metaDB *FileMetaDB) save() error {
+	serialized, err := metaDB.meta.Toml()
```
kennytm: Saving doesn't need `metaDB.lock.Lock()`?

checker/checker.go
kennytm: I think `utils.Decrypt` should at least be modified to return the annotated error, as this `"can not decrypt password"` message has been repeated too many times in this PR

dm/worker/config.go
```diff
@@ -211,25 +189,36 @@ func (c *Config) verify() error {
...
-		_, err := gtid.ParserGTID(c.Flavor, c.RelayBinlogGTID)
+		_, err = gtid.ParserGTID(c.Flavor, c.RelayBinlogGTID)
 		if err != nil {
 			return errors.Annotatef(err, "relay-binlog-gtid %s", c.RelayBinlogGTID)
 		}
 	}
-	return nil
+
+	_, err = c.DecryptPassword()
+	return err
```
amyangfei: better `return errors.Trace(err)`

dm/worker/config.go
```diff
@@ -211,25 +189,36 @@ func (c *Config) verify() error {
...
 
 // configFromFile loads config from file.
 func (c *Config) configFromFile(path string) error {
 	_, err := toml.DecodeFile(path, c)
-	return errors.Trace(err)
+	if err != nil {
+		return errors.Trace(err)
+	}
+
+	err = c.verify()
```
amyangfei: return `errors.Trace(c.verify)` for simplicity. btw we can upgrade `pingcap/errors` to `v0.11.1`, which has added nil error check before adding trace stack

dm/worker/config.go
```diff
@@ -246,25 +235,34 @@ func (c *Config) UpdateConfigFile(content string) error {
...
 	if err != nil {
 		return errors.Trace(err)
 	}
 
-	if len(c.From.Password) > 0 {
-		pswd, err = utils.Decrypt(c.From.Password)
+	return nil
+}
+
+// DecryptPassword return a decrypted config replica in config
```
amyangfei:
```suggestion
// DecryptPassword returns a decrypted config replica in config
```

dm/worker/meta.go
```diff
@@ -0,0 +1,157 @@
...
+	if err != nil {
+		return "", errors.Trace(err)
+	}
+	return b.String(), nil
+}
+
+// DecodeFile loads and decodes config from file
+func (m *Meta) DecodeFile(fpath string) error {
+	_, err := toml.DecodeFile(fpath, m)
+	if err != nil {
```
amyangfei: better `return errors.Trace(err)`

dm/worker/meta.go
```diff
@@ -0,0 +1,157 @@
...
+		return errors.Trace(err)
+	}
+
+	return nil
+}
+
+// Decode loads config from file data
+func (m *Meta) Decode(data string) error {
+	_, err := toml.Decode(data, m)
+	if err != nil {
```
amyangfei: ditto

dm/worker/meta.go
```diff
@@ -0,0 +1,157 @@
...
+}
+
+// FileMetaDB stores meta information in disk
+type FileMetaDB struct {
+	lock sync.Mutex // we need to ensure only a thread can access to `metaDB` at a time
+	meta *Meta
+	path string
+}
+
+// NewFileMetaDB return a meta file db
```
amyangfei:
```suggestion
// NewFileMetaDB returns a meta file db
```

dm/worker/meta.go
amyangfei: lock is hold in each external caller of `save`?

amyangfei: Lock is hold in each external caller of save, so we don't need acquire lock here.

kennytm: Hmm okay.

tests/_utils/restart_dm_worker
```diff
@@ -0,0 +1,21 @@
...
+workdir=$1
+port=$2
+conf=$3
+
+PWD=$(pwd)
+binary=$PWD/bin/dm-worker.test
+
+echo "[$(date)] <<<<<< START DM-WORKER on port $port, config: $conf >>>>>>"
+cd $workdir
+$binary -test.coverprofile="$TEST_DIR/cov.$TEST_NAME.worker.$port.out" DEVEL \
```
amyangfei: It seems this `restart_dm_worker` has quite similar logic to `run_dm_worker` script, is there any purpose to use this one? In addition, the content of coverage file `$TEST_DIR/cov.$TEST_NAME.worker.$port.out` will be truncated here, we'd better to use append stream to this file(`>>`) or choose another file name.

dm/worker/meta.go
GregoryIan: it's also a good pattern

checker/checker.go
GregoryIan: we can avoid to contain duplicate information

dm/worker/config.go
GregoryIan: I think it's also a good pattern rather than depending a special verison

dm/worker/config.go
```diff
@@ -246,25 +235,34 @@ func (c *Config) UpdateConfigFile(content string) error {
...
+func (c *Config) DecryptPassword() (*Config, error) {
+	// try decrypt password for To DB
```
csuzhangxc: why add this comment? in fact, it's **From** DB?

tests/_utils/restart_dm_worker
GregoryIan: remove `mkdir` and `ln` opeations, maybe we also add other operation in it later, how do you think about it?

amyangfei: `mkdir -p` is reentrant, and we can also update `ln` to `[ -f $workdir/bin/mydumper ] || ln -s $PWD/bin/mydumper $workdir/bin/mydumper` to make it reentrant

dm/worker/config.go
GregoryIan: remove it

dm/worker/meta.go
```diff
@@ -0,0 +1,157 @@
...
+	if err := os.MkdirAll(dir, 0700); err != nil {
+		return nil, errors.Annotatef(err, "create meta directory")
+	}
+
+	fd, err := os.Open(metaDB.path)
+	if os.IsNotExist(err) {
+		log.Warnf("failed to open meta file %s, going to create a new one: %v", metaDB.path, err)
+		return metaDB, nil
+	} else if err != nil {
+		return nil, errors.Trace(err)
```
csuzhangxc: `Annotatef` the `path`?

dm/worker/meta.go
```diff
@@ -0,0 +1,157 @@
...
+	}
+	defer fd.Close()
+
+	err = metaDB.meta.DecodeFile(metaDB.path)
+	if err != nil {
+		return metaDB, errors.Trace(err)
```
csuzhangxc: ditto?

dm/worker/meta.go
GregoryIan: failed to open, we would get  `&PathError{"open", name, e}` which contains path information

dm/worker/meta_test.go
```diff
@@ -0,0 +1,56 @@
...
+
+	meta := metaDB.Get()
+	c.Assert(meta.SubTasks, HasLen, 0)
+
+	err = metaDB.Set(&config.SubTaskConfig{
+		Name: "task1",
+	})
+	c.Assert(err, IsNil)
+
+	meta = metaDB.Get()
```
csuzhangxc: add a case for `Del`?

dm/worker/server.go
```diff
@@ -177,7 +185,12 @@ func (s *Server) OperateSubTask(ctx context.Context, req *pb.OperateSubTaskReque
 	var err error
 	switch req.Op {
 	case pb.TaskOp_Stop:
-		err = s.worker.StopSubTask(name)
+		if err = s.worker.meta.Del(name); err != nil {
+			log.Errorf("update task %s into meta: %v", name, errors.ErrorStack(err))
+			resp.Msg = fmt.Sprintf("update task %s into meta: %v", name, errors.ErrorStack(err))
```
csuzhangxc: you can reuse this message both for `log.Errorf` and `resp.Msg`.

dm/worker/server.go
```diff
@@ -211,6 +224,14 @@ func (s *Server) UpdateSubTask(ctx context.Context, req *pb.UpdateSubTaskRequest
 		}, nil
 	}
 
+	if err = s.worker.meta.Set(cfg); err != nil {
+		log.Errorf("[server] update task %s into meta: %v", cfg, errors.ErrorStack(err))
+		return &pb.CommonWorkerResponse{
+			Result: false,
+			Msg:    fmt.Sprintf("update task %s into meta: %v", cfg, errors.ErrorStack(err)),
```
csuzhangxc: ditto

go.mod
```diff
@@ -27,6 +27,7 @@ require (
 	github.com/spf13/cobra v0.0.3
 	github.com/spf13/pflag v1.0.3 // indirect
 	golang.org/x/crypto v0.0.0-20190103213133-ff983b9c42bc // indirect
+	golang.org/x/net v0.0.0-20190108225652-1e06a53dbb7e
```
csuzhangxc: `golang.org/x/net` already removed in later PR after you created the branch, you do not need to add it back.

dm/worker/config.go
```diff
@@ -77,6 +77,7 @@ type Config struct {
 	EnableGTID  bool   `toml:"enable-gtid" json:"enable-gtid"`
 	AutoFixGTID bool   `toml:"auto-fix-gtid" json:"auto-fix-gtid"`
 	RelayDir    string `toml:"relay-dir" json:"relay-dir"`
+	MetaDir     string `toml:"meta-dir" json:"meta-dir"`
```
amyangfei: Do we need one more option such as `EnableAutoRestore` to control whether to restore tasks after dm-worker restart?

GregoryIan: why need to control it?

amyangfei: For example one task causes DM-worker crash (root cause is DM's bug), and if user restarts the DM-worker, the task auto recovers and DM-worker crashes again.

GregoryIan: it shows that there'e something wrong with task verification , and it is really an exception handling scenario, we should handle it in other way, may not just ignore

amyangfei: I mean the task starts normally, runs for a while, and then meets an unexpected error. The scenario is extreme but once it happens, it is hard to handle. (Maybe user can change the content of dm_worker_meta file)

GregoryIan:
@csuzhangxc @amyangfei PTAL

dm/worker/config.go
amyangfei: We’d better unify whether to check `err ! = nil` before returning `errors.Trace(err)` ?
btw, go’s error system is still controversial, it is not a big problem here

dm/worker/server.go
GregoryIan: LOL, in fact I want to change it

dm/worker/config.go
GregoryIan: yep, we should implement it in dmctl

PR #89, opened 2019-03-22
loader: fix load progress

csuzhangxc:

if we pause/resume task when loading, the **progress** may show an incorrect value. It's because we forgot to reset `totalDataSize`/`finishedDataSize` when resuming.

reset `totalDataSize`/`finishedDataSize` when resuming.

### Check List 

Tests 

	 - `pause-task` when loading
 	 - `resume-task`
	 - observe `finishedBytes`/`totalBytes`/`progress` in `query-status`'s output


csuzhangxc:
> Besides adding test to some status function for each unit is work in progress.

cool!

csuzhangxc:
@kennytm PTAL

Issue #91, opened 2019-03-25
make relay unit more robust


csuzhangxc:
## Feature Request

**Is your feature request related to a problem? Please describe:**

Currently, we need the following assumptions to make relay unit workers correctly:

1. master always sends all binlog events in its binlog file
2. master sends an event until all of the previous events sent

But these assumptions may not always satisfied.

**Describe the feature you'd like:**

We can change the assumption to the following:

- **master always sends all binlog events a slave needed**

In other words, the correctness of the relay unit relies on **the sending behavior of the master**, not binlog file on the master.

In practice, we can use some dummy events to fill the relay log file if no corresponding events received from the master.

PR #92, opened 2019-03-25
relay: use Reader interface read binlog events

csuzhangxc:

This PR is a part work of #91 

Introduced a binlog event `Reader` interface, so we can implement at least 3 kinds of Reader later:

1. TCPReader: read binlog events from a master server through TCP stream
2. FileReader: read binlog events from binlog files (or relay log files)
3. InMemoryReader: read binlog events from memory, used as a mock master to do testing

In this PR, I only implement a `TCPReader` and refactor a litter code for relay writer to make review easier.

Later, the following things will be done:

1. split relay writer to _reader_ --> _translator_ --> _writer_
2. add an InMemoryReader to test 3 parts above
3. refactor sync unit (relay reader) to read through a FileReader and use an InMemoryReader to test it

- add a `TCPReader`
- refactor relay writer to use the `TCPReader`

### Check List 

Tests 

 - Unit test

Code changes

### DIFF
--- a/pkg/binlog/reader/tcp.go
+++ b/pkg/binlog/reader/tcp.go
@@ -31,9 +31,9 @@ import (
 
 // TCPReader is a binlog event reader which read binlog events from a TCP stream.
 type TCPReader struct {
-	stageMu sync.Mutex
-	stage   readerStage
+	mu sync.Mutex
 
+	stage     readerStage
 	syncerCfg replication.BinlogSyncerConfig
 	syncer    *replication.BinlogSyncer
 	streamer  *replication.BinlogStreamer
@@ -41,8 +41,8 @@ type TCPReader struct {
 
 // TCPReaderStatus represents the status of a TCPReader.
 type TCPReaderStatus struct {
-	Stage      string `json:"stage"`
-	Connection uint32 `json:"connection"`
+	Stage  string `json:"stage"`
+	ConnID uint32 `json:"connection"`
 }
 
 // String implements Stringer.String.
@@ -64,11 +64,11 @@ func NewTCPReader(syncerCfg replication.BinlogSyncerConfig) Reader {
 
 // StartSyncByPos implements Reader.StartSyncByPos.
 func (r *TCPReader) StartSyncByPos(pos gmysql.Position) error {
-	r.stageMu.Lock()
-	defer r.stageMu.Unlock()
+	r.mu.Lock()
+	defer r.mu.Unlock()
 
 	if r.stage != stageNew {
-		return errors.NotValidf("stage %d, expect %d", r.stage, stageNew)
+		return errors.Errorf("stage %d, expect %d", r.stage, stageNew)
 	}
 
 	streamer, err := r.syncer.StartSync(pos)
@@ -83,11 +83,11 @@ func (r *TCPReader) StartSyncByPos(pos gmysql.Position) error {
 
 // StartSyncByGTID implements Reader.StartSyncByGTID.
 func (r *TCPReader) StartSyncByGTID(gSet gtid.Set) error {
-	r.stageMu.Lock()
-	defer r.stageMu.Unlock()
+	r.mu.Lock()
+	defer r.mu.Unlock()
 
 	if r.stage != stageNew {
-		return errors.NotValidf("stage %d, expect %d", r.stage, stageNew)
+		return errors.Errorf("stage %d, expect %d", r.stage, stageNew)
 	}
 
 	if gSet == nil {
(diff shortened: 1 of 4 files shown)


csuzhangxc:
@GregoryIan @amyangfei PTAL
I want to open several PRs for #91, this one is a part of it.

pkg/binlog/reader/tcp.go
```diff
@@ -0,0 +1,163 @@
...
+
+	syncerCfg replication.BinlogSyncerConfig
+	syncer    *replication.BinlogSyncer
+	streamer  *replication.BinlogStreamer
+}
+
+// TCPReaderStatus represents the status of a TCPReader.
+type TCPReaderStatus struct {
+	Stage      string `json:"stage"`
+	Connection uint32 `json:"connection"`
```
GregoryIan: what's meaning of `Connection`?

pkg/binlog/reader/tcp.go
```diff
@@ -0,0 +1,163 @@
...
+	}
+}
+
+// StartSyncByPos implements Reader.StartSyncByPos.
+func (r *TCPReader) StartSyncByPos(pos gmysql.Position) error {
+	r.stageMu.Lock()
+	defer r.stageMu.Unlock()
+
+	if r.stage != stageNew {
+		return errors.NotValidf("stage %d, expect %d", r.stage, stageNew)
```
GregoryIan: is log `stage %d, expect %d is not valid`?

pkg/binlog/reader/tcp.go
```diff
@@ -0,0 +1,163 @@
...
+		return errors.NotValidf("already closed")
+	}
+
+	connID := r.syncer.LastConnectionID()
+	if connID > 0 {
+		dsn := fmt.Sprintf("%s:%s@tcp(%s:%d)/?charset=utf8mb4",
+			r.syncerCfg.User, r.syncerCfg.Password, r.syncerCfg.Host, r.syncerCfg.Port)
+		db, err := sql.Open("mysql", dsn)
+		if err != nil {
+			return errors.Annotate(err, "open connection to the master")
```
GregoryIan: I think we can output which master

pkg/binlog/reader/tcp.go
```diff
@@ -0,0 +1,163 @@
...
+		}
+		defer db.Close()
+		err = utils.KillConn(db, connID)
+		if err != nil {
+			return errors.Annotatef(err, "kill connection %d", connID)
```
GregoryIan: ditto

pkg/binlog/reader/tcp.go
```diff
@@ -0,0 +1,163 @@
...
+
+	r.stage = stageClosed
+	return nil
+}
+
+// GetEvent implements Reader.GetEvent.
+func (r *TCPReader) GetEvent(ctx context.Context, checkStage bool) (*replication.BinlogEvent, error) {
+	if checkStage {
+		r.stageMu.Lock()
+		if r.stage != stagePrepared {
```
GregoryIan: would atomic operation be better?

pkg/binlog/reader/tcp.go
```diff
@@ -0,0 +1,163 @@
...
+	}
+
+	r.streamer = streamer
+	r.stage = stagePrepared
+	return nil
+}
+
+// StartSyncByGTID implements Reader.StartSyncByGTID.
+func (r *TCPReader) StartSyncByGTID(gSet gtid.Set) error {
+	r.stageMu.Lock()
```
GregoryIan: in fact this lock prevent more than `stage` field

GregoryIan:
Rest LGTM. well done

pkg/binlog/reader/tcp.go
csuzhangxc: renamed to `ConnID`, but keep the json flag as `connection`.

pkg/binlog/reader/tcp.go
csuzhangxc: changed to `stage %s, expect %s` only.

pkg/binlog/reader/tcp.go
csuzhangxc: added

pkg/binlog/reader/tcp.go
csuzhangxc: added

pkg/binlog/reader/tcp.go
csuzhangxc: changed to use atomic operation, and removed `checkStage`.

pkg/binlog/reader/tcp.go
csuzhangxc: yes, rename to `mu`.

pkg/binlog/reader/tcp.go
```diff
@@ -0,0 +1,159 @@
...
+	}
+
+	return r.streamer.GetEvent(ctx)
+}
+
+// Status implements Reader.Status.
+func (r *TCPReader) Status() interface{} {
+	r.mu.Lock()
+	stage := r.stage
+	r.mu.Unlock()
```
amyangfei: why only lock for stage getter? besides I find some redundant lock protection, such as 
```
r.mu.Lock()
defer r.mu.Unlock()
r.stage.Get()
```
Is there any consideration to do that?

csuzhangxc: I change to use `stage = r.stage.Get` without lock.

We use `stage` both in comparing(L150) and recording(L154).
Saving `stage` in L148, so we can get the same value both in L150 and L154.

GregoryIan:
LGTM

PR #95, opened 2019-03-28
syncer: fix the error message about executing multiple DDLs

kennytm:

Previous it logged the `ddl` constant in `job.go` making the error message always showing "... and current ddl ddl for source table ..." which is useless. It is supposed to log the parameter `ddls`.

Replaced "ddl" with "ddls".

### Check List 

Tests 

 - Unit test
    * And mainly verified that `Printf("%q", []string{"a", "b"})` has the desired output.

Code changes

Side effects

Related changes


kennytm:
PTAL @amyangfei @GregoryIan

PR #99, opened 2019-04-02
*: change `max_allowed_packet` to 64MB

csuzhangxc:

the default value of `max_allowed_packet` in MySQL server and _go-sql-driver/mysql_ driver is 4MB, but in TiDB server it's 64MB.
ref:

- <https://dev.mysql.com/doc/refman/5.7/en/server-system-variables.html#sysvar_max_allowed_packet>
- <https://github.com/go-sql-driver/mysql#maxallowedpacket>

- change the default value of `max_allowed_packet` to 64MB
	- in fact, in the clinet this can be 1GB, ref <https://dev.mysql.com/doc/refman/5.7/en/server-system-variables.html#sysvar_max_allowed_packet>
- add a `max-allowed-packet` config item for DB config
- change `charset` from `utf8` to `utf8mb4`

we can refactor DB operation to a pkg and add test cases for config later.

### Check List 

Tests 

 	- config with/without `max_allowed_packet` in DM-worker's config and task's config
	- hack the code to see the DSN

Side effects

 - Increased code complexity

### DIFF
--- a/dm/config/subtask.go
+++ b/dm/config/subtask.go
@@ -47,12 +47,17 @@ const (
 	CmdSyncer CmdName = "syncer"
 )
 
+var (
+	defaultMaxAllowedPacket = 64 * 1024 * 1024 // 64MiB, equal to TiDB's default
+)
+
 // DBConfig is the DB configuration.
 type DBConfig struct {
-	Host     string `toml:"host" json:"host" yaml:"host"`
-	Port     int    `toml:"port" json:"port" yaml:"port"`
-	User     string `toml:"user" json:"user" yaml:"user"`
-	Password string `toml:"password" json:"-" yaml:"password"` // omit it for privacy
+	Host             string `toml:"host" json:"host" yaml:"host"`
+	Port             int    `toml:"port" json:"port" yaml:"port"`
+	User             string `toml:"user" json:"user" yaml:"user"`
+	Password         string `toml:"password" json:"-" yaml:"password"` // omit it for privacy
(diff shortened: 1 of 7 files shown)


GregoryIan:
LGTM

dm/config/subtask.go
```diff
@@ -47,12 +47,17 @@ const (
...
 type DBConfig struct {
-	Host     string `toml:"host" json:"host" yaml:"host"`
-	Port     int    `toml:"port" json:"port" yaml:"port"`
-	User     string `toml:"user" json:"user" yaml:"user"`
-	Password string `toml:"password" json:"-" yaml:"password"` // omit it for privacy
+	Host             string `toml:"host" json:"host" yaml:"host"`
+	Port             int    `toml:"port" json:"port" yaml:"port"`
+	User             string `toml:"user" json:"user" yaml:"user"`
+	Password         string `toml:"password" json:"-" yaml:"password"` // omit it for privacy
+	MaxAllowedPacket *int   `toml:"max-allowed-packet" json:"max-allowed-packet" yaml:"max-allowed-packet"`
```
amyangfei: why use a `* int` here?

csuzhangxc: differ `0` from _not set_(default)

amyangfei: so we allow user to set it to zero, which means no limit to max_allowed_packet ?

csuzhangxc: zero means 
> automatically fetch the max_allowed_packet variable from server on every connection.

from <https://github.com/go-sql-driver/mysql#maxallowedpacket>

amyangfei: ok

PR #100, opened 2019-04-03
*: add a `.name` suffix to avoid data overwrite between tasks

csuzhangxc:

when running multiple tasks, the user may specify the same `loader.dir` between tasks, then the dumped data maybe overwrote.

- append a `.name` prefix to the `loader.dir` if not exists

### Check List 

Tests 

	 - start a full/all task
	 - observe the dumped data dir name

### DIFF
--- a/dm/config/subtask.go
+++ b/dm/config/subtask.go
@@ -18,6 +18,7 @@ import (
 	"encoding/json"
 	"flag"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/BurntSushi/toml"
@@ -266,6 +267,12 @@ func (c *SubTaskConfig) adjust() error {
 		if err != nil {
 			return errors.Annotatef(err, "invalid timezone string: %s", c.Timezone)
 		}
+	}
+
+	dirSuffix := "." + c.Name
+	if !strings.HasSuffix(c.LoaderConfig.Dir, dirSuffix) { // check to support multiple times calling
+		// if not ends with the task name, we append the task name to the tail
+		c.LoaderConfig.Dir += dirSuffix
 	}
 
 	c.From.Adjust()
--- a/dm/master/task.yaml
+++ b/dm/master/task.yaml
@@ -69,7 +69,7 @@ mysql-instances:             # one or more source database, config more source d
 
     loader:                  # local loader rule
       pool-size: 16
-      dir: "./dumped_data"   # must be unique between tasks for the same instance
+      dir: "./dumped_data"   # if not ends with `.name`, then `.name` will be appended
 
     syncer:
       worker-count: 16
--- a/loader/loader.go
+++ b/loader/loader.go
@@ -772,7 +772,20 @@ func (l *Loader) prepare() error {
 
 	// check if mydumper dir data exists.
 	if !utils.IsDirExists(l.cfg.Dir) {
-		return errors.Errorf("%s is not exists or it's not a dir", l.cfg.Dir)
+		// compatibility with no `.name` suffix
+		dirSuffix := "." + l.cfg.Name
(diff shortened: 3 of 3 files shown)


csuzhangxc:
@GregoryIan @amyangfei PTAL

GregoryIan:
can we adjust in task config to avoid compatibility issue？

csuzhangxc:
> can we adjust in task config to avoid compatibility issue？

I consider this, the affected scope only when restarting task with checkpoint from _loader_. If we want to avoid this compatibility issue, we need to scan a backoff dir (without suffix), this may make the code a litter ugly, do you think should we do it?

csuzhangxc:
@GregoryIan compatibility issue fixed, PTAL again

GregoryIan:
if we update task, dm-worker doesn't update loader-dir. If there are already two task use same loader-dir, they  both  also find the existing dir (without suffix), right?

csuzhangxc:
> if we update task, dm-worker doesn't update loader-dir. If there are already two task use same loader-dir, they both also find the existing dir (without suffix), right?

yeh, this PR can't fix the same loader-dir already existing problem.

GregoryIan:
two question:
* if we adjust `loader-dir ` in task config, the case that we can't correct are only there are task whose stage is dump; but in current implementation,  new task would also meet problem, right?
* if we adjust `loader-dir`, are there problem about directory permissions？

csuzhangxc:
> if we adjust loader-dir in task config, the case that we can't correct are only there are task whose stage is dump; but in current implementation, new task would also meet problem, right?

I don't understand your question well, but:
1. in dump stage, resume/re-start are the same, it will always remove old data dir if exists and create a new one, then dump the data
	- if upgrade DM in this stage, it will just look like start a fresh task
	- in fact, if two tasks using the same data dir (with old DM version), an error will happen in this stage because the later one will remove the previous one's data
2. in load stage, DM only try to load the data already dumped, two tasks should not load data from the same dir (with or without `.name` suffix because the error has occurred in dump stage)
	- if upgrade DM in this stage, it will load data from the old (without suffix) data dir

> if we adjust loader-dir, are there problem about directory permissions？

DM need permissions for the parent directory of the data dir (not only for this data-dir), because when dumping, old data dir will be removed and a new only will be created. Maybe we can refactor to only clear the data in the dir not the dir itself later, then only permissions for data-dir are needed.

GregoryIan:
question 1 - I feel it's not complete correct
* two dump job does not necessarily give an error，because it need to look at the state of the runtime
* if there isn't error in dumping stage, loader would load from same dir
* for two task whose stage is already load, it would same with old implemention
for this question, if we don't have a way to let it be more correct, we can choose a simpler implementation 

quesion 2 -  should start from the statement, not the implementation. We can declare the permissions we need and what we might do. it will definitely be better than creating a directory separately.

GregoryIan:
@amyangfei what about you?

amyangfei:
* If we don't have to auto avoid name confliction, we can add one more check option, such as `check_loader_dir_conflict`, which will check `loader.dir` name confliction in task precheck. In this way DM doesn't add suffix, just lets user to choose a new `loader.dir` name or skip the precheck option `check_loader_dir_conflict`
* On the other hand, if we must support same `loader.dir` name in different tasks without any data confliction
> quesion 2 - should start from the statement, not the implementation. We can declare the permissions we need and what we might do. it will definitely be better than creating a directory separately.

I think it is not a problem, we just let user to configure a `loader.dir`, `dirname` and `dirname.taskname` have no root difference in current implement. We always need an `execute` privilege of a directory.

GregoryIan:
LGTM

PR #102, opened 2019-04-08
*: slice capacity prealloc tuning

amyangfei:
fix some slice prealloc size

1. prealloc slice capacity if we know its final length determinately
2. use `strings.Builder` when we generate column list 

### Check List 

Tests 

 - Unit test
 - Integration test

### DIFF
--- a/syncer/dml.go
+++ b/syncer/dml.go
@@ -287,9 +287,9 @@ func genColumnList(columns []*column) string {
 	var buf strings.Builder
 	for i, column := range columns {
 		if i != len(columns)-1 {
-			fmt.Fprintf(&buf, "`%s`,", column.name)
+			buf.WriteString("`" + column.name + "`,")
 		} else {
-			fmt.Fprintf(&buf, "`%s`", column.name)
+			buf.WriteString("`" + column.name + "`")
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


amyangfei:
PTAL @GregoryIan @csuzhangxc

syncer/dml.go
```diff
@@ -284,17 +284,16 @@ func genDeleteSQL(schema string, table string, value []interface{}, columns []*c
...
+	var buf strings.Builder
 	for i, column := range columns {
-		name := fmt.Sprintf("`%s`", column.name)
-		columnList = append(columnList, []byte(name)...)
-
 		if i != len(columns)-1 {
-			columnList = append(columnList, ',')
+			fmt.Fprintf(&buf, "`%s`,", column.name)
+		} else {
+			fmt.Fprintf(&buf, "`%s`", column.name)
```
suzaku: Using `&buf` would make the string builder escape to heap.

It would be faster if we do 

```go
buf.WriteString("`" + column.name + "`")
```

Proof: https://play.golang.org/p/Pjl36V-QGho
Analysis: `go test -gcflags="-m" -v --bench . escape_test.go`

suzaku:
LGTM

PR #103, opened 2019-04-08
tests: add retry in metric check, make it more robust

amyangfei:
in all mode test, DM-worker is just restarted before checking metric, the metric may have not been updated in time.

add retry in metric check

### Check List 

Tests 

 - Integration test

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -57,5 +57,5 @@ Several convenient commands are provided:
 * `check_port_alive <PORT>` - Wrapper to check a port is alive, at most 20 times.
 * `check_port <HOST> <PORT>` - Checks a host:port is alive.
 * `wait_process_exit <process_name>` - Wait for one or more processes to exit by given process name.
-* `check_metric <PORT> <METRIC_NAME> <VALUE PATTERN LIST>...` - check metric value from prometheus.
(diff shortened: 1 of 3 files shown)


amyangfei:
PTAL @GregoryIan @csuzhangxc

tests/README.md
```diff
@@ -57,5 +57,5 @@ Several convenient commands are provided:
 * `check_port_alive <PORT>` - Wrapper to check a port is alive, at most 20 times.
 * `check_port <HOST> <PORT>` - Checks a host:port is alive.
 * `wait_process_exit <process_name>` - Wait for one or more processes to exit by given process name.
-* `check_metric <PORT> <METRIC_NAME> <VALUE PATTERN LIST>...` - check metric value from prometheus.
+* `check_metric <PORT> <METRIC_NAME> <RETRY_TIME> <VALUE PATTERN LIST>...` - check metric value from prometheus.
```
csuzhangxc: is `RETRY_TIME` means time with second unit (e.g. 3 means 3s) or just times without unit.

amyangfei: here it means n time without second unit, each time in check_metric will wait 1s if check failed. I changed to `RETRY_COUNT`
btw if we want to use time unit, such as 1s, 1min, `RETRY_INTERVAL` may be better

GregoryIan:
LGTM

Issue #105, opened 2019-04-09
Building dm requires running mysqld


kolbe:
It looks like simply building dm requires a running mysqld for some reason. Maybe for a test suite, maybe for something else? It seems like this should not be a requirement simply for a build... or at minimum it ought to be better documented.

Is the solution as easy as removing "`test`" from the "`build`" target in Makefile?

amyangfei:
thanks @kolbe 
we have removed `test` from `build` target in Makefile, so close this issue now

PR #106, opened 2019-04-09
tests: update test entry in Makefile

amyangfei:
fix issue #105 

1. rename `make test` to `make unit_test`
2. add new `make test` including `make unit_test` and `make integration_test`

### Check List 

Tests 

 - Unit test
 - Integration test

Related changes

 - Need to update CI script


amyangfei:
PTAL @GregoryIan @csuzhangxc

PR #112, opened 2019-04-16
tests: add incremental mode test

amyangfei:
add incremental mode test

1. add test case, prepare data -> start task in full mode -> stop task -> incremental data -> start task in incremental mode
~2. remove useless code in checkpoint~ I will do it in another pr

### Check List 

Tests 

 - Integration test


amyangfei:
PTAL @GregoryIan @csuzhangxc

tests/incremental_mode/run.sh
```diff
@@ -0,0 +1,67 @@
...
+    run_dm_worker $WORK_DIR/worker1 $WORKER1_PORT $cur/conf/dm-worker1.toml
+    run_dm_worker $WORK_DIR/worker2 $WORKER2_PORT $cur/conf/dm-worker2.toml
+    run_dm_master $WORK_DIR/master $MASTER_PORT $cur/conf/dm-master.toml
+
+    check_port_alive $MASTER_PORT
+    check_port_alive $WORKER1_PORT
+    check_port_alive $WORKER2_PORT
+
+    cd $cur && GO111MODULE=on go build -o bin/dmctl dmctl.go && cd -
+    cd $cur && GO111MODULE=on go build -o bin/dmctl_stop dmctl_stop.go && cd -
```
csuzhangxc: can we re-use these binaries between multiple test cases?

amyangfei: added global dmctl tools, now each test case can use them without re-compiling

csuzhangxc: do we need to add retry for _dm_start_task.go_?

amyangfei: currently not needed, If we need retry mechanism we can wrap the binary from caller

csuzhangxc:
LGTM

tests/dmctl_tools/dmctl_operate_task.go
```diff
@@ -15,18 +15,28 @@ package main
...
+	if len(os.Args) != 3 {
+		utils.ExitWithError(fmt.Errorf("invlid args: %v", os.Args))
+	}
 	cli, err := utils.CreateDmCtl("127.0.0.1:8261")
 	if err != nil {
 		utils.ExitWithError(err)
 	}
-	conf := os.Args[1]
-	err = utils.StartTask(context.Background(), cli, conf, nil)
+	op, err := strconv.ParseInt(os.Args[1], 10, 64)
```
GregoryIan: can we use `string` as type of  argumen `op`? and then convert `op` to `pb.TaskOp` in this function

tests/utils/dmctl.go
```diff
@@ -58,3 +58,22 @@ func StartTask(ctx context.Context, cli pb.MasterClient, configFile string, work
 
 	return nil
 }
+
```
GregoryIan: why not moving it to `dm_tools` directory?

tests/dmctl_tools/dmctl_operate_task.go
amyangfei: changed to use string type argument, feel free to covert to pb.TaskOp then

tests/utils/dmctl.go
amyangfei: we put binary entry in `dm_tools` directory, and `utils` directory behave more like pkg

GregoryIan:
LGTM

PR #114, opened 2019-04-16
tests: fix random fail in decrypt unit test

amyangfei:
sometimes unit test in decrypt_test.go failed as following:
https://internal.pingcap.net/idc-jenkins/blue/organizations/jenkins/test_dm_master/detail/test_dm_master/75/pipeline/44

This is because we check whether the 16th char in decrypt text is '@'
https://github.com/pingcap/dm/blob/a5578266da5a08706b82ff8d65fb166b1a9a7975/pkg/encrypt/encrypt.go#L77
but in our test decrypt text, the 17th char could also be '@'

This commit only removes all '@' char from test decrypt text, which ensures the unit test case will always run successfully.

### Check List 

Tests 

 - Unit test

### DIFF
--- a/pkg/encrypt/encrypt_test.go
+++ b/pkg/encrypt/encrypt_test.go
@@ -85,7 +85,8 @@ func (t *testEncryptSuite) TestEncrypt(c *C) {
 	block, err := aes.NewCipher(secretKey)
 	c.Assert(err, IsNil)
 	blockSize := block.BlockSize()
-	plaintext3, err := Decrypt(append(ciphertext[1:blockSize+1], append([]byte{ivSep[0]}, ciphertext[blockSize+1:]...)...))
+	c.Assert(len(ciphertext), Greater, blockSize+2)
+	plaintext3, err := Decrypt(append(ciphertext[1:blockSize+1], append([]byte{ivSep[0]}, ciphertext[blockSize+2:]...)...))
 	c.Assert(err, IsNil)
 	c.Assert(plaintext3, Not(DeepEquals), plaintext)
 }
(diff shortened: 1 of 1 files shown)


pkg/encrypt/encrypt_test.go
```diff
@@ -64,6 +75,6 @@ func (t *testEncryptSuite) TestEncrypt(c *C) {
 	c.Assert(err, NotNil)
 
 	// invalid content
-	_, err = Decrypt(ciphertext[1:])
+	_, err = Decrypt(removeChar(ciphertext, ivSep[0]))
```
GregoryIan: are there something wrong with our implemention? I think old case is right

GregoryIan: ping @amyangfei

amyangfei: https://github.com/pingcap/dm/blob/39e09ddcf97bb6752ecdfba4e0c7c96826bc783b/pkg/encrypt/encrypt.go#L77-L79
the decrypt function checks the specific byte is `iv separator` only. In the old case, the ciphertext is  
```
[iv of block size, often 16] + @ + encrypted-plaintext
```
and we remove the first byte, the ciphertext becomes
```
[the last 15 bytes of original iv + @] + `iv separator` + [new encrypted-plaintext]
```
We can always decrypt the new encrypted-plaintext with new iv, as long long some length condition is satisfied. Our encrypt mechanism doesn't maintain enough information to decide whether the new ciphertext is valid

GregoryIan: I know it, we should add a special case for this `fail` case

GregoryIan:
LGTM

PR #116, opened 2019-04-18
*: refactor subtask management and communication model

GregoryIan:
The `subtask state machine` and `worker controller` mechanisms are not ideal now. 
Users use direct interaction to manipulate tasks, like user issue a create task request and the request fails, dm-worker wouldn't leave any information about it. the implementation is not conducive to debugging and make restoring tasks more difficult when dm-worker restarts. ref https://github.com/pingcap/dm/issues/77

* Introducing log in dm-worker to store task request
   * use leveldb as local kv db,  store `task meta` and `task operation log` in it
   * wrap `task operation log` as a queue-like object, providing operation
      * peek oldest non-handled log
      * forward handled pointer and mark the result of this log
   * make DM-worker's communication framework asynchronous
     * DM-worker would return success while it store the request operation into log
     * a background goroutine `worker/handleTask` would apply these logs in order and mark it's result
     * add a `waitOperationOk` function  for DM-server to make the asynchronous request be asynchronous
   * simplify  `subtask state machine`
     * only leave `Run`, `Close` operations and make them reentrant
      
I try to keep the original code of subtask.go and worker.go, and we will further simplify them while improve the test coverage.
In fact because the code is too much, I think it's better to issue first edition firstly. Don't worry, most of them are unit test.

**Some review concerns**
* review the design and implement from multiple pointer
   * log 
   * rpc interface
   * subtask state machine
   * worker controller
* some point are really different
   * asynchronous request -  even we make the asynchronous request be synchronous, but in fact it's different with old fashion, the former takes effect more slowly. In some failpoint panic testing , we should change checking result way
  * there a lot of problems related to the use and release and of resources, I remove `pause` and `resume` function, it may affect some mechanisms like `online ddl`

    

### Check List 

Tests 

 - Unit test
 - Integration test
 - Manual test
   * start a dm cluster
   * start/pause/stop/resume task
 - No code

Code changes
 - Has persistent data change

Side effects
 - Breaking backward compatibility -  we should upgrade dm-worker/master at the same time


GregoryIan:
@amyangfei @csuzhangxc PTAL, and notice `Some review concerns` in pr description

dm/worker/log.go
```diff
@@ -0,0 +1,498 @@
...
+// HandledPointerKey is key of HandledPointer which point to the last handled log
+var HandledPointerKey = []byte("!DM!handledPointer")
+
+// Pointer is a logic pointer that point to a location of log
+type Pointer struct {
+	Location int64
+}
+
+// MarshalBinary never return not nil err now
+func (p *Pointer) MarshalBinary() []byte {
```
csuzhangxc: not implement `BinaryMarshaler.MarshalBinary() (data []byte, err error)` (because it never returns a not nil err), but use it directly?

dm/worker/log.go
```diff
@@ -0,0 +1,498 @@
...
+		logs = make([]*pb.TaskLog, 0, 4)
+	)
+	iter := db.NewIterator(util.BytesPrefix(TaskLogPrefix), nil)
+	startLocation := handledPointer.Location + 1
+	for ok := iter.Seek(EncodeTaskLogKey(startLocation)); ok; ok = iter.Next() {
+		logBytes := iter.Value()
+		log := &pb.TaskLog{}
+		err = log.Unmarshal(logBytes)
+		if err != nil {
+			err = errors.Annotatef(err, "unmarshal task log %s", logBytes)
```
csuzhangxc: `%v`/`% X`/`%# X` or any other formater? also in L203 (maybe more?)

dm/worker/log.go
```diff
@@ -0,0 +1,498 @@
...
+			break
+		}
+
+		if log.Id >= endPointer.Location {
+			// move to next location
+			endPointer.Location = log.Id + 1
+			logs = append(logs, log)
+		} else {
+			panic("out of sorted from level db")
```
csuzhangxc: including something about `log`?

dm/worker/log.go
```diff
@@ -0,0 +1,498 @@
...
+		}
+	}
+	iter.Release()
+	if err != nil {
+		return nil, errors.Trace(err)
+	}
+
+	err = iter.Error()
+	if err != nil {
+		return nil, errors.Annotatef(err, "fetch logs from meta")
```
csuzhangxc:
```suggestion
		return nil, errors.Annotate(err, "fetch logs from meta")
```

dm/worker/log.go
```diff
@@ -0,0 +1,498 @@
...
+	if whetherNil(h) {
+		return nil, errors.Trace(ErrInValidHandler)
+	}
+
+	logBytes, err := h.Get(EncodeTaskLogKey(id), nil)
+	if err != nil {
+		if err == leveldb.ErrNotFound {
+			return nil, nil
+		}
+		return nil, errors.Annotatef(err, "get task meta from leveldb")
```
csuzhangxc: including `id` or `EncodeTaskLogKey(id)`?

dm/worker/log.go
```diff
@@ -0,0 +1,498 @@
...
+		return errors.Trace(ErrInValidHandler)
+	}
+
+	handledPointer := Pointer{
+		Location: ID,
+	}
+
+	err := db.Put(HandledPointerKey, handledPointer.MarshalBinary(), nil)
+	if err != nil {
+		return errors.Trace(err)
```
csuzhangxc: including `handledPointer`? and try to update more `errors.Trace`?

dm/worker/log.go
```diff
@@ -0,0 +1,498 @@
...
+
+// GC deletes useless log
+func (logger *Logger) GC(ctx context.Context, db *leveldb.DB) {
+	ticker := time.NewTicker(time.Hour)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+			log.Infof("[worker] meta gc goroutine exist!")
```
csuzhangxc: seems some log message not including `[worker]` mark in this file and some including other marks, is it expectant?

dm/worker/log.go
```diff
@@ -0,0 +1,498 @@
...
+	err := iter.Error()
+	if err != nil {
+		log.Errorf("query logs from meta error %v", err)
+	}
+
+	if batch.Len() > 0 {
+		err := db.Write(batch, nil)
+		if err != nil {
+			log.Errorf("fail to delete keys from kv db %v", err)
+		}
```
csuzhangxc: log something about this last batch?

dm/worker/db.go
```diff
@@ -0,0 +1,66 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package worker
+
+import (
+	"github.com/pingcap/dm/pkg/log"
+	"github.com/syndtr/goleveldb/leveldb"
```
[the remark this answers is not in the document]
GregoryIan: for example?

dm/worker/log.go
GregoryIan: It's also ok

dm/worker/log.go
```diff
@@ -0,0 +1,498 @@
...
+type Getter interface {
+	Get(key []byte, ro *opt.ReadOptions) ([]byte, error)
+}
+
+// HandledPointerKey is key of HandledPointer which point to the last handled log
+var HandledPointerKey = []byte("!DM!handledPointer")
+
+// Pointer is a logic pointer that point to a location of log
+type Pointer struct {
+	Location int64
```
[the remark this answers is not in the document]
GregoryIan: leave it `int64`, negative value can indicate invalid

dm/worker/log.go
```diff
@@ -0,0 +1,498 @@
...
+	}
+
+	return p, nil
+}
+
+var (
+	defaultGCForwardLog int64 = 10000
+
+	// TaskLogPrefix is  prefix of task log key
+	TaskLogPrefix = []byte("!DM!TaskLog")
```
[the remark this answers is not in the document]
GregoryIan: export all variables and functions of log for other tools

dm/worker/log.go
GregoryIan: `BinaryMarshaler.MarshalBinary() (data []byte, err error)` is better

dm/worker/log.go
GregoryIan: we need to define a unify log format for this case

dm/worker/log.go
```diff
@@ -0,0 +1,498 @@
...
+
+		if log.Id >= endPointer.Location {
+			// move to next location
+			endPointer.Location = log.Id + 1
+			logs = append(logs, log)
+		} else {
+			panic("out of sorted from level db")
+		}
+	}
+	iter.Release()
```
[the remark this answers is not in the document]
GregoryIan: leave it as best practice

dm/worker/db.go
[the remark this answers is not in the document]
csuzhangxc:
```
import (
	"bytes"
	"context"
	"fmt" // builtin

	"github.com/pingcap/errors"
	"github.com/pingcap/tidb-tools/pkg/check" // 3rd party

	"github.com/pingcap/dm/pkg/utils" // self project
)
```

dm/worker/log.go
csuzhangxc: yeh, only one unnecessary call to `fmt.Sprintf`.

dm/worker/meta.go
```diff
@@ -74,95 +77,240 @@ func (m *Meta) adjust() error {
...
-	} else if err != nil {
+	// restore from old metadata
+	oldPath := path.Join(dir, "meta")
+	fd, err := os.Open(oldPath)
+	// old metadata file exists, recover metadata from it
+	if err == nil {
+		fd.Close()
+		err = meta.recoverMetaFromOldFashion(oldPath)
+		if err != nil {
+			log.Errorf("[worker metadata]fail to recover from old metadata file %s, meta file may be correuption, error message: %v", oldPath, err)
```
csuzhangxc:
```suggestion
			log.Errorf("[worker metadata]fail to recover from old metadata file %s, meta file may be corrupt, error message: %v", oldPath, err)
```

dm/worker/meta.go
```diff
@@ -74,95 +77,240 @@ func (m *Meta) adjust() error {
...
-	if os.IsNotExist(err) {
-		log.Warnf("failed to open meta file %s, going to create a new one: %v", metaDB.path, err)
-		return metaDB, nil
-	} else if err != nil {
+	// restore from old metadata
+	oldPath := path.Join(dir, "meta")
+	fd, err := os.Open(oldPath)
+	// old metadata file exists, recover metadata from it
+	if err == nil {
+		fd.Close()
```
csuzhangxc: can we move this `Open`/`Close` into `recoverMetaFromOldFashion`? and do all of these (try/do) in `recoverMetaFromOldFashion`?

dm/worker/meta.go
```diff
@@ -74,95 +77,240 @@ func (m *Meta) adjust() error {
...
-		return metaDB, errors.Trace(err)
+		return nil, errors.Trace(err)
 	}
 
-	return metaDB, nil
+	meta.ctx, meta.cancel = context.WithCancel(context.Background())
+
+	meta.wg.Add(1)
+	go func() {
+		meta.wg.Done()
```
csuzhangxc:
```suggestion
		defer meta.wg.Done()
```

dm/worker/meta.go
```diff
@@ -74,95 +77,240 @@ func (m *Meta) adjust() error {
...
 }
 
-func (metaDB *FileMetaDB) save() error {
-	serialized, err := metaDB.meta.Toml()
+// LoadTaskMeta returns meta of all tasks
+func (meta *Metadata) LoadTaskMeta() map[string]*pb.TaskMeta {
+	meta.Lock()
+	defer meta.Unlock()
+
+	tasks := make(map[string]*pb.TaskMeta)
```
csuzhangxc:
```suggestion
	tasks := make(map[string]*pb.TaskMeta, len(meta.tasks))
```

dm/worker/meta.go
```diff
@@ -74,95 +77,240 @@ func (m *Meta) adjust() error {
...
+		task = CloneTaskMeta(t)
+	}
+	meta.RUnlock()
+
+	return
+}
+
+// GetTaskLog returns task log by give log ID
+func (meta *Metadata) GetTaskLog(opLogID int64) (*pb.TaskLog, error) {
+	log, err := meta.log.GetTaskLog(meta.db, opLogID)
```
csuzhangxc: why not use cache in `meta.logs`?

dm/worker/meta.go
```diff
@@ -74,95 +77,240 @@ func (m *Meta) adjust() error {
...
+func (meta *Metadata) MarkOperation(log *pb.TaskLog) error {
+	meta.Lock()
+	defer meta.Unlock()
+
+	if len(meta.logs) == 0 {
+		return errors.NotFoundf("any task operation log")
+	}
+
+	if meta.logs[0].Id != log.Id {
+		return errors.Errorf("please handle task oepration order by log ID, the log need to be handled is %+v, not %+v", meta.logs[0], log)
```
csuzhangxc:
```suggestion
		return errors.Errorf("please handle task operation order by log ID, the log need to be handled is %+v, not %+v", meta.logs[0], log)
```

dm/worker/server.go
```diff
@@ -136,115 +134,114 @@ func (s *Server) Close() {
...
+			Msg:    "",
+		},
+		Op:    pb.TaskOp_Update,
+		LogID: opLogID,
+	}, nil
+}
 
-	err = s.worker.UpdateSubTask(cfg)
+// QueryTaskOperation implements WorkerServer.QueryTaskOperation
+func (s *Server) QueryTaskOperation(ctx context.Context, req *pb.QueryTaskOperationRequest) (*pb.QueryTaskOperationResponse, error) {
```
csuzhangxc: do we use it now?

dm/worker/server.go
```diff
@@ -136,115 +134,114 @@ func (s *Server) Close() {
...
+	opLogID := req.LogID
+
+	opLog, err := s.worker.meta.GetTaskLog(opLogID)
 	if err != nil {
-		log.Errorf("[server] update sub task %s error %v", cfg.Name, errors.ErrorStack(err))
-		return &pb.CommonWorkerResponse{
-			Result: false,
-			Msg:    errors.ErrorStack(err),
-		}, nil
+		log.Errorf("fail to get operation log %d of task %s", opLogID, taskName)
```
csuzhangxc: use `errors.Annotatef`?

dm/master/server.go
```diff
@@ -247,13 +247,25 @@ func (s *Server) StartTask(ctx context.Context, req *pb.StartTaskRequest) (*pb.S
...
 				}
 			}
-			workerResp.Worker = worker
-			workerRespCh <- workerResp
+			err = s.waitOperationOk(ctx, cli, stCfg.Name, workerResp.LogID)
+			if err != nil {
+				workerResp = &pb.OperateSubTaskResponse{
+					Meta: &pb.CommonWorkerResponse{
+						Result: false,
+						Msg:    errors.ErrorStack(err),
```
csuzhangxc: is it possible this `err` overwrite `err` in L253?

dm/master/server.go
```diff
@@ -247,13 +247,25 @@ func (s *Server) StartTask(ctx context.Context, req *pb.StartTaskRequest) (*pb.S
...
+					},
 				}
 			}
-			workerResp.Worker = worker
-			workerRespCh <- workerResp
+			err = s.waitOperationOk(ctx, cli, stCfg.Name, workerResp.LogID)
```
csuzhangxc: `if err != nil`, then is `workerResp.LogID` valid?

dm/master/server.go
```diff
@@ -1752,3 +1789,35 @@ func (s *Server) generateSubTask(ctx context.Context, task string) (*config.Task
...
+		if err != nil {
+			log.Errorf("fail to query task operation %v", err)
+		} else if res.Log.Success {
+			return nil
+		} else if len(res.Log.Message) != 0 {
+			return errors.New(res.Log.Message)
+		}
+
+		log.Infof("wait task %s op log %d, current res %+v", name, opLogID, res)
+		time.Sleep(retryInterval)
```
csuzhangxc: use `time.After`?

dm/master/server.go
```diff
@@ -1752,3 +1789,35 @@ func (s *Server) generateSubTask(ctx context.Context, task string) (*config.Task
...
+		})
+		if err != nil {
+			log.Errorf("fail to query task operation %v", err)
+		} else if res.Log.Success {
+			return nil
+		} else if len(res.Log.Message) != 0 {
+			return errors.New(res.Log.Message)
+		}
+
+		log.Infof("wait task %s op log %d, current res %+v", name, opLogID, res)
```
csuzhangxc: replace `res` with `result`?

dm/worker/worker.go
```diff
@@ -662,3 +696,191 @@ func (w *Worker) MigrateRelay(ctx context.Context, binlogName string, binlogPos
...
+			switch opLog.Task.Op {
+			case pb.TaskOp_Start:
+				if exist {
+					err = errors.AlreadyExistsf("sub task %s", opLog.Task.Name)
+					break
+				}
+
+				if w.relayPurger.Purging() {
+					if retryCnt < maxRetryCount {
+						retryCnt++
```
csuzhangxc: reset to zero in some place?

dm/worker/worker.go
```diff
@@ -662,3 +696,191 @@ func (w *Worker) MigrateRelay(ctx context.Context, binlogName string, binlogPos
...
+					break
+				}
+
+				taskCfg := new(config.SubTaskConfig)
+				if err1 := taskCfg.Decode(string(opLog.Task.Task)); err1 != nil {
+					err = errors.Annotatef(err1, "decode subtask config error in handleTask")
+					break
+				}
+
+				cfgDecrypted, err := taskCfg.DecryptPassword()
```
csuzhangxc: this `err` hide the one in L787, is expectant?

dm/worker/worker.go
```diff
@@ -662,3 +696,191 @@ func (w *Worker) MigrateRelay(ctx context.Context, binlogName string, binlogPos
...
+			// fill current task config
+			if len(opLog.Task.Task) == 0 {
+				tm := w.meta.GetTask(opLog.Task.Name)
+				opLog.Task.Task = append([]byte{}, tm.Task...)
+			}
+
+			err = w.meta.MarkOperation(opLog)
+			w.Unlock()
+			if err != nil {
+				log.Errorf("failto mark subtask %s operation %+v", opLog.Task.Name, opLog)
```
csuzhangxc:
```suggestion
				log.Errorf("fail to mark subtask %s operation %+v", opLog.Task.Name, opLog)
```

dm/worker/db.go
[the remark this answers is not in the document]
GregoryIan: Any formatting tools recommend doing this? Currently, our import is formatted by some tools generally. If change the standard, it is best to declare it and change them all

dm/worker/log_test.go
```diff
@@ -0,0 +1,300 @@
...
+
+	bs, _ := p.MarshalBinary()
+	c.Assert(db.Put(HandledPointerKey, bs, nil), IsNil)
+	p, err = LoadHandledPointer(db)
+	c.Assert(err, IsNil)
+	c.Assert(p.Location, Equals, int64(1))
+
+	c.Assert(db.Put(HandledPointerKey, []byte("xx"), nil), IsNil)
+	_, err = LoadHandledPointer(db)
+	c.Assert(strings.Contains(err.Error(), "not valid length data as"), IsTrue)
```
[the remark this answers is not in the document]
GregoryIan: sure

dm/worker/log_test.go
```diff
@@ -0,0 +1,300 @@
...
+package worker
+
+import (
+	"strings"
+
+	. "github.com/pingcap/check"
+	"github.com/pingcap/dm/dm/pb"
+)
+
+func (t *testWorker) TestPointer(c *C) {
```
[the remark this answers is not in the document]
GregoryIan: Sure

dm/worker/db.go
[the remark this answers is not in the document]
csuzhangxc: I do not search for tools yet, only put them into 3 sections and then use _fmt_ to format the order.

dm/proto/dmworker.proto
```diff
@@ -4,12 +4,13 @@ package pb;
...
 service Worker {
-    rpc StartSubTask (StartSubTaskRequest) returns (CommonWorkerResponse) {}
+    rpc StartSubTask (StartSubTaskRequest) returns (OperateSubTaskResponse) {}
     rpc OperateSubTask (OperateSubTaskRequest) returns (OperateSubTaskResponse) {}
-    rpc UpdateSubTask (UpdateSubTaskRequest) returns (CommonWorkerResponse) {}
+    rpc UpdateSubTask (UpdateSubTaskRequest) returns (OperateSubTaskResponse) {}
 
     rpc QueryStatus (QueryStatusRequest) returns (QueryStatusResponse) {}
     rpc QueryError (QueryErrorRequest) returns (QueryErrorResponse) {}
+    rpc QueryTaskOperation(QueryTaskOperationRequest) returns (QueryTaskOperationResponse) {}
```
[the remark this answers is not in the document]
GregoryIan: I suggest we only keep  subtask config, and call rest all `task`. if not, would we have task meta and subtask meta, task operation and subtask operation?

dm/worker/meta.go
GregoryIan: Sure

dm/worker/meta.go
GregoryIan: easier

dm/worker/meta_test.go
```diff
@@ -14,48 +14,364 @@
...
+
+	tasks[testTask1.Name] = &pb.TaskMeta{
+		Op:   pb.TaskOp_Stop,
+		Name: testTask1.Name,
+		Task: []byte("xxxxx"),
+	}
+
+	// test clone
+	tasks2 := meta.LoadTaskMeta()
+	c.Assert(tasks2[testTask1.Name], DeepEquals, testTask1Meta)
```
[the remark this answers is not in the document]
GregoryIan: test clone, `Task` is changed (`xxxxx`)

dm/worker/server.go
GregoryIan: lol, we should have a standard: how to add log in every caller

dm/worker/status.go
```diff
@@ -81,25 +82,30 @@ func (w *Worker) Status(stName string) []*pb.SubTaskStatus {
...
-				stStatus.Status = &pb.SubTaskStatus_Check{Check: us.(*pb.CheckStatus)}
-			case pb.UnitType_Dump:
-				stStatus.Status = &pb.SubTaskStatus_Dump{Dump: us.(*pb.DumpStatus)}
-			case pb.UnitType_Load:
-				stStatus.Status = &pb.SubTaskStatus_Load{Load: us.(*pb.LoadStatus)}
-			case pb.UnitType_Sync:
-				stStatus.Status = &pb.SubTaskStatus_Sync{Sync: us.(*pb.SyncStatus)}
+			if cu != nil {
+				stStatus.Unit = cu.Type()
+				if st.Stage() == pb.Stage_Running {
```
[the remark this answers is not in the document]
GregoryIan: we don't store `status` now

dm/worker/subtask.go
```diff
@@ -141,10 +147,27 @@ func (st *SubTask) Init() error {
 
 // Run runs the sub task
 func (st *SubTask) Run() {
+	if st.Stage() == pb.Stage_Finished || st.Stage() == pb.Stage_Running {
```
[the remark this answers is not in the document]
GregoryIan: yep, not exist anymore

dm/worker/subtask.go
```diff
@@ -364,56 +382,8 @@ func (st *SubTask) Close() {
 
 	st.cancel()
 	st.closeUnits() // close all un-closed units
-	st.setStageIfNot(pb.Stage_Finished, pb.Stage_Stopped)
+	st.setStageIfNot(pb.Stage_Finished, pb.Stage_Paused)
 	st.wg.Wait()
-
-	close(st.DDLInfo)
```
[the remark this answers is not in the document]
GregoryIan: do it in Init

dm/worker/worker_test.go
```diff
@@ -0,0 +1,21 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package worker
+
+import (
+	. "github.com/pingcap/check"
+)
+
+func (t *testWorker) TestOperateTask(c *C) {
```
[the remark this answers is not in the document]
GregoryIan: yep

dm/worker/worker.go
```diff
@@ -165,119 +189,137 @@ func (w *Worker) Close() {
...
-		return errors.Trace(err)
+		return 0, errors.Annotatef(err, "[worker] encode subtask %+v into toml format", cfg)
 	}
 
-	w.subTasks[cfg.Name] = st
+	var opLogID int64
+	opLogID, err = w.operateSubTask(&pb.TaskMeta{
+		Op:   pb.TaskOp_Start,
+		Name: cfg.Name,
+		Task: append([]byte{}, cfgStr...),
```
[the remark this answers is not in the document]
GregoryIan: which is better?

dm/worker/meta_test.go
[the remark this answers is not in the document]
csuzhangxc: ❓ `Task` is changed in L178~L182?

dm/proto/dmworker.proto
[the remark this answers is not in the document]
csuzhangxc: then what about _subtask.go_ in dm-worker.

dm/worker/meta.go
```diff
@@ -74,95 +77,240 @@ func (m *Meta) adjust() error {
...
+	if err != nil {
+		if !os.IsNotExist(err) {
+			return errors.Trace(err)
+		}
+		return nil
+	}
+	fd.Close()
 
-	metaDB.meta.SubTasks[subTask.Name] = subTask
+	data, err := ioutil.ReadFile(path)
```
csuzhangxc: Can L270~278 be replaced by handing this `err`? then only need to open the file once.

dm/master/server.go
```diff
@@ -247,13 +247,26 @@ func (s *Server) StartTask(ctx context.Context, req *pb.StartTaskRequest) (*pb.S
 			}
 			workerResp, err := cli.StartSubTask(ctx, &pb.StartSubTaskRequest{Task: stCfgToml})
 			if err != nil {
```
amyangfei: can L249-266, L344-L362, L460-477 extract into a single function?

dm/master/server.go
amyangfei: this line can be removed now

dm/master/server.go
```diff
@@ -1752,3 +1792,35 @@ func (s *Server) generateSubTask(ctx context.Context, task string) (*config.Task
...
+		time.Sleep(retryInterval)
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-time.After(retryInterval):
+		}
+
+	}
+
+	return errors.Errorf("request is timeout, but request may be successful")
```
amyangfei:
```suggestion
	return errors.New("request is timeout, but request may be successful")
```

amyangfei:
```suggestion
	return errors.New("request is timeout, but request may be successful")
```

dm/worker/status.go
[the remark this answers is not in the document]
csuzhangxc: why? if paused, we can not query status anymore?

dm/worker/subtask.go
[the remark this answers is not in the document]
csuzhangxc: in `Init()`, we only create a new one, by not close the old one?

dm/proto/dmworker.proto
```diff
@@ -68,10 +70,19 @@ message OperateSubTaskRequest {
...
-    string worker = 3; // worker name, set by dm-master
-    string msg = 4;
+    CommonWorkerResponse meta = 1;
+    TaskOp op = 2;
+    int64 logID = 3;
+}
+
+message QueryTaskOperationRequest {
+    string name = 1; // sub task's name
+	int64 logID = 2; // operation log ID
```
amyangfei: this two lines are not alignment. also the following two lines

dm/master/server.go
```diff
@@ -1752,3 +1792,35 @@ func (s *Server) generateSubTask(ctx context.Context, task string) (*config.Task
...
+		if err != nil {
+			log.Errorf("fail to query task operation %v", err)
+		} else if res.Log.Success {
+			return nil
+		} else if len(res.Log.Message) != 0 {
+			return errors.New(res.Log.Message)
+		}
+
+		log.Infof("wait task %s op log %d, current result %+v", name, opLogID, res)
+		time.Sleep(retryInterval)
```
csuzhangxc: can remove it now, we have added `case <-time.After(retryInterval):` in L1820

dm/worker/db.go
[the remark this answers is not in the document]
amyangfei: [golang CodeReview Recommendations](https://github.com/golang/go/wiki/CodeReviewComments#imports) only tells us
> Imports are organized in groups, with blank lines between them. The standard library packages are always in the first group.
Besides I have tried the recommended [goimports](https://godoc.org/golang.org/x/tools/cmd/goimports) tool on DM, it only separates standard library package and adds some package name assign which is not same to its package path

dm/worker/worker.go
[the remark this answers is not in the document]
csuzhangxc: the performance may be little difference (not tested), but I think the second one is simpler. you can choose the one you like.

dm/worker/status.go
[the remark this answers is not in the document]
GregoryIan: now our pause equal stop, if subtask was stopped, no status will show in our framework

dm/worker/meta_test.go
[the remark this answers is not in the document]
GregoryIan: ```	Task: []byte("xxxxx"),```

dm/worker/status.go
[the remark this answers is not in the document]
csuzhangxc: then, if the task paused when errors occurred, can we query the status?

dm/worker/subtask.go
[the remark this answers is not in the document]
GregoryIan: `gc` would take over it, and I really don't know whether it would affect `SendBackDDLInfo` and what's useful of `SendBackDDLInfo`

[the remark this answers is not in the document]
csuzhangxc: previously, if the DM-master resolve the DDL lock failed, the DDL info will be sent back to DM-worker, so after DM-master restarted it can fetch DDL info from DM-worker again. but now `SendBackDDLInfo` is useless, because `SaveDDLInfo` introduced...

I think the better solution is to store the DDL info in levelDB, and we can do it later.

dm/worker/meta_test.go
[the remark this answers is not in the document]
GregoryIan: I add some comment for you, it's reverse logic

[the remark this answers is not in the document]
csuzhangxc: you mean, you want to verify that `tasks` in L172 is a clone of internal info in `meta`?

dm/worker/db.go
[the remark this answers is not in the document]
GregoryIan: unify to split 3 groups

dm/proto/dmworker.proto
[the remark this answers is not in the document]
GregoryIan: `subtask.go` is still `subtask.go`, but relate to `meta` and `operation` we use `task`

dm/worker/meta_test.go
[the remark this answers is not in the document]
GregoryIan: yep

GregoryIan:
I put modifications of subtask.go in other PR, PTAL again @csuzhangxc @amyangfei

dm/worker/status.go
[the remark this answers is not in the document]
GregoryIan: after task's paused
* if dm-worker not restarted, it would show status
* if dm-worker restarted, because we don't run subtask(it was paused), it would not show status

dm/master/server.go
```diff
@@ -246,14 +246,9 @@ func (s *Server) StartTask(ctx context.Context, req *pb.StartTaskRequest) (*pb.S
...
 			workerResp, err := cli.StartSubTask(ctx, &pb.StartSubTaskRequest{Task: stCfgToml})
-			if err != nil {
-				workerResp = &pb.CommonWorkerResponse{
-					Result: false,
-					Msg:    errors.ErrorStack(err),
-				}
-			}
-			workerResp.Worker = worker
-			workerRespCh <- workerResp
+			workerResp = s.handleOperationResult(ctx, cli, stCfg.Name, workerResp.LogID, err, workerResp)
```
csuzhangxc: if `err == nil`, `workerResp.LogID` may panic? only pass `workerResp` is enough?

dm/master/server.go
```diff
@@ -1752,3 +1739,57 @@ func (s *Server) generateSubTask(ctx context.Context, task string) (*config.Task
...
 }
+
+var (
+	maxRetryNum   = 30
+	retryInterval = time.Second
+)
+
+func (s *Server) waitOperationOk(ctx context.Context, cli pb.WorkerClient, name string, opLogID int64) error {
+	for num := 0; num < maxRetryNum; num++ {
+		res, err := cli.QueryTaskOperation(ctx, &pb.QueryTaskOperationRequest{
```
csuzhangxc: we can construct this `QueryTaskOperationRequest` object out of the `for` loop.

dm/master/server.go
```diff
@@ -1752,3 +1739,57 @@ func (s *Server) generateSubTask(ctx context.Context, task string) (*config.Task
...
+		log.Infof("wait task %s op log %d, current result %+v", name, opLogID, res)
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-time.After(retryInterval):
+		}
+
+	}
+
+	return errors.New("request is timeout, but request may be successful")
```
csuzhangxc: including the task name and log ID.

dm/proto/dmworker.proto
```diff
@@ -68,10 +70,19 @@ message OperateSubTaskRequest {
...
+}
+
+message QueryTaskOperationResponse {
+     CommonWorkerResponse meta = 1;
```
csuzhangxc: fix the indent? (5 spaces in this line)

dm/worker/log_test.go
```diff
@@ -0,0 +1,308 @@
...
+	c.Assert(db.Put(HandledPointerKey, []byte("xx"), nil), IsNil)
+	_, err = LoadHandledPointer(db)
+	c.Assert(err, ErrorMatches, ".*not valid length data as.*")
+}
+
+func (t *testLog) TestTaskLogKey(c *C) {
+	var id int64 = 1
+	idc, err := DecodeTaskLogKey(EncodeTaskLogKey(id))
+	c.Assert(err, IsNil)
+	c.Assert(idc, DeepEquals, id)
```
csuzhangxc:
```suggestion
	c.Assert(idc, Equals, id)
```

dm/worker/log_test.go
```diff
@@ -0,0 +1,308 @@
...
+
+	_, err = DecodeTaskLogKey([]byte("xx"))
+	c.Assert(err, ErrorMatches, ".*not valid length data as.*")
+}
+
+func (t *testLog) TestTaskLog(c *C) {
+	logger := new(Logger)
```
csuzhangxc: how about providing a function like `NewLogger` to create a `Logger` object, so if more fields added to `Logger`, less code needed to change.

dm/worker/log_test.go
```diff
@@ -0,0 +1,308 @@
...
+	c.Assert(err, IsNil)
+	c.Assert(hp.Location, Equals, int64(1))
+
+	c.Assert(logger.ForwardTo(db, 2), IsNil)
+	c.Assert(logger.handledPointer.Location, Equals, int64(2))
+	hp, err = LoadHandledPointer(db)
+	c.Assert(err, IsNil)
+	c.Assert(hp.Location, Equals, int64(2))
+
+	c.Assert(logger.ForwardTo(db, 0), IsNil)
```
csuzhangxc: do we need to support `ForwardTo` to _backward_?

dm/master/server.go
GregoryIan: it's response, unnecessary

dm/worker/log.go
```diff
@@ -0,0 +1,501 @@
...
+	defaultGCForwardLog int64 = 10000
+
+	// TaskLogPrefix is  prefix of task log key
+	TaskLogPrefix = []byte("!DM!TaskLog")
+)
+
+// DecodeTaskLogKey decodes task log key and returns its log ID
+func DecodeTaskLogKey(key []byte) (int64, error) {
+	if len(key) != len(TaskLogPrefix)+8 {
+		return 0, errors.Errorf("not valid length data as task log key %v", key)
```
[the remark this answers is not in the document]
GregoryIan: unnecessary

dm/worker/log.go
```diff
@@ -0,0 +1,501 @@
...
+	if err != nil {
+		return errors.Annotatef(err, "marshal task log %+v", opLog)
+	}
+
+	err = db.Put(EncodeTaskLogKey(opLog.Id), logBytes, nil)
+	if err != nil {
+		return errors.Annotatef(err, "save task log %d", opLog.Id)
+	}
+
+	return errors.Trace(logger.ForwardTo(db, opLog.Id))
```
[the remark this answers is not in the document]
GregoryIan: `db Putter` can be a transaction handler, and it's also ok to retry, it's advantage of idempotency

dm/worker/log_test.go
GregoryIan: now it's enough, we can do it while we need it

dm/worker/log_test.go
GregoryIan: Depending on the caller

dm/worker/meta_test.go
```diff
@@ -14,48 +14,366 @@
...
+	log1 = meta.PeekLog()
+	c.Assert(log1.Task, DeepEquals, testTask1Meta)
+	c.Assert(log1.Id, DeepEquals, int64(1))
+
+	log1C, err := meta.GetTaskLog(1)
+	c.Assert(err, IsNil)
+	c.Assert(log1C, DeepEquals, log1)
+
+	log2, err := meta.GetTaskLog(2)
+	c.Assert(log2.Task, DeepEquals, testTask2Meta)
```
csuzhangxc: maybe you should check `err` first, the same bellow.

dm/worker/meta_test.go
```diff
@@ -14,48 +14,366 @@
...
+		Id:   2,
+		Task: testTask2Meta,
+	})
+	c.Assert(err, ErrorMatches, ".*please handle task operation order by log ID.*")
+
+	// make sucessful  task1 create log
+	err = meta.MarkOperation(&pb.TaskLog{
+		Id:      1,
+		Task:    testTask1Meta,
+		Success: true,
```
csuzhangxc: add some comments about `Sucess`? when it will be `true` and what effect it will have?

dm/worker/subtask_test.go
```diff
@@ -0,0 +1,434 @@
...
+
+	// pause twice
+	c.Assert(st.Pause(), IsNil)
+	c.Assert(st.Stage(), Equals, pb.Stage_Paused)
+	c.Assert(st.CurrUnit(), Equals, mockDumper)
+	if st.Result() != nil && (!st.Result().IsCanceled || len(st.Result().Errors) > 0) {
+		c.Fatalf("result %+v is not right after closing", st.Result())
+	}
+
+	c.Assert(st.Pause(), NotNil)
```
csuzhangxc: use `ErrorMatches` for non-nil error to match the exact error message?

dm/worker/worker.go
```diff
@@ -49,62 +51,62 @@ type Worker struct {
...
-func (w *Worker) Start() {
-	w.closed.Set(closedFalse)
-	w.wg.Add(1)
-	defer w.wg.Done()
+	InitConditionHub(w)
 
-	log.Info("[worker] start running")
+	err = w.restoreSubTask()
+	if err != nil {
+		return nil, errors.Trace(err)
```
csuzhangxc: if `restoreSubTask ` fails here, should we need to release resources acquired before? like close the KV DB.

dm/worker/worker.go
```diff
@@ -165,119 +189,137 @@ func (w *Worker) Close() {
...
-	cfg.EnableGTID = w.cfg.EnableGTID
+	if w.closed.Get() == closedTrue {
+		return 0, errors.NotValidf("worker already closed")
+	}
 
-	// we can remove this from SubTaskConfig later, because syncer will always read from relay
-	cfg.AutoFixGTID = w.cfg.AutoFixGTID
+	cfgStr, err := cfg.Toml()
+	if err != nil {
+		return 0, errors.Annotatef(err, "[worker] encode subtask %+v into toml format", cfg)
```
csuzhangxc:
```suggestion
		return 0, errors.Annotatef(err, "encode subtask %+v into toml format", cfg)
```

dm/worker/worker.go
```diff
@@ -165,119 +189,137 @@ func (w *Worker) Close() {
...
+	var opLogID int64
+	opLogID, err := w.operateSubTask(&pb.TaskMeta{
+		Name: name,
+		Op:   pb.TaskOp_Pause,
+	})
+	if err != nil {
+		return 0, errors.Trace(err)
+	}
+
+	return opLogID, nil
```
csuzhangxc: extract similar code from `StopSubTask`/`PauseSubTask`/`ResumeSubTask` to a function has an arg named `op`?

dm/worker/worker.go
```diff
@@ -391,7 +426,10 @@ func (w *Worker) doFetchDDLInfo(ctx context.Context, ch chan<- *pb.DDLInfo) {
 	if !ok {
 		return // should not go here
 	}
+
+	w.RLock()
 	w.subTasks[v.Task].SaveDDLInfo(v)
```
csuzhangxc: can `w.subTasks[v.Task]` become `nil`?

dm/worker/worker.go
```diff
@@ -662,3 +703,194 @@ func (w *Worker) MigrateRelay(ctx context.Context, binlogName string, binlogPos
...
+func (w *Worker) restoreSubTask() error {
+	tasks := w.meta.LoadTaskMeta()
+
+	w.Lock()
+	defer w.Unlock()
+	w.subTasks = make(map[string]*SubTask)
+	for name, task := range tasks {
+		taskCfg := new(config.SubTaskConfig)
+		if err := taskCfg.Decode(string(task.Task)); err != nil {
+			return errors.Annotate(err, "decode subtask config error in restoreSubTask")
```
csuzhangxc: including `task.Task` in error message?

dm/worker/worker.go
```diff
@@ -662,3 +703,194 @@ func (w *Worker) MigrateRelay(ctx context.Context, binlogName string, binlogPos
...
+		}
+
+		cfgDecrypted, err := taskCfg.DecryptPassword()
+		if err != nil {
+			return errors.Trace(err)
+		}
+
+		log.Infof("[worker] start sub task with config: %v", cfgDecrypted)
```
csuzhangxc: the decrypted password will be logged. the same in L828.

csuzhangxc:
@GregoryIan should we provide some commands to query content in KV DB? because we cannot read it as plaintext now.

GregoryIan:
@csuzhangxc not now

dm/worker/meta_test.go
GregoryIan: both ok

dm/worker/subtask.go
```diff
@@ -141,10 +152,27 @@ func (st *SubTask) Init() error {
 
 // Run runs the sub task
 func (st *SubTask) Run() {
+	if st.Stage() == pb.Stage_Finished || st.Stage() == pb.Stage_Running {
```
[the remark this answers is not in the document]
GregoryIan: both are clear, not unclear case

dm/worker/subtask.go
```diff
@@ -141,10 +152,27 @@ func (st *SubTask) Init() error {
...
+		log.Warnf("[subtask] %s is %s", st.cfg.Name, st.Stage())
+		return
+	}
+
+	err := st.Init()
+	if err != nil {
+		log.Errorf("[subtask] fail to initial %v", err)
```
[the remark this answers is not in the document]
GregoryIan: log is already in `Init` 🤔

dm/worker/subtask_test.go
GregoryIan: no need now, I would change it

dm/worker/worker.go
GregoryIan: it's a part of startup logic, it seems also ok, letm see

dm/worker/worker.go
GregoryIan: 🤔 The essential problem is that the worker's concurrency model is really a bit bad, isn't it, but this pr hasn't planned to handle this yet

dm/worker/worker.go
GregoryIan: `json encode` wouldn't contain password in String() function?

dm/worker/worker.go
GregoryIan: sure

dm/worker/worker.go
GregoryIan: sure

dm/worker/meta_test.go
GregoryIan: sure

dm/worker/subtask.go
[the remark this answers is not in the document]
csuzhangxc: OK~

dm/worker/worker.go
```diff
@@ -84,9 +84,6 @@ func NewWorker(cfg *Config) (*Worker, error) {
 
 	// initial relay holder
 	err := w.relayHolder.Init()
-	if err != nil {
```
csuzhangxc: why remove this?

dm/worker/worker.go
csuzhangxc: oh, that's right.

dm/worker/worker.go
GregoryIan: update a wrong branch

dm/worker/worker.go
```diff
@@ -662,3 +651,194 @@ func (w *Worker) MigrateRelay(ctx context.Context, binlogName string, binlogPos
...
+					err = errors.AlreadyExistsf("sub task %s", opLog.Task.Name)
+					break
+				}
+
+				if w.relayPurger.Purging() {
+					if retryCnt < maxRetryCount {
+						retryCnt++
+						log.Warnf("relay log purger is purging, cannot start sub task %s, would try again later", opLog.Task.Name)
+						w.Unlock()
+						continue Loop
```
amyangfei: Is there any difference to use `continue Loop` and `continue` here?

GregoryIan: no,  it's a coding habits to make structure more clear

dm/worker/log.go
```diff
@@ -0,0 +1,501 @@
...
+
+// GC deletes useless log
+func (logger *Logger) GC(ctx context.Context, db *leveldb.DB) {
+	ticker := time.NewTicker(time.Hour)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+			log.Infof("[task log gc] goroutine exist!")
```
amyangfei:
```suggestion
			log.Info("[task log gc] goroutine exits!")
```

dm/worker/worker.go
```diff
@@ -165,119 +189,85 @@ func (w *Worker) Close() {
 	// cancel status output ticker and wait for return
 	w.cancel()
```
amyangfei: the close sequence is close
`relayHolder` -> `relayPurger` -> `meta log` -> `worker it self`
we'd better to ensure the component closed later must not reference to object defined in closed earlier. If not ensured, the behavior may be unpredictable.
For example, currently in worker there is a `handleTask` background routine, in this routine `meta` and `relayPurger` may be used. I'm not sure whether this is the root case that causes the not exited state dm-worker.

GregoryIan: Anyway, you're right. I admire you

GregoryIan: I review the implementation, it's not luck to find the deadlock, but I still fix it and re-run integrated test

dm/master/server.go
```diff
@@ -246,14 +246,9 @@ func (s *Server) StartTask(ctx context.Context, req *pb.StartTaskRequest) (*pb.S
...
 			workerResp, err := cli.StartSubTask(ctx, &pb.StartSubTaskRequest{Task: stCfgToml})
-			if err != nil {
-				workerResp = &pb.CommonWorkerResponse{
-					Result: false,
-					Msg:    errors.ErrorStack(err),
-				}
-			}
-			workerResp.Worker = worker
-			workerRespCh <- workerResp
+			workerResp = s.handleOperationResult(ctx, cli, stCfg.Name, err, workerResp)
```
csuzhangxc: some fields in `workerResp` (like `LogID`) will be overwritten by this `handleOperationResult`.
if `handleOperationResult` success it's fine, but if `handleOperationResult` failed, some fields lost.

GregoryIan: it wouldn't fail, definition is it that like you said - `handleOperationResult` would handle and return `response`

csuzhangxc:
```
» resume-task test-2
{
    "op": "Resume",
    "result": true,
    "msg": "",
    "workers": [
        {
            "meta": {
                "result": false,
                "worker": "127.0.0.1:8262",
                "msg": "current stage is not paused not valid\ngithub.com/pingcap/dm/dm/master.(*Server).waitOperationOk\n\t/Users/zhangxc/gopath/src/github.com/pingcap/dm/dm/master/server.go:1761\ngithub.com/pingcap/dm/dm/master.(*Server).handleOperationResult\n\t/Users/zhangxc/gopath/src/github.com/pingcap/dm/dm/master/server.go:1786\ngithub.com/pingcap/dm/dm/master.(*Server).OperateTask.func1\n\t/Users/zhangxc/gopath/src/github.com/pingcap/dm/dm/master/server.go:327\nruntime.goexit\n\t/usr/local/go/src/runtime/asm_amd64.s:1337"
            },
            "op": "Resume",
            "logID": "0"
        }
    ]
}
```

for the example above, `logID` lost.

GregoryIan: yep, you're right. I mistakenly think it is CommonWorkerResponse

PR #117, opened 2019-04-18
*: add `Writer` for relay log

csuzhangxc:

this PR is a part of #91.

- add a `FileWriter` to support to write binlog events into files
	- write a binlog file header for any new binlog file
	- ignore any fake `RotateEvent`
	- ignore any duplicate events
	- fill any potential holes in the binlog file with dummy events
- The `FileWriter` support to recover any binlog file with uncompleted events/transactions

### Check List 

Tests 

 - Unit test

Code changes

Side effects

 - Increased code complexity


relay/writer/file_test.go
```diff
@@ -0,0 +1,115 @@
...
+
+func (t *testFileWriterSuite) TestRelayDir(c *check.C) {
+	cfg := &FileConfig{}
+
+	// no dir specified
+	w1 := NewFileWriter(cfg)
+	defer w1.Close()
+	err := w1.Start()
+	c.Assert(err, check.NotNil)
+	c.Assert(strings.Contains(err.Error(), "no such file or directory"), check.IsTrue)
```
kennytm:
```suggestion
	c.Assert(err, check.ErrorMatches, ".*no such file or directory.*")
```

The check that `err, NotNil` can be removed since `ErrorMatches` also asserts that `err` is not nil.

csuzhangxc:
@GregoryIan @amyangfei PTAL

relay/writer/file.go
```diff
@@ -332,23 +332,28 @@ func (w *FileWriter) handleDuplicateEventsExist(ev *replication.BinlogEvent) (*R
 // 2.
 //    a. update the position with the event's position if the transaction finished
 //    b. update the GTID set with the event's GTID if the transaction finished
-// 3. truncate any un-finished events/transactions
+// 3. truncate any uncompleted events/transactions
```
kennytm: "unfinished" is fine, or use "incomplete", but "uncompleted" is definitely wrong.

relay/writer/writer.go
```diff
@@ -0,0 +1,67 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package writer
+
+import (
+	"github.com/pingcap/dm/pkg/gtid"
```
amyangfei: this is out of accord with our import rule

relay/writer/file.go
```diff
@@ -0,0 +1,377 @@
...
+	rotateEv, ok := ev.Event.(*replication.RotateEvent)
+	if !ok {
+		return nil, errors.NotValidf("except RotateEvent, but got %+v", ev.Header)
+	}
+
+	// update binlog filename if needed
+	var currFile = w.filename.Get()
+	nextFile := string(rotateEv.NextLogName)
+	if nextFile > currFile {
+		w.filename.Set(nextFile) // record the next filename, but not create it.
```
amyangfei: is it better to record the filename after fake RotateEvent and `w.out != nil` check

pkg/binlog/event/event.go
```diff
@@ -27,6 +27,12 @@ import (
 	"github.com/pingcap/dm/pkg/gtid"
 )
 
+// flags used in Header
+const (
+	HeaderFlagSuppressUse uint16 = 0x0008 // LOG_EVENT_SUPPRESS_USE_F
+	HeaderFlagRelayLog    uint16 = 0x0040 // LOG_EVENT_RELAY_LOG_F
```
amyangfei: we can use `LOG_EVENT_SUPPRESS_USE_F` and `LOG_EVENT_RELAY_LOG_F` from go-mysql/replication/const.go directly?

pkg/binlog/event/helper.go
```diff
@@ -0,0 +1,90 @@
...
+		return nil, errors.Errorf("the event should be a MariadbGTIDListEvent, but got %T", e.Event)
+	}
+
+	ggSet, err := gmysql.ParseMariadbGTIDSet("")
+	if err != nil {
+		return nil, errors.Annotatef(err, "parse empty MariaDB GTID set")
+	}
+	mGSet := ggSet.(*gmysql.MariadbGTIDSet)
+	for _, set := range gtidListEv.GTIDs {
+		setClone := set
```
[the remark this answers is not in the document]
csuzhangxc: when using `range` iterating, each element in `gtidListEv.GTIDs` are copied to the **same** `set` variable, then if we use `&set` (**pointer**) we may get only **one of** element in `gtidListEv.GTIDs`, so we use another local variable to hold the element and make `&setClone` happy.

add a comment.

relay/writer/file.go
csuzhangxc: sometimes, we may only get a _fake_ RotateEvent before the FormatDescriptionEvent and no _filename_ specified when creating the writer, then we need the _filename_ in the fake RotateEvent when writing the FormatDescriptionEvent.

add a comment.

pkg/binlog/event/event.go
```diff
@@ -788,3 +801,57 @@ func GenMariaDBGTIDEvent(header *replication.EventHeader, latestPos uint32, sequ
...
+// ref: https://dev.mysql.com/doc/internals/en/user-var-event.html
+// ref: https://github.com/MariaDB/server/blob/a765b19e5ca31a3d866cdbc8bef3a6f4e5e44688/sql/log_event.cc#L4950
+func GenDummyEvent(header *replication.EventHeader, latestPos uint32, eventSize uint32) (*replication.BinlogEvent, error) {
+	if eventSize < MinUserVarEventLen {
+		return nil, errors.Errorf("required dummy event size (%d) is too small, the minimum supported size is %d", eventSize, MinUserVarEventLen)
+	}
+
+	// modify flag in the header
+	headerClone := *header // do a copy
+	headerClone.Flags |= HeaderFlagSuppressUse
```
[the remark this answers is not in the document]
csuzhangxc: added

pkg/binlog/event/event_test.go
```diff
@@ -716,3 +716,84 @@ func (t *testEventSuite) TestGenMariaDBGTIDEvent(c *C) {
...
+		ServerID:  11,
+		Flags:     replication.LOG_EVENT_THREAD_SPECIFIC_F,
+	}
+	eventSizeList = append(eventSizeList, MinUserVarEventLen) // add a USER_VAR_EVENT
+	for _, eventSize = range eventSizeList {
+		ev, err2 := GenDummyEvent(header, latestPos, eventSize)
+		c.Assert(err2, IsNil)
+		c.Assert(ev, NotNil)
+		// verify the header flag
+		c.Assert(userVarEv.Header.Flags&replication.LOG_EVENT_THREAD_SPECIFIC_F, Equals, uint16(0))
```
GregoryIan: replace `userVarEv` with `ev`, and can we verify it in cases above directly?

pkg/binlog/event/helper.go
```diff
@@ -0,0 +1,90 @@
...
+	switch ev := e.Event.(type) {
+	case *replication.MariadbGTIDListEvent:
+		gtidListEv = ev
+	default:
+		return nil, errors.Errorf("the event should be a MariadbGTIDListEvent, but got %T", e.Event)
+	}
+
+	ggSet, err := gmysql.ParseMariadbGTIDSet("")
+	if err != nil {
+		return nil, errors.Annotatef(err, "parse empty MariaDB GTID set")
```
GregoryIan:
```suggestion
		return nil, errors.Annotatef(err, "initial a MariaDB GTID set")
```

pkg/binlog/event/helper.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+	mGSet := ggSet.(*gmysql.MariadbGTIDSet)
+	for _, set := range gtidListEv.GTIDs {
```
GregoryIan: it's `mariadb GTID` not `set`

pkg/binlog/event/event_test.go
csuzhangxc: changed to verify it in cases above directly

pkg/binlog/event/helper.go
csuzhangxc: change it to `mGTID`, use `set` only because we have `mGSet.AddSet` below. 😿

pkg/binlog/writer/file.go
```diff
@@ -0,0 +1,159 @@
...
+	Stage    string `json:"stage"`
+	Filename string `json:"filename"`
+	Offset   int64  `json:"offset"`
+}
+
+// String implements Stringer.String.
+func (s *FileWriterStatus) String() string {
+	data, err := json.Marshal(s)
+	if err != nil {
+		log.Errorf("[FileWriterStatus] marshal status to json error %v", err)
```
GregoryIan:
```suggestion
		log.Errorf("[FileWriterStatus] marshal status %+v to json error %v", s, err)
```

GregoryIan: if error is not nil, return the error message instead of empty string?

pkg/binlog/writer/file.go
```diff
@@ -0,0 +1,159 @@
...
+	w.mu.Lock()
+	defer w.mu.Unlock()
+
+	if w.stage != stageNew {
+		return errors.Errorf("stage %s, expect %s, already started", w.stage, stageNew)
+	}
+
+	f, err := os.OpenFile(w.cfg.Filename, os.O_RDWR|os.O_CREATE|os.O_APPEND, 0644)
+	if err != nil {
+		return errors.Annotatef(err, "open file %s", w.cfg.Filename)
```
GregoryIan: it seems `err` contains filename information

pkg/binlog/writer/file.go
```diff
@@ -0,0 +1,159 @@
...
+	w.stage = stagePrepared
+	return nil
+}
+
+// Close implements Writer.Close.
+func (w *FileWriter) Close() error {
+	w.mu.Lock()
+	defer w.mu.Unlock()
+
+	if w.stage == stageClosed {
```
GregoryIan: also check `stage == prepared`?

pkg/binlog/writer/file.go
```diff
@@ -0,0 +1,159 @@
...
+	"github.com/pingcap/dm/pkg/log"
+)
+
+// FileWriter is a binlog event writer which writes binlog events to a file.
+type FileWriter struct {
+	cfg *FileWriterConfig
+
+	mu     sync.RWMutex
+	stage  writerStage
+	offset sync2.AtomicInt64
```
GregoryIan: I think we can save it and fetch offset from file object, it's also in memory

csuzhangxc: currently, we still can fetch status even if `FileWriter` closed.

pkg/binlog/writer/file_test.go
```diff
@@ -0,0 +1,133 @@
...
+	c.Assert(w.Flush(), NotNil)
+
+	// try to read the data back
+	f, err := os.Open(filename)
+	c.Assert(err, IsNil)
+	defer f.Close()
+
+	dataLen := allData.Len()
+	buf := make([]byte, dataLen)
+	n, err := f.Read(buf)
```
GregoryIan: better to read all content

pkg/binlog/writer/file.go
csuzhangxc: changed to return an error message.

pkg/binlog/writer/file.go
csuzhangxc: change to use `errors.Trace` directly.

pkg/binlog/writer/file.go
csuzhangxc: changed to use check `stage != stagePrepared`, and only support to close it when is `stagePrepared`.

pkg/binlog/writer/file_test.go
csuzhangxc: addressed

relay/writer/stage.go
```diff
@@ -0,0 +1,36 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package writer
+
+type writerStage int32
+
+const (
```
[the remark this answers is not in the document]
csuzhangxc: I'm considering, but I'm not sure they are common. Any advice?

GregoryIan:
unit test is broken

csuzhangxc:
> unit test is broken

@GregoryIan tests passed. but the problem seems caused by un-flushed binlog events, see https://github.com/pingcap/dm/pull/143

relay/writer/stage.go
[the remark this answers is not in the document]
GregoryIan: if a object don't have more state like `paused`， it's common

relay/writer/file_util.go
```diff
@@ -0,0 +1,268 @@
...
+		return false, errors.Annotatef(err, "open file %s", filename)
+	}
+	defer f.Close()
+
+	fileHeaderLen := len(replication.BinLogFileHeader)
+	buff := make([]byte, fileHeaderLen)
+	n, err := f.Read(buff)
+	if n == 0 && err == io.EOF {
+		return false, nil // empty file
+	} else if n != fileHeaderLen {
```
GregoryIan: check` err != nil && err != EOF` firstly at this branch

relay/writer/file_util.go
```diff
@@ -0,0 +1,268 @@
...
+	"github.com/siddontang/go-mysql/replication"
+
+	"github.com/pingcap/dm/pkg/binlog/event"
+	"github.com/pingcap/dm/pkg/binlog/reader"
+	"github.com/pingcap/dm/pkg/gtid"
+	"github.com/pingcap/dm/relay/common"
+)
+
+// checkBinlogHeaderExist checks if the file has a binlog file header.
+func checkBinlogHeaderExist(filename string) (bool, error) {
```
GregoryIan: can we split it into two functions to save open file operation, like in `checkFormatDescriptionEventExist`

1. checkBinlogHeaderExist(filename string)
2. checkBinlogHeaderExist(fd *File)

relay/writer/file_util.go
```diff
@@ -0,0 +1,268 @@
...
+		return false, errors.Errorf("no binlog file header at the beginning for %s", filename)
+	}
+
+	f, err := os.Open(filename)
+	if err != nil {
+		return false, errors.Annotatef(err, "open file %s", filename)
+	}
+	defer f.Close()
+
+	// check whether only the file header
```
GregoryIan: I think we can save this logic

relay/writer/file_util.go
```diff
@@ -0,0 +1,268 @@
...
+			return errors.Errorf("got %+v, expect FormatDescriptionEvent", e.Header)
+		}
+		found = true
+		return nil
+	}
+
+	eof, err := replication.NewBinlogParser().ParseSingleEvent(f, onEventFunc)
+	if err != nil {
+		return false, errors.Annotatef(err, "parse %s", filename)
+	} else if eof {
```
GregoryIan: would `eof == true && found == true` happen?

relay/writer/file_util.go
```diff
@@ -0,0 +1,268 @@
...
+	if err != nil {
+		return false, errors.Annotatef(err, "get stat for %s", filename)
+	}
+	evStartPos := int64(ev.Header.LogPos - ev.Header.EventSize)
+	evEndPos := int64(ev.Header.LogPos)
+	if fs.Size() <= evStartPos {
+		return false, nil // the event not in the file
+	} else if fs.Size() < evEndPos {
+		// the file can not hold the whole event, often because the file is corrupt
+		return false, errors.Errorf("file size %d is between event's start pos (%d) and end pos (%d)",
```
GregoryIan: are there undetermined case, like flush partial binlog

relay/writer/file_util.go
```diff
@@ -0,0 +1,268 @@
...
+	var ev2 *replication.BinlogEvent
+	for {
+		ev2, err = r.GetEvent(ctx)
+		if err != nil {
+			return false, errors.Annotatef(err, "get event from %s", syncStartPos)
+		} else if ev2 == nil {
+			// in fact, this should not happen
+			return false, errors.Errorf("get none event from %s", syncStartPos)
+		}
+		if evStartPos > int64(len(replication.BinLogFileHeader)) &&
```
GregoryIan: do we need `evStartPos > int64(len(replication.BinLogFileHeader))`?

relay/writer/file_util.go
```diff
@@ -0,0 +1,268 @@
...
+		// the file can not hold the whole event, often because the file is corrupt
+		return false, errors.Errorf("file size %d is between event's start pos (%d) and end pos (%d)",
+			fs.Size(), evStartPos, evEndPos)
+	}
+
+	// 2. use a FileReader to read events from the start pos of the passed-in event
+	rCfg := &reader.FileReaderConfig{
+		EnableRawMode: true,
+	}
+	r := reader.NewFileReader(rCfg)
```
GregoryIan: can we read file directly if we know event size and raw data of `ev`?

relay/writer/file_util.go
```diff
@@ -0,0 +1,268 @@
...
+			u, _ := uuid.FromBytes(ev.SID)
+			nextGTIDStr = fmt.Sprintf("%s:%d", u.String(), ev.GNO)
+		case *replication.MariadbGTIDEvent:
+			if latestGSet == nil {
+				return 0, nil, errors.Errorf("should have a MariadbGTIDListEvent before the MariadbGTIDEvent %+v", e.Header)
+			}
+			// learn from: https://github.com/siddontang/go-mysql/blob/c6ab05a85eb86dc51a27ceed6d2f366a32874a24/replication/binlogsyncer.go#L745
+			GTID := ev.GTID
+			nextGTIDStr = fmt.Sprintf("%d-%d-%d", GTID.DomainID, GTID.ServerID, GTID.SequenceNumber)
+		case *replication.MariadbGTIDListEvent:
```
GregoryIan: I remember  event parser of `go-mysql` handles these logic, can we use it's logic in file reader?

relay/writer/file.go
```diff
@@ -0,0 +1,381 @@
...
+
+	return nil
+}
+
+// Close implements Writer.Close.
+func (w *FileWriter) Close() error {
+	w.mu.Lock()
+	defer w.mu.Unlock()
+
+	if w.stage == stageClosed {
```
GregoryIan: check `stagePrepared`

relay/writer/stage.go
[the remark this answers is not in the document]
GregoryIan: and if we provide a state machine that support concurrency, it would be cool

relay/writer/file.go
```diff
@@ -0,0 +1,381 @@
...
+	if w.out != nil {
+		log.Infof("[relay] closing previous underlying binlog writer with status %v", w.out.Status())
+		err := w.out.Close()
+		if err != nil {
+			return nil, errors.Annotate(err, "close previous underlying binlog writer")
+		}
+	}
+
+	// open/create a new binlog file
+	filename := filepath.Join(w.cfg.RelayDir, w.filename.Get())
```
GregoryIan: I think we should add a forcing file name correctness checking

relay/writer/file.go
```diff
@@ -0,0 +1,381 @@
...
+// FileWriter implements Writer interface.
+type FileWriter struct {
+	cfg *FileConfig
+
+	mu    sync.RWMutex
+	stage writerStage
+
+	// underlying binlog writer,
+	// it will be created/started until needed.
+	out bw.Writer
```
GregoryIan: use `FileWriter` directly?

relay/writer/file.go
```diff
@@ -0,0 +1,381 @@
...
+
+	filename sync2.AtomicString // current binlog filename
+	offset   sync2.AtomicInt64  // the file size may > 4GB
```
GregoryIan: can we save it and use offset of `FileWriter` directly?

relay/writer/file.go
```diff
@@ -0,0 +1,381 @@
...
+		// we can only receive a fake RotateEvent before the FormatDescriptionEvent.
+		w.filename.Set(nextFile)
+	}
+
+	// write the RotateEvent if not fake
+	var ignore bool
+	if ev.Header.Timestamp == 0 || ev.Header.LogPos == 0 {
+		// skip fake rotate event
+		ignore = true
+	} else if w.out == nil {
```
GregoryIan: must it open `out` at `handleFormatDescriptionEvent`?

relay/writer/file.go
```diff
@@ -0,0 +1,381 @@
...
+
+// handleFileHoleExist tries to handle a potential hole after this event wrote.
+// A hole exists often because some binlog events not sent by the master.
+// NOTE: handle cases when file size > 4GB
+func (w *FileWriter) handleFileHoleExist(ev *replication.BinlogEvent) error {
+	// 1. detect whether a hole exists
+	evStartPos := int64(ev.Header.LogPos - ev.Header.EventSize)
+	fileOffset := w.offset.Get()
+	holeSize := evStartPos - fileOffset
+	if holeSize <= 0 {
```
GregoryIan: does `< 0` implie error :thinking: ?

GregoryIan:
Great Work!

relay/writer/stage.go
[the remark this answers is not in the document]
csuzhangxc: move `Stage` into _pkg/binlog/common_ and replace original `readerStage`/`writerStage`. if needed, we can share it between others pkgs.

and We can keep considering how to provide a state machine that supports concurrency.

relay/writer/file_util.go
csuzhangxc: addressed

relay/writer/file_util.go
csuzhangxc: 👍 splited

relay/writer/file_util.go
csuzhangxc: we need to diff two kinds of EOF when parsing the FormatDescriptionEvent:
1. only the file header
2. part of the FormatDescriptionEvent header

if we save this logic, we need another logic to diff the returned values of `ParseSingleEvent`.

relay/writer/file_util.go
csuzhangxc: in the current implementation of _go-mysql_, should not happen.
I added a `if` to do the check if _go-mysql_ changed someday, we can observe it.

relay/writer/file_util.go
csuzhangxc: I think if one _fd_ write some data but not flushed other _fd_ still can see them, is it right? If it is, then this `if` should be `false`. 

NOTE: these file util functions are not concurrency-safe. 

if only flushed partial binlog event and this `if` is `true`, then we need to _recover_ the binlog file.

relay/writer/file_util.go
csuzhangxc: needed, if `evStartPos <= int64(len(replication.BinLogFileHeader))`, then we are check the FORMAT_DESCRIPTION_EVENT itself, so should not skip it.

relay/writer/file_util.go
csuzhangxc: we can do the read by using the file directly, but we also need to decode it to a binlog event that needs the _FORMAT_DESCRIPTION_EVENT_, so use `FileReader.StartSyncByPos` is simpler.

relay/writer/file_util.go
csuzhangxc: you mean computes GTID set in `FileReader` rather than in this _file_util_?
I want `FileReader` returns uniform binlog event to the caller, but the `replication.BinlogEvent` cannot carry other fields easily.

relay/writer/file.go
csuzhangxc: addressed

relay/writer/file.go
csuzhangxc: added checking, and some similar logic with _pkg/streamer/file.go_ can be refactored later.

relay/writer/file.go
csuzhangxc: I choose to use an interface because may mock it for testing later.

relay/writer/file.go
csuzhangxc: 👍 addressed

relay/writer/file.go
csuzhangxc: yes, before any non-fake events (needed to be written) are sent, a valid `FormatDescriptionEvent` event will always be sent, so we can open `out` at `handleFormatDescriptionEvent`

relay/writer/file.go
csuzhangxc: `< 0 ` implies duplicate events may exist, that should be handled in `handleDuplicateEventsExist`, in this function (`handleFileHoleExist`) we only handle a potential hole.

csuzhangxc:
@GregoryIan @amyangfei PTAL again

relay/writer/file_util.go
GregoryIan: if `ParseSingleEvent parsed too many events` is questionable, should we check log pos? it must be right

relay/writer/file_util.go
GregoryIan: if  is writing, it's possible. you should define to use it in non-concurrent scene

relay/writer/file_util.go
GregoryIan: start position of real `FORMAT_DESCRIPTION_EVENT` is `len(replication.BinLogFileHeader)`, right?  and can we use `len(replication.BinLogFileHeader)` instead of `checkFormatDescriptionEventExist`?

relay/writer/file_util.go
GregoryIan: Must need to decode?

relay/writer/file_util.go
csuzhangxc: added a check for log pos, and return `true` if `eof == true && found == true`

relay/writer/file_util.go
csuzhangxc: add some comments for these util functions.

relay/writer/file_util.go
csuzhangxc: changed to compare the raw data.

relay/writer/file.go
```diff
@@ -0,0 +1,381 @@
...
+	if w.stage != common.StagePrepared {
+		return nil, errors.Errorf("stage %s, expect %s, please start the writer first", w.stage, common.StagePrepared)
+	}
+
+	return w.doRecovering(p)
+}
+
+// WriteEvent implements Writer.WriteEvent.
+func (w *FileWriter) WriteEvent(ev *replication.BinlogEvent) (*Result, error) {
+	w.mu.RLock()
```
GregoryIan: `Lock` or `RLock`?

csuzhangxc: The purpose of this mutex is to ensure only do the write operation in `StagePrepared`, so I use `RLock`. for `WriteEvent` itself is not supported by concurrent usage, and I added a comment in _relay/writer/write.go_ L62.

relay/writer/file.go
```diff
@@ -0,0 +1,381 @@
...
+	defer w.mu.RUnlock()
+
+	if w.stage != common.StagePrepared {
+		return errors.Errorf("stage %s, expect %s, please start the writer first", w.stage, common.StagePrepared)
+	}
+
+	if w.out != nil {
+		return w.out.Flush()
+	}
+	return nil
```
GregoryIan: it's better to return errors

csuzhangxc: 👍 addressed

relay/writer/file.go
```diff
@@ -135,7 +135,7 @@ func (w *FileWriter) Flush() error {
 	if w.out != nil {
 		return w.out.Flush()
 	}
-	return nil
+	return errors.Errorf("no underlying writer opened")
```
kennytm: Just use `errors.New()` if you aren't formatting anything.

csuzhangxc: my fault, addressed.

pkg/binlog/filename.go
```diff
@@ -0,0 +1,43 @@
...
+	ErrInvalidBinlogFilename = errors.New("invalid binlog filename")
+)
+
+// VerifyBinlogFilename verify whether is a valid MySQL/MariaDB binlog filename.
+// valid format is `base + '.' + seq`.
+func VerifyBinlogFilename(filename string) error {
+	parts := strings.Split(filename, binlogFilenameSep)
+	if len(parts) != 2 {
+		return errors.Annotatef(ErrInvalidBinlogFilename, "filename %s", filename)
+	} else if _, err := strconv.Atoi(parts[1]); err != nil {
```
amyangfei: seems not check binlog filename like `mysql-bin.-1000`

csuzhangxc: good catch! addressed.

relay/writer/file.go
csuzhangxc: changed to use `Mutex` instead of `RWMutex`

relay/writer/file.go
GregoryIan: please add a comment

relay/writer/file.go
```diff
@@ -0,0 +1,381 @@
...
+
+	// 3. write the dummy event
+	err = w.out.Write(dummyEv.RawData)
+	return errors.Trace(err)
+}
+
+// handleDuplicateEventsExist tries to handle a potential duplicate event in the binlog file.
+func (w *FileWriter) handleDuplicateEventsExist(ev *replication.BinlogEvent) (*Result, error) {
+	filename := filepath.Join(w.cfg.RelayDir, w.filename.Get())
+	duplicate, err := checkIsDuplicateEvent(filename, ev)
```
GregoryIan: is it inefficient that call it for every event?

relay/writer/file.go
csuzhangxc: there is a comment below this line.

relay/writer/file.go
csuzhangxc: in most cases, `checkIsDuplicateEvent` only checks start/end pos with the file size (has a comment in `checkIsDuplicateEvent`), so I think it is not inefficient. and if we want to make it work correctly, we must do this check.

pkg/binlog/writer/file.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+}
+
+// Start implements Writer.Start.
+func (w *FileWriter) Start() error {
+	w.mu.Lock()
+	defer w.mu.Unlock()
+
+	if w.stage != common.StageNew {
+		return errors.Errorf("stage %s, expect %s, already started", w.stage, common.StageNew)
```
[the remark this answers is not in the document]
csuzhangxc: currently, yes.

relay/writer/file_util.go
GregoryIan: ok

relay/writer/file.go
GregoryIan: I'm worrid that `os.Stat` is an inefficient function

csuzhangxc: so, pass file size into the function as an arg?

csuzhangxc: for `os.Stat`, it will call `syscall.Stat`; for `File.Stat`, it will call `syscall.Fstat`.

from <http://man7.org/linux/man-pages/man2/stat.2.html>.
> fstat() is identical to stat(), except that the file about which
       information is to be retrieved is specified by the file descriptor
       fd.

if we changed to pass `File.Stat()` into the function, the difference of efficiency is small? or should we need to pass an in-memory offset into the function? @GregoryIan what's your opinion.

GregoryIan: I think we can choose one
* check duplicate event only if we need to check, like `holeSize < 0`
* or parse a `fd` or `stat information` as arg?

relay/writer/file_util.go
```diff
@@ -0,0 +1,262 @@
...
+		} else if (e.Header.LogPos - e.Header.EventSize) != uint32(fileHeaderLen) {
+			return errors.Errorf("wrong offset %d for FormatDescriptionEvent, should be %d", e.Header.LogPos, fileHeaderLen)
+		}
+		found = true
+		return nil
+	}
+
+	// only parse single event
+	eof, err := replication.NewBinlogParser().ParseSingleEvent(f, onEventFunc)
+	if err != nil {
```
GregoryIan: if some events after `format description event` are damaged, but `found == true`, should we return true?

relay/writer/file.go
csuzhangxc: I choose the first one.

relay/writer/file_util.go
csuzhangxc: I changed to return `true` even `err != nil` in https://github.com/pingcap/dm/pull/117/commits/6d54140244c2bbf5adc5087948505703e81f14a3, because this function only checks if `FormatDescriptionEvent` exists.

GregoryIan: make sense

GregoryIan:
Rest LGTM

relay/writer/file.go
```diff
@@ -216,43 +216,57 @@ func (w *FileWriter) handleFormatDescriptionEvent(ev *replication.BinlogEvent) (
 // handle RotateEvent:
 //   1. update binlog filename if needed
 //   2. write the RotateEvent if not fake
+// NOTE: we only see fake event for RotateEvent in MySQL source code,
+//       if see fake event for other event type, then handle them.
```
amyangfei: shall we add some log if `handlePotentialHoleOrDuplicate` in not expected event type handler

csuzhangxc: add some log for `handleFileHoleExist` and `handleDuplicateEventsExist`

PR #120, opened 2019-04-19
dashboard: change dashboard name

csuzhangxc:

now we have two dashboards, then change `DM` to `DM-task` is better.

- change dashboard name from `DM` to `DM-task`.
- also, change name from `DM-instances` to `DM-instance`

### Check List 

Tests 

 - No code

Related changes


GregoryIan:
LGTM

PR #122, opened 2019-04-22
*: upgrade failpoint dep

amyangfei:
one of failpoint commit changes Eval returns sequence, but doesn't synchronize change in runtime, which will lead to following error and CI failure
```
loader/loader.go:140: non-bool ok (type failpoint.Value) used as if condition
loader/loader.go:141: invalid type assertion: val.(int) (non-interface type bool on left)
```

we should unify failpoint to the latest version, both in command line tool and runtime

### Check List 

Tests 

 - Unit test
 - Integration test


amyangfei:
PTAL @GregoryIan @csuzhangxc

PR #123, opened 2019-04-23
syncer: refine drop schema in sharding

amyangfei:
fix drop schema in sharding mode
add schema/table operation test in sharding mode

1. ignore schema only sharding group in dropSchemaInSharding, as we always processing schema leave later
    - if we don't filter schema only group, we will have some redundant sharding leave processing, and we must change `UnpackTableID` to support schema only
2. add integration test

### Check List 

Tests 

 - Integration test


amyangfei:
PTAL @GregoryIan @csuzhangxc

syncer/ddl.go
```diff
@@ -239,6 +239,11 @@ func (s *Syncer) dropSchemaInSharding(sourceSchema string) error {
 	sources := make(map[string][][]string)
 	sgs := s.sgk.Groups()
 	for name, sg := range sgs {
+		if sg.IsSchemaOnly {
+			// in sharding group leave handling, we always process schema group,
```
GregoryIan:
```suggestion
			// in sharding group leave handling, we always don't process schema group,
```
?

GregoryIan:
well done! LGTM, and we can check count of data > 7000 is not changed at last

syncer/ddl.go
amyangfei: https://github.com/pingcap/dm/blob/eabc76e625b9a67e37dc3c1493824bf2d690435a/syncer/sharding_group.go#L419-L437
`process schema group` means we remove sources pair for schema group here

GregoryIan: lol, so long a way

amyangfei:
> well done! LGTM, and we can check count of data > 7000 is not changed at last

add checksum validation for data > 7000

GregoryIan:
LGTM

PR #125, opened 2019-04-24
tests: refine process exit path in failpoint test

amyangfei:
In some failpoint injection, we `panic` or `os.Exit(1)` directly, which will lose coverage statistics of the injection code

What we want to achieve is:
1. we have a way to inject error, and the code after injection must be not executed. This ensures the logical correctness.
2. the gotest coverage should collect the execute record of our injection code. This ensures coverage accuracy.

1. run `main` function in a separate routine, currently apply change to `DM-worker.test` only
2. add a mocked `utils.OsExit` function to simulate running process terminate

### Check List 

Tests 

 - Integration test


cmd/dm-master/main_test.go
```diff
@@ -32,6 +34,12 @@ func TestRunMain(t *testing.T) {
 		}
 	}
 
+	defer func() {
+		if err := recover(); err != nil {
```
csuzhangxc: if we `panic` in a separate goroutine, this `recover` in the main goroutine cannot recover it.
Can this achieve our goals?

amyangfei: yep, this still need to solve, I will dig into it later.

GregoryIan:
any update?

cmd/dm-worker/main_test.go
```diff
@@ -32,6 +40,25 @@ func TestRunMain(t *testing.T) {
...
+	go func() {
+		main()
+		close(waitCh)
+	}()
+
+	select {
+	case <-waitCh:
+		return
+	case <-exit:
+		return
```
csuzhangxc: so the method is to exit from this _main_ routine to terminal the separate `main()` routine (and all other routines)?

loader/loader.go
```diff
@@ -141,7 +141,7 @@ func (w *Worker) run(ctx context.Context, fileJobQueue chan *fileJob, workerWg *
 					threshold, _ := val.(int)
 					if job.offset >= int64(threshold) {
 						log.Warnf("[failpoint] load offset %d execeeds threshold %d, will panic", job.offset, threshold)
-						panic("LoadExceedOffsetExit")
+						utils.OsExit(1)
```
csuzhangxc: can `utils.OsExit` carry some msg like `LoadExceedOffsetExit`?

cmd/dm-worker/main_test.go
amyangfei: yes. (to achieve the goal-2, `main` routine must exit normally)

loader/loader.go
amyangfei: The error message is not necessary, as we can log enough information before `utils.OsExit`, besides the function type `OsExit func(int)` is more natural to an `os.Exit` mock

GregoryIan:
LGTM

PR #127, opened 2019-04-24
tests: add process unit periodic status test

amyangfei:

add load/sync periodic status test

1. inject failpoint to control print status interval
2. slow down some data processing, in order DM-worker has enough time to print status
3. replace check port with check rpc request, which is more robust for DM-worker/DM-master available detection

### Check List 

Tests 

 - Integration test


amyangfei:
PTAL @GregoryIan @csuzhangxc

loader/loader.go
```diff
@@ -145,6 +145,11 @@ func (w *Worker) run(ctx context.Context, fileJobQueue chan *fileJob, workerWg *
 					}
 				})
 
+				failpoint.Inject("LoadDataSlowDown", func(val failpoint.Value) {
```
lonng: It's better to change this to `failpoint.Inject("LoadDataSlowDown", nil)`, and modify the failpoint expression to `sleep(1000)`, e.g

`GO_FAILPOINTS=github.com/pingcap/dm/LoadDataSlowDown=sleep(1000)`

syncer/syncer.go
```diff
@@ -1753,6 +1759,12 @@ func (s *Syncer) genRouter() error {
 func (s *Syncer) printStatus(ctx context.Context) {
 	defer s.wg.Done()
 
+	failpoint.Inject("PrintStatusCheckSeconds", func(val failpoint.Value) {
+		seconds, _ := val.(int)
```
lonng: The `val` will be `struct{}{}` if the failpoint expression is `return` which without any value and seconds will be 0. Two ways to fix it:

- type assert directly and make it panic if failpoint expression is unexpected.
- set a default value for statusTime if `!ok.`

loader/loader.go
amyangfei: thanks, addressed

loader/loader.go
```diff
@@ -145,10 +145,7 @@ func (w *Worker) run(ctx context.Context, fileJobQueue chan *fileJob, workerWg *
 					}
 				})
 
-				failpoint.Inject("LoadDataSlowDown", func(val failpoint.Value) {
-					ms, _ := val.(int)
-					time.Sleep(time.Duration(ms) * time.Millisecond)
-				})
+				failpoint.Inject("LoadDataSlowDown", func() {})
```
lonng: It's better to user nil closure because of empty function will cause line number inconsistent. e.g.

```go
failpoint.Inject("LoadDataSlowDown", func() {})
// GENERATE
if _, ok := failpoint.Eval(_curpkg_("LoadDataSlowDown")); ok {
}

// TO KEEP LINE NUMBER CONSISTENT
failpoint.Inject("LoadDataSlowDown", nil)
// GENERATE
failpoint.Eval(_curpkg_("LoadDataSlowDown"))
```

syncer/syncer.go
```diff
@@ -1127,11 +1127,7 @@ func (s *Syncer) Run(ctx context.Context) (err error) {
 		}
 		s.binlogSizeCount.Add(int64(e.Header.EventSize))
 
-		failpoint.Inject("ProcessBinlogSlowDown", func(val failpoint.Value) {
-			ms, _ := val.(int)
-			log.Infof("process binlog slow down: %s", time.Now())
-			time.Sleep(time.Duration(ms) * time.Millisecond)
-		})
+		failpoint.Inject("ProcessBinlogSlowDown", func() {})
```
lonng: (ditto)

tests/print_status/run.sh
```diff
@@ -0,0 +1,68 @@
...
+
+cur=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
+source $cur/../_utils/test_prepare
+WORK_DIR=$TEST_DIR/$TEST_NAME
+
+function run() {
+    run_sql_file $cur/data/db.prepare.sql $MYSQL_HOST1 $MYSQL_PORT1
+
+    # in load stage, the dumped file split into 14 insert segments, we slow down 14 * 100 ms
+    # in sync stage, there are approximately 530~550 binlog events, we slow down 3 * 530 ms
```
csuzhangxc:
```suggestion
    # in sync stage, there are approximately 530~550 binlog events, we slow down 530 * 3 ms
```
or update L12 to make them unified?

tests/print_status/run.sh
```diff
@@ -0,0 +1,68 @@
...
+    [ $status_count -ge 2 ]
+    count=0
+    cat $status_file|while read -r line; do
+        total_file_count=$(echo "$line"|awk '{print $(NF-4)}'|tr -d ",")
+        [ $total_file_count -eq 3 ]
+        ((count+=1))
+        if [ $count -eq $status_count ]; then
+            finished_bytes=$(echo "$line"|awk '{print $3}'|tr -d ",")
+            total_bytes=$(echo "$line"|awk '{print $6}'|tr -d ",")
+            [[ "$finished_bytes" -eq "$total_bytes" ]]
```
csuzhangxc: 👍

GregoryIan:
LGTM

PR #133, opened 2019-05-06
dmctl: remove worker mode and offline mode

amyangfei:

dmctl supports master mode only

remove code related to worker mode and offline mode

### Check List 

Tests 

 - Unit test
 - Integration test


amyangfei:
PTAL @GregoryIan @csuzhangxc

PR #139, opened 2019-05-08
test: use rpc way to check port

amyangfei:
we introduce rpc check in pr #127, but some tests miss the change

use rpc available check in all the integration tests

### Check List 

Tests 

 - Integration test


amyangfei:
PTAL @GregoryIan @csuzhangxc

PR #143, opened 2019-05-10
tests: fix print_status check, in some case log may not flushed to disk

amyangfei:
fix occasional CI failed, such as https://internal.pingcap.net/idc-jenkins/blue/organizations/jenkins/dm_ghpr_test/detail/dm_ghpr_test/726/pipeline

after dig into log, I find syncer log is not flushed to disk during test running in time.

1. disable heartbeat in `print_status` test, so we will have determined binlog event counts
2. add retry mechanism in syncer print status check
3. **update DM-worker, DM-master startup and check rpc available sequence** to following
```
start DM-worker1
check DM-worker1 rpc service is available
start DM-worker2
check DM-worker2 rpc service is available
start DM-server
check DM-server rpc service is available
```

btw, why there are 92 binlog events in test case but there exists 94 SQL statements, this is because the following two statements change zero rows.
```
UPDATE `t_1` SET `c12` = 'a' WHERE `id` = 50;
UPDATE `t_1` SET `c12` = 'c' WHERE `id` = 87;
```

### Check List 

Tests 

 - Integration test

### DIFF
--- a/tests/print_status/run.sh
+++ b/tests/print_status/run.sh
@@ -10,11 +10,14 @@ function run() {
     run_sql_file $cur/data/db.prepare.sql $MYSQL_HOST1 $MYSQL_PORT1
 
     # in load stage, the dumped file split into 14 insert segments, we slow down 14 * 100 ms
-    # in sync stage, there are approximately 530~550 binlog events, we slow down 530 * 3 ms
+    # in sync stage, there are 92 group of binlog events, including an XIDEvent,
+    # TableMapEvent, QueryEvent, GTIDEvent, and a specific Event in each group.
+    # so we slow down 460 * 4 ms. Besides the log may be not flushed to disk asap,
+    # we need to add some retry mechanism
     inject_points=("github.com/pingcap/dm/loader/PrintStatusCheckSeconds=return(1)"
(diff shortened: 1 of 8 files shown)


csuzhangxc:
@amyangfei is this PR ready for review now.

amyangfei:
> @amyangfei is this PR ready for review now.

it is ready for review now, but I will make several times of `run-all-test` to see if result is stable

amyangfei:
PTAL @GregoryIan @csuzhangxc

tests/print_status/run.sh
```diff
@@ -50,6 +52,23 @@ function run() {
     check_sync_diff $WORK_DIR $cur/conf/diff_config.toml
 
     # check sync unit print status
+    i=0
+    while [ $i -lt 3 ]
+    do
+        event_count=$(grep "receive binlog event with header" $WORK_DIR/worker1/log/dm-worker.log|grep -v relay|grep "XIDEvent"|wc -l)
```
csuzhangxc: grep from _debug_ level log?

amyangfei: yep, the log level in integration test is DEBUG

csuzhangxc: OK. but notice the change of the log.

GregoryIan:
LGTM

PR #146, opened 2019-05-14
tests: add codecov config

amyangfei:
changed code may be not always able to covered by unit test

add codecov config, mainly copy config from TiDB repo
pay attention to two fields
```
+  status:
+    project:
+      default:
+        threshold: 3 #Allow the coverage to drop by threshold%, and posting a success status.
+    patch:
+      default:
+        target: 0%
+    changes: no
```

### Check List 

Tests 

 - Unit test

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -31,6 +31,7 @@ comment:
 ignore:
   - "LICENSES"
   - "*_test.go"
+  - "*.pb.go"
   - ".git"
   - "*.yaml"
   - "*.toml"
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,39 @@
+codecov:
+  notify:
+    require_ci_to_pass: yes
+
+coverage:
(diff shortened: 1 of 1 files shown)


amyangfei:
PTAL @GregoryIan @csuzhangxc

GregoryIan:
LGTM

.codecov.yml
```diff
@@ -0,0 +1,39 @@
...
+      macro: no
+
+comment:
+  layout: "header, diff"
+  behavior: default
+  require_changes: no
+
+ignore:
+  - "LICENSES"
+  - "*_test.go"
```
csuzhangxc: whether needing to ignore ".pb.go" files?

amyangfei: yep, can also add.
but we have already removed these files in build scripts

amyangfei: added

PR #151, opened 2019-05-17
*: refine log of checker

GregoryIan:

Eliminate some doubtful logs in checker

add more information into error and log

### Check List 

Tests 

 - No code

### DIFF
--- a/checker/checker.go
+++ b/checker/checker.go
@@ -122,7 +122,7 @@ func (c *Checker) Init() (err error) {
 		}
 		instance.sourceDB, err = dbutil.OpenDB(*instance.sourceDBinfo)
 		if err != nil {
-			return errors.Trace(err)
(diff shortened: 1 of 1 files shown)


checker/checker.go
```diff
@@ -122,7 +122,7 @@ func (c *Checker) Init() (err error) {
 		}
 		instance.sourceDB, err = dbutil.OpenDB(*instance.sourceDBinfo)
 		if err != nil {
-			return errors.Trace(err)
+			return errors.Annotatef(err, "failed to open source DSN %s:***@%s:%d", instance.cfg.From.User, instance.cfg.From.Host, instance.cfg.From.Port)
```
amyangfei: shall we differ use and no password scenario, it is more explicit. (although we can detect whether password is used in DSN from the `err`)
if password is not empty, annotate error with `user:***@host:port` ,
without password, annotate error with `user@host:port`

GregoryIan: I think no need

amyangfei: ok

PR #152, opened 2019-05-17
pprof: enable pprof in dm-master and dm-worker

amyangfei:

enable pprof in dm-master, dm-worker, dm-tracer server

add pprof route

### Check List 

Tests 

 - Unit test
 - Integration test

### DIFF
--- a/dm/master/status.go
+++ b/dm/master/status.go
@@ -16,6 +16,7 @@ package master
 import (
 	"net"
 	"net/http"
+	"net/http/pprof"
 
 	"github.com/pingcap/dm/pkg/log"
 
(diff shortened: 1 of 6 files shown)


amyangfei:
PTAL @GregoryIan @WangXiangUSTC

dm/master/status.go
```diff
@@ -39,6 +40,12 @@ func (h *statusHandler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 func InitStatus(lis net.Listener) {
 	mux := http.NewServeMux()
 	mux.Handle("/status", &statusHandler{})
+	mux.HandleFunc("/debug/pprof/", pprof.Index)
+	mux.HandleFunc("/debug/pprof/cmdline", pprof.Cmdline)
+	mux.HandleFunc("/debug/pprof/profile", pprof.Profile)
+	mux.HandleFunc("/debug/pprof/symbol", pprof.Symbol)
+	mux.HandleFunc("/debug/pprof/trace", pprof.Trace)
```
kennytm: Why we need to handle these four separately ?

amyangfei: These four are not builtin profile, and their handler must be added to url route if we want to use them.

dm/master/status.go
```diff
@@ -39,6 +40,12 @@ func (h *statusHandler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 func InitStatus(lis net.Listener) {
 	mux := http.NewServeMux()
 	mux.Handle("/status", &statusHandler{})
+	mux.HandleFunc("/debug/pprof/", pprof.Index)
```
csuzhangxc: wo~~ the pprof was disabled because we introduced `mux` instead of the original builtin _http_, I think you can remove `_ "net/http/pprof"` in 3 _main.go_ now.

GregoryIan:
LGTM

PR #156, opened 2019-05-24
*: refactor log of dmctl

GregoryIan:
refactor log of dmctl

use pingcap/log repo

### DIFF
--- a/checker/checker.go
+++ b/checker/checker.go
@@ -23,6 +23,8 @@ import (
 	"sync"
 	"time"
 
+	"go.uber.org/zap"
+
 	_ "github.com/go-sql-driver/mysql" // for mysql
 	"github.com/pingcap/errors"
 	"github.com/pingcap/tidb-tools/pkg/check"
@@ -54,6 +56,8 @@ type mysqlInstance struct {
 type Checker struct {
 	closed sync2.AtomicBool
 
+	logger log.Logger
+
 	instances []*mysqlInstance
 
 	checkList     []check.Checker
@@ -69,6 +73,7 @@ func NewChecker(cfgs []*config.SubTaskConfig, checkingItems map[string]string) *
 	c := &Checker{
 		instances:     make([]*mysqlInstance, 0, len(cfgs)),
 		checkingItems: checkingItems,
+		logger:        log.With(zap.String("unit", "checker")),
 	}
 
 	for _, cfg := range cfgs {
@@ -204,7 +209,7 @@ func (c *Checker) Init() (err error) {
 		}
 	}
 
-	log.Infof(c.displayCheckingItems())
+	c.logger.Info(c.displayCheckingItems())
 	return nil
 }
 
@@ -272,14 +277,14 @@ func (c *Checker) closeDBs() {
 	for _, instance := range c.instances {
 		if instance.sourceDB != nil {
 			if err := dbutil.CloseDB(instance.sourceDB); err != nil {
-				log.Errorf("close source db %+v error %v", instance.sourceDBinfo, err)
+				c.logger.Error("close source db", zap.Stringer("db", instance.sourceDBinfo), log.ShortError(err))
 			}
 			instance.sourceDB = nil
 		}
 
 		if instance.targetDB != nil {
 			if err := dbutil.CloseDB(instance.targetDB); err != nil {
-				log.Errorf("close target db %+v error %v", instance.targetDBInfo, err)
+				c.logger.Error("close target db", zap.Stringer("db", instance.targetDBInfo), log.ShortError(err))
 			}
 			instance.targetDB = nil
 		}
@@ -289,7 +294,7 @@ func (c *Checker) closeDBs() {
 // Pause implements Unit interface
 func (c *Checker) Pause() {
 	if c.closed.Get() {
-		log.Warn("[checker] try to pause, but already closed")
+		c.logger.Warn("try to pause, but already closed")
 		return
 	}
 }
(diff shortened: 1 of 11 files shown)


pkg/log/log.go
```diff
@@ -14,196 +14,111 @@
...
-			if k != "file" && k != "line" {
-				fmt.Fprintf(b, " %v=%v", k, v)
-			}
-		}
-	}
-
-	b.WriteByte('\n')
-
-	return b.Bytes(), nil
+// L returns the current logger for Lightning.
```
kennytm: 🙃

csuzhangxc:
@GregoryIan do not need to care test cases failed caused by _log_ now?

GregoryIan:
@csuzhangxc no need, it must fail

pkg/log/log.go
GregoryIan: really don't want to import lightning repo, it would be nice if It was in a public library.

kennytm: Yes we could copy the common parts into tidb-tools

checker/checker.go
```diff
@@ -289,15 +294,15 @@ func (c *Checker) closeDBs() {
...
+		c.logger.Warn("try to pause, but already closed")
 		return
 	}
 }
 
 // Resume resumes the paused process
 func (c *Checker) Resume(ctx context.Context, pr chan pb.ProcessResult) {
 	if c.closed.Get() {
-		log.Warn("[checker] try to resume, but already closed")
+		c.logger.Warn("[checker] try to resume, but already closed")
```
csuzhangxc:
```suggestion
		c.logger.Warn("try to resume, but already closed")
```

cmd/dm-ctl/main.go
```diff
@@ -48,8 +48,14 @@ func main() {
 
 	// now, we use checker in dmctl while it using some pkg which log some thing when running
 	// to make dmctl output more clear, simply redirect log to file rather output to stdout
-	log.SetLevelByString("info")
-	log.SetOutputByName("dmctl.log")
+	err = log.InitLogger(&log.Config{
+		File:  "dmctl.log",
```
csuzhangxc: is it needed to _Sync_ log to file?

checker/checker.go
```diff
@@ -23,6 +23,8 @@ import (
 	"sync"
 	"time"
 
+	"go.uber.org/zap"
+
```
csuzhangxc: no extra blank line needed? ditto for other files.

pkg/parser/common.go
```diff
@@ -38,13 +40,12 @@ var (
 // Parse wraps parser.Parse(), makes `parser` suitable for dm
 func Parse(p *parser.Parser, sql, charset, collation string) (stmt []ast.StmtNode, err error) {
 	stmts, warnings, err := p.Parse(sql, charset, collation)
-
-	for _, warning := range warnings {
-		log.Warnf("parsing sql %s:%v", sql, warning)
+	if err != nil {
+		log.L().Error("parse statement", zap.String("sql", sql), zap.Reflect("error message", err))
```
csuzhangxc: can we use `zap.Error(err)` or `log.ShortError(err)` directly here?

pkg/utils/common.go
```diff
@@ -179,27 +181,27 @@ func querySQL(db *sql.DB, query string, maxRetry int) (*sql.Rows, error) {
...
+			log.L().Warn("query retry", zap.String("query", query), log.ShortError(err))
 			continue
 		}
 
 		return rows, nil
 	}
 
 	if err != nil {
-		log.Errorf("query sql[%s] failed %v", query, errors.ErrorStack(err))
+		log.L().Error("query retry", zap.String("query", query), zap.Error(err))
```
csuzhangxc: is `log.ShortError(err)` better?

GregoryIan: as old log `errors.ErrorStack`, I use `zap.Error`

pkg/parser/common.go
GregoryIan: yep we can, I just want to make it same with warning messages

cmd/dm-ctl/main.go
GregoryIan: I  thought it,  I even think we don't need log here, so I don't add it

GregoryIan:
@csuzhangxc PTAL again

pkg/utils/printer.go
```diff
@@ -16,10 +16,9 @@ package utils
 import (
 	"fmt"
 
+	"github.com/pingcap/dm/pkg/log"
 	"go.uber.org/zap"
 	"go.uber.org/zap/zapcore"
-
-	"github.com/pingcap/dm/pkg/log"
```
csuzhangxc: we can keep this _import_ in a separate group here.

pkg/utils/common.go
```diff
@@ -179,27 +178,27 @@ func querySQL(db *sql.DB, query string, maxRetry int) (*sql.Rows, error) {
...
+			log.L().Warn("query retry", zap.String("query", query), log.ShortError(err))
 			continue
 		}
 
 		return rows, nil
 	}
 
 	if err != nil {
-		log.Errorf("query sql[%s] failed %v", query, errors.ErrorStack(err))
+		log.L().Error("query retry", zap.String("query", query), log.ShortError(err))
```
amyangfei: here is query failed at last, no more retry.
btw the logic is redundant, the `err` is always not nil in L196

PR #158, opened 2019-05-24
go.mod: update most dependencies to latest versions

kennytm:

Update go.mod.

Update go.mod, in particular:

* tidb-tools → master, to include pingcap/tidb-tools#236
* tidb → v3.0.0-rc.1
* errors → v0.11.4

Cannot update failpoint since it needs a matching failpoint-ctl version.

Cannot update grpc, maybe caused by grpc/grpc-go#2636 (`TracerTestSuite.TestCollectSyncerEvent` will fail with "failed to connect rpc port for 100 retries in every 10ms")

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes

 - Need to update the documentation (pingcap/docs#1166)


csuzhangxc:
@amyangfei PTAL

PR #160, opened 2019-05-30
syncer: add pipeline struct

WangXiangUSTC:
add a new struct pipeline, will use use this struct to refine syncer

a pipeline will have multiple pipes, every pipe have input and output, pipeline recevie data from syncer, and get output data from the  pipe before, and then send data to the next pipe.

can look at #147 to see how pipeline works


syncer/pipe.go
```diff
@@ -0,0 +1,210 @@
...
+import (
+	"sync"
+
+	"github.com/pingcap/dm/dm/config"
+	"github.com/pingcap/dm/pkg/gtid"
+	"github.com/pingcap/dm/pkg/log"
+	"github.com/pingcap/errors"
+	"github.com/siddontang/go-mysql/mysql"
+	"github.com/siddontang/go-mysql/replication"
+	"github.com/siddontang/go/sync2"
```
csuzhangxc: split these `import` into 3 groups:
1. builtin
2. third party pkgs
3. project pkgs

syncer/pipe.go
```diff
@@ -0,0 +1,210 @@
...
+)
+
+// Pipe is the littlest process unit in syncer
+type Pipe interface {
+	// Name is the pipe's name
+	Name() string
+
+	// Init initializes the pipe
+	Init(*config.SubTaskConfig, func() error) error
```
csuzhangxc: any explanation about this `func() error`?

syncer/pipe.go
```diff
@@ -0,0 +1,210 @@
...
+
+	// Update updates the pipe's config
+	Update(*config.SubTaskConfig)
```
csuzhangxc: whether needing to return an `error`?

syncer/pipe.go
```diff
@@ -0,0 +1,210 @@
...
+
+	// Input receives data
+	Input() chan *PipeData
+
+	// Output exports data already be processed
+	Output() chan *PipeData
```
csuzhangxc: the purpose of defining these two methods (`Input` and `Output`) is to make running pipes concurrently someday?

syncer/pipe.go
```diff
@@ -0,0 +1,210 @@
...
+	Run()
+
+	// Close closes the pipe
+	Close()
+
+	// Wait waits all the data in the pipe is processed
+	Wait()
+
+	// SetErrorChan set the error channel, send error to this channel when meet error
+	SetErrorChan(chan error)
```
csuzhangxc: what is the advantage of using independent `Output` and `SetErrorChan`?

syncer/pipe.go
```diff
@@ -0,0 +1,210 @@
...
+	tp           opType
+	sourceSchema string
+	sourceTable  string
+	targetSchema string
+	targetTable  string
+	sqls         []string
+	args         [][]interface{}
+	keys         [][]string
+	retry        bool
+	pos          mysql.Position
```
csuzhangxc: use a clearer name compare to `currentPos`?

syncer/pipe.go
```diff
@@ -0,0 +1,210 @@
...
+	currentPos   mysql.Position // exactly binlog position of current SQL
+	gtidSet      gtid.Set
```
csuzhangxc: can also update the name to match updated `pos`

syncer/pipe.go
```diff
@@ -0,0 +1,210 @@
...
+	Close()
+
+	// Wait waits all the data in the pipe is processed
+	Wait()
+
+	// SetErrorChan set the error channel, send error to this channel when meet error
+	SetErrorChan(chan error)
+
+	// SetResolveFunc set the resolveFunc, which used when data is resolved, used in the last pipe, or data is skiped
+	SetResolveFunc(func())
```
csuzhangxc: can we do this from the `PipeData` of `Output`?

syncer/pipe.go
```diff
@@ -0,0 +1,210 @@
...
+
+func (p *Pipeline) link(output, input chan *PipeData) {
+	defer p.wg.Done()
+
+	select {
+	case data := <-output:
+		select {
+		case input <- data:
+		case <-p.closeCh:
+			return
```
csuzhangxc: log `data` here because we drop it?

csuzhangxc:
@GregoryIan @amyangfei PTAL

syncer/pipe.go
WangXiangUSTC: `Output` used to send out data be processed success,  `SetErrorChan` used to report error

syncer/pipe.go
WangXiangUSTC: there is a `link` function used to link two pipe, get data from one pipe's output, and then send data to another pipe's input

syncer/pipe.go
WangXiangUSTC: because some data will be filter in pipe, so can't get from output. Or we add a variable in `PipeData` to let pipeline know this data should be filter?

syncer/pipe.go
WangXiangUSTC: done

syncer/pipe.go
WangXiangUSTC: done

syncer/pipe.go
WangXiangUSTC: which name you think is better?

syncer/pipe.go
WangXiangUSTC: done

syncer/pipe.go
WangXiangUSTC: done

syncer/pipe.go
WangXiangUSTC: https://github.com/pingcap/dm/pull/160/commits/a5450bcd2bbc1336d43d6621f757b0f1608b4c53

syncer/pipe.go
csuzhangxc: maybe `currentGTIDSet`? but we do not use GTID set in syncer now.

syncer/pipe.go
csuzhangxc: so, why call this extra init _func_ in the caller?

syncer/pipe.go
csuzhangxc: what is the advantage of using a `chan` rather than just passing/returning as values? speed up the processing?

WangXiangUSTC: every pipe will process data async, this will enhance the performance

syncer/pipe.go
WangXiangUSTC: I think this is also a part of initial work, should do in Init func

syncer/pipe.go
WangXiangUSTC: changed https://github.com/pingcap/dm/pull/160/commits/cffc1644209543d0389898d643181299436be500
keep this variable now?

WangXiangUSTC:
@amyangfei PTAL

amyangfei:
it is better to rebase `master` into `refine_syncer` branch. currently the code change of this pr is hard to distinguish

WangXiangUSTC:
done @amyangfei

syncer/pipe.go
```diff
@@ -0,0 +1,212 @@
...
+// NewPipeline returns a new pipeline
+func NewPipeline() *Pipeline {
+	return &Pipeline{
+		pipes: make([]Pipe, 0, 5),
+	}
+}
+
+// AddPipe adds a pipe to this pipeline
+func (p *Pipeline) AddPipe(pipe Pipe) {
+	if len(p.pipes) == 0 {
```
amyangfei:
```suggestion
	if p.pipes == nil {
```
If the `Pipeline` is created by `NewPipeline`, we can simply append pipe to p.pipes

syncer/pipe.go
```diff
@@ -0,0 +1,212 @@
...
+	if p.isClosed.Get() {
+		return errors.New("pipeline is closed")
+	}
+
+	select {
+	case p.pipes[0].Input() <- data:
+		p.dataWg.Add(1)
+		return nil
+	case <-p.closeCh:
+		return errors.New("pipeline's context is done")
```
amyangfei: create a error object, may be used in other place

syncer/pipe.go
```diff
@@ -0,0 +1,212 @@
...
+	p.wg.Wait()
+
+	for _, pipe := range p.pipes {
+		pipe.Close()
+	}
+
+	close(p.errCh)
+}
+
+// Errors returns a channel for receive this pipeline's error
```
amyangfei:
```suggestion
// Errors returns a channel for receiving this pipeline's error
```

syncer/pipe.go
```diff
@@ -0,0 +1,212 @@
...
+	for _, pipe := range p.pipes {
+		pipe.SetErrorChan(p.errCh)
+		pipe.SetResolveFunc(func() {
+			p.dataWg.Done()
+		})
+		pipe.Run()
+	}
+
+	for i := 0; i < len(p.pipes)-1; i++ {
+		if i+1 < len(p.pipes) {
```
amyangfei: why use the `if condition` here?

WangXiangUSTC: don't need, removed

syncer/pipe.go
WangXiangUSTC: done

WangXiangUSTC:
@csuzhangxc @amyangfei PTAL again

syncer/pipe.go
csuzhangxc: now, one pipe is always bound to a special previous pipe and a special next pipe, right?

syncer/pipe.go
csuzhangxc: keeping it is fine.

syncer/pipe.go
WangXiangUSTC: yes, do you have any suggestions?

syncer/pipe.go
csuzhangxc: IMO, there is an `Init` in `Syncer` which calls all `init`s of Pipes. but maybe your choice is better and we can keep considering this later.

syncer/pipe.go
csuzhangxc: maybe extract data channel from Pipe? one or more Pipe send data into the channel and one or more Pipe receives data from the channel, but we need to keep the order in some cases, this may be not easy. you can keep considering this.

syncer/pipe.go
```diff
@@ -0,0 +1,216 @@
...
+	if p.isClosed.Get() {
+		return errPipelineClose
+	}
+
+	select {
+	case p.pipes[0].Input() <- data:
+		p.dataWg.Add(1)
+		return nil
+	case <-p.closeCh:
+		return errPipelineCtxDone
```
csuzhangxc: no context used in this `Input`, use another name?

syncer/pipe.go
```diff
@@ -0,0 +1,216 @@
...
+func (p *Pipeline) Start() {
+	log.Info("start pipeline")
+
+	p.closeCh = make(chan interface{})
+	p.errCh = make(chan error, 10)
+
+	for _, pipe := range p.pipes {
+		pipe.SetErrorChan(p.errCh)
+		pipe.SetResolveFunc(func() {
+			p.dataWg.Done()
```
csuzhangxc: when do we call this resolveFunc?

syncer/pipe.go
WangXiangUSTC: changed to `errPipelineClose`

syncer/pipe.go
WangXiangUSTC: had a comment for func `SetResolveFunc`: `SetResolveFunc set the resolveFunc, which used when data is resolved, used in the last pipe, or data is skiped`

PR #163, opened 2019-05-30
tests: support run integration tests in parallel

amyangfei:
run integration tests in parallel.
average CI time drop from 9-10min to less than 5min

1. split to multiple integration tests
2. new added integration tests will be run in a single test. If want to split more, change the jenkins pipeline script.

### Check List 

Tests 

 - Integration test

Related changes

 - Need to update jenkins CI pipeline script


amyangfei:
PTAL @GregoryIan @csuzhangxc

PR #171, opened 2019-06-12
relay: use new `Reader`, `Transformer` and `Writer`

csuzhangxc:

this PR is a part of #91.

use new `Reader`, `Transformer` and `Writer` to:
1. read binlog events from upstream
2. transform some binlog events
3. write binlog events into relay log files
4. do recover for relay log file if needed before processing
5. refine some code and add some test cases about this refactoring
	- more test cases for the _interface methods_ can be added later.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

 - Has persistent data change

Side effects


csuzhangxc:
@GregoryIan @amyangfei PTAL

pkg/binlog/reader/tcp_test.go
```diff
@@ -90,10 +90,23 @@ func (t *testTCPReaderSuite) setUpData(c *C) {
 	query := fmt.Sprintf("DROP DATABASE `%s`", dbName)
 	_, err := t.db.Exec(query)
 
-	// delete previous binlog files/events.
-	query = "RESET MASTER"
-	_, err = t.db.Exec(query)
-	c.Assert(err, IsNil)
+	maxRetryCount := 5
+	for i := 0; i < maxRetryCount; i++ {
```
GregoryIan: maybe we can have a function in `pkg/unit`, like 
```
func WaitSomething(backoff int, fn func() bool) bool {
	for i := 0; i < backoff; i++ {
		if fn() {
			return true
		}

		time.Sleep(10 * time.Millisecond)
	}

	return false
}
```

relay/reader/error.go
```diff
@@ -28,3 +28,13 @@ func isIgnorableError(err error) bool {
 	}
 	return false
 }
+
+// isRetryableError checks whether the error is retryable.
+func isRetryableError(err error) bool {
```
GregoryIan: should we rename it to a better name, add some worker like retry from where

relay/reader/reader_test.go
```diff
@@ -60,52 +60,52 @@ func (t *testReaderSuite) TestInterface(c *C) {
...
 		}
 	}()
 	obtained := make([]*replication.BinlogEvent, 0, len(cases))
 	for {
-		ev, err2 := r.GetEvent(ctx)
-		c.Assert(err2, IsNil)
-		obtained = append(obtained, ev)
+		result, err2 := r.GetEvent(ctx)
+		c.Assert(err2, check.IsNil)
+		obtained = append(obtained, result.Event)
```
GregoryIan: `c.Assert(result.Event, NotNil)`

relay/relay.go
```diff
@@ -263,261 +242,180 @@ func (r *Relay) process(parentCtx context.Context) error {
...
-	closeGapSyncer := func() {
-		addRelayFlag = false // reset to false when closing the gap syncer
-		if gapReader != nil {
-			gapReader.Close()
-			gapReader = nil
-		}
-		gapSyncStartPos = nil
-		gapSyncEndPos = nil
+	if r.cfg.EnableGTID && latestPos.Compare(minCheckpoint) <= 0 {
+		log.Warnf("[relay] no corresponding position specified for GTID sets %s, skip recovering", latestGTID)
```
GregoryIan: what's meaning of it?

relay/relay.go
```diff
@@ -263,261 +242,180 @@ func (r *Relay) process(parentCtx context.Context) error {
...
-				if currentGtidSet.Equal(r.gSetWhenSwitch) {
-					addRelayFlag = true
-					log.Infof("[relay] binlog events in the gap for file %s already exists in previous sub directory, adding a LOG_EVENT_RELAY_LOG_F flag to them. current GTID set %s", r.fd.Name(), currentGtidSet)
-				} else if currentGtidSet.Contain(r.gSetWhenSwitch) {
-					// only after the first gap filled, currentGtidSet can be the superset of gSetWhenSwitch
-					log.Infof("[relay] binlog events in the gap for file %s not exists in previous sub directory, no need to add a LOG_EVENT_RELAY_LOG_F flag. current GTID set %s, previous sub directory end GTID set %s", r.fd.Name(), currentGtidSet, r.gSetWhenSwitch)
-				} else {
-					return errors.Errorf("currnet GTID set %s is not equal or superset of previous sub directory end GTID set %s", currentGtidSet, r.gSetWhenSwitch)
-				}
+				return errors.Annotate(err, "save meta after recovered")
```
GregoryIan: add latest position and latest gtids into error?

relay/relay.go
```diff
@@ -263,261 +242,180 @@ func (r *Relay) process(parentCtx context.Context) error {
...
+		rResult, err := reader2.GetEvent(ctx2)
+		cancel2()
 		binlogReadDurationHistogram.Observe(time.Since(readTimer).Seconds())
 
 		if err != nil {
-			switch errors.Cause(err) {
-			case context.Canceled:
+			if rResult.ErrIgnorable {
+				log.Infof("[relay] get ignorable error %v when reading binlog event", err)
 				return nil
```
GregoryIan: `return nil` or `continue`?

relay/relay.go
```diff
@@ -263,261 +242,180 @@ func (r *Relay) process(parentCtx context.Context) error {
...
+			log.Errorf("[relay] update last GTID set to %v error %v", tResult.GTIDSet, err)
+		}
 		if !r.cfg.EnableGTID {
 			// if go-mysql set RawModeEnabled to true
 			// then it will only parse FormatDescriptionEvent and RotateEvent
 			// then check `e.Event.(type)` for `QueryEvent` and `XIDEvent` will never be true
 			// so we need to update pos for all events
 			// and also save pos for all events
-			lastPos.Pos = e.Header.LogPos
+			if e.Header.EventType != replication.ROTATE_EVENT {
```
GregoryIan: we can't swith replication mode, right? like from position to gtid replication

relay/relay.go
```diff
@@ -263,261 +242,180 @@ func (r *Relay) process(parentCtx context.Context) error {
...
+			} else if rResult.ErrRetryable {
+				log.Infof("[relay] get retryable error %v when reading binlog event", err)
 				continue
+			}
+			switch err {
 			case replication.ErrChecksumMismatch:
 				relayLogDataCorruptionCounter.Inc()
 			case replication.ErrSyncClosed, replication.ErrNeedSyncAgain:
-				// do nothing
+				// do nothing, but the error will be returned
```
GregoryIan: also stat it into metric?

relay/relay.go
```diff
@@ -263,261 +242,180 @@ func (r *Relay) process(parentCtx context.Context) error {
...
-			return errors.Errorf("some events missing, current event %+v, lastPos %v, current GTID %v, relay file size %d", e.Header, lastPos, lastGTID, fSize)
 		}
+		relayLogWriteDurationHistogram.Observe(time.Since(writeTimer).Seconds())
 
+		// 4. update meta and metrics
+		needSavePos := tResult.CanSaveGTID
+		lastPos.Pos = tResult.LogPos
+		err = lastGTID.Set(tResult.GTIDSet)
+		if err != nil {
+			log.Errorf("[relay] update last GTID set to %v error %v", tResult.GTIDSet, err)
```
GregoryIan: it's different with 
```
			err = r.meta.Save(lastPos, lastGTID)
			if err != nil {
				return errors.Trace(err)
			}
```
one is lenient, one is strict, can both be lenient？

relay/relay.go
csuzhangxc: if `enable-gtid` is `true`, user can only specify `relay-binlog-gtid` without `relay-binlog-name`, then we do not know which relay file should be recovered (the latest relay file is not accurate because the user can change `relay-binlog-gtid` to skip some relay log files).

pkg/binlog/reader/tcp_test.go
csuzhangxc: done in https://github.com/pingcap/dm/pull/171/commits/6c3fab5c440efc069bd43a66c762a980909b9c2a.

relay/relay.go
csuzhangxc: should `return nil`, `ErrIgnoreable` may not a good name.

relay/relay.go
csuzhangxc: can't switch replication mode online now, but can switch after restarted.

relay/relay.go
csuzhangxc: removed `r.cfg.EnableGTID` check in https://github.com/pingcap/dm/pull/171/commits/201f96028d56a0aaa1c17373381a391bdbe8e75a.

relay/relay.go
csuzhangxc: done in https://github.com/pingcap/dm/pull/171/commits/201f96028d56a0aaa1c17373381a391bdbe8e75a.

relay/relay.go
csuzhangxc: change both to be strict in https://github.com/pingcap/dm/pull/171/commits/201f96028d56a0aaa1c17373381a391bdbe8e75a.

GregoryIan:
Rest LGTM

relay/reader/error.go
csuzhangxc: let `Reader` itself to retry in https://github.com/pingcap/dm/pull/171/commits/071bc72611e21e93629c04f28fb439ccd9cb4f38. keep its name still be `isRetryableError`.

relay/relay.go
csuzhangxc: removed in https://github.com/pingcap/dm/pull/171/commits/071bc72611e21e93629c04f28fb439ccd9cb4f38, let _relay_ to handle it.

relay/relay.go
csuzhangxc: it will be recorded in `relayExitWithErrorCounter`, should we need to add another metric like `relayGoMySQLError`?

pkg/binlog/reader/tcp_test.go
```diff
@@ -90,10 +90,23 @@ func (t *testTCPReaderSuite) setUpData(c *C) {
...
+		// check whether other test cases have wrote any events.
+		time.Sleep(time.Second)
+		_, gs, err2 := utils.GetMasterStatus(t.db, flavor)
+		c.Assert(err2, IsNil)
+		if len(gs.String()) > 0 {
+			return false // some events exist now, try again later.
+		}
+		return true
+	}
+	utils.WaitSomething(backoff, waitTime, waitFn)
```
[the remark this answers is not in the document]
csuzhangxc: `WaitSomething` can only decrease the possibility of errors.
I observed after `RESET MASTER`, other test cases write binlog event, may it because multi cases run in different goroutine at the same time?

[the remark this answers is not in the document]
GregoryIan: yes?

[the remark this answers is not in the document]
csuzhangxc: 1. I observer the fact
2. see <https://stackoverflow.com/questions/24375966/does-go-test-run-unit-tests-concurrently>.

relay/reader/reader.go
```diff
@@ -27,6 +28,16 @@ import (
 	"github.com/pingcap/dm/pkg/log"
 )
 
+const (
+	// event timeout when trying to read events from upstream master server.
+	eventTimeout = 1 * time.Hour
```
GregoryIan: In order to the delay requirements of replication, let it 10 or 1minute?

GregoryIan:
LGTM

relay/reader/reader.go
csuzhangxc: It only takes effect for getting events from _go-mysql_ (the events are already read from upstream server).
In fact, we can even use a `Cancel` rather than a `Timeout`, but use a `Timeout` we can log something to indicate still running.

the timeout between is controlled by `slaveReadTimeout` (default 1m) and `masterHeartbeatPeriod` (default 30s).

GregoryIan: every step will have an impact on the delay, for example, if a master-slave switch happend, there may be no data in 1h, delay would be 1h

csuzhangxc: > every step will have an impact on the delay

agree.

> if a master-slave switch happend

An error will get with small delay

GregoryIan: right, set a reasonable small value and let user or dm know something ASAP

csuzhangxc: changed to `10m` in https://github.com/pingcap/dm/pull/171/commits/0527d1660587a7f35404197d29bfc2650a9d561d.

dm/worker/relay_test.go
```diff
@@ -15,22 +15,119 @@ package worker
...
+}
+
+// InjectInitError injects init error
+func (d *DummyRelay) InjectInitError(err error) {
+	d.initErr = err
+}
+
+// Process implements Process interface
+func (d *DummyRelay) Process(ctx context.Context, pr chan pb.ProcessResult) {
+	select {
```
amyangfei: here we can use a simple channel send/receive instead of select with a single case

pkg/binlog/reader/tcp_test.go
```diff
@@ -90,10 +90,23 @@ func (t *testTCPReaderSuite) setUpData(c *C) {
...
+	waitFn := func() bool {
+		// delete previous binlog files/events. if other test cases writing events, they may be failed.
+		query = "RESET MASTER"
+		_, err = t.db.Exec(query)
+		c.Assert(err, IsNil)
+		// check whether other test cases have wrote any events.
+		time.Sleep(time.Second)
+		_, gs, err2 := utils.GetMasterStatus(t.db, flavor)
+		c.Assert(err2, IsNil)
+		if len(gs.String()) > 0 {
```
amyangfei: `return gs.String() == ""` is simpler?

pkg/utils/util_test.go
```diff
@@ -142,3 +144,32 @@ func (t *testUtilsSuite) TestCompareBinlogPos(c *C) {
...
+		return false
+	}
+	c.Assert(WaitSomething(backoff, waitTime, f1), IsFalse)
+	c.Assert(count, Equals, backoff)
+
+	count = 0 // reset
+	// wait success
+	f2 := func() bool {
+		count++
+		if count >= 5 {
```
amyangfei: `return count >= 5`

relay/relay.go
```diff
@@ -263,261 +238,173 @@ func (r *Relay) process(parentCtx context.Context) error {
...
-					}
-					if tryReSync && r.cfg.EnableGTID && r.cfg.AutoFixGTID {
-						err = r.reSetUpReader(r.syncerCfg)
-						if err != nil {
-							return errors.Annotatef(err, "try auto switch with GTID")
-						}
-						tryReSync = false // do not support repeat try re-sync
-						continue
-					}
+					// TODO: try auto fix GTID, and can support auto switching between upstream server later.
```
amyangfei: do we need to add some log here

relay/relay.go
```diff
@@ -529,135 +416,13 @@ func (r *Relay) process(parentCtx context.Context) error {
...
-	fullPath := path.Join(r.meta.Dir(), filename)
-	fd, err := os.OpenFile(fullPath, os.O_CREATE|os.O_RDWR, 0644)
-	if err != nil {
-		return false, errors.Annotatef(err, "file full path %s", fullPath)
-	}
-	r.fd = fd
-
-	// record current active relay log file, and keep it until newer file opened
-	// when current file's fd closed, we should not reset this, because it may re-open again
-	r.setActiveRelayLog(filename)
```
amyangfei: do we need to call `setActiveRelayLog` somewhere?

dm/worker/relay_test.go
csuzhangxc: refined in https://github.com/pingcap/dm/pull/171/commits/f9b1663d8a755052f0307248dd1f9a62ead84dc5.

pkg/binlog/reader/tcp_test.go
csuzhangxc: done in https://github.com/pingcap/dm/pull/171/commits/f9b1663d8a755052f0307248dd1f9a62ead84dc5.

pkg/utils/util_test.go
csuzhangxc: done in https://github.com/pingcap/dm/pull/171/commits/f9b1663d8a755052f0307248dd1f9a62ead84dc5.

relay/relay.go
csuzhangxc: added some log in https://github.com/pingcap/dm/pull/171/commits/f9b1663d8a755052f0307248dd1f9a62ead84dc5.

relay/relay.go
csuzhangxc: Good catch! add a call to `setActiveRelayLog ` in https://github.com/pingcap/dm/pull/171/commits/f9b1663d8a755052f0307248dd1f9a62ead84dc5.

PR #179, opened 2019-06-19
loader: refine some error messages

amyangfei:

`dump` and `load` unit in DM-worker use different black-white list config,  so dumped files may not contain SQL files we want to use in loader.

The root solution should unify the task config used in `dump` unit and `load` unit, which needs a lot of work and will not be done in this pr.
In this pr we only refine the error messages to help DBA to distinguish different error scenario.

### Check List 

Tests 

    - start a fresh task with `full` mode, set `extra-args: "-m"` config in dump unit. see error message.
    - start a fresh task with `full` mode, dump schema that does not list in load unit do-dbs. see error message.

### DIFF
--- a/loader/loader.go
+++ b/loader/loader.go
@@ -686,7 +686,7 @@ func (l *Loader) prepareDbFiles(files map[string]struct{}) error {
 		return errors.New("invalid mydumper files for there are no `-schema-create.sql` files found")
 	}
 	if len(l.db2Tables) == 0 {
-		return errors.New("no available `-schema-create.sql` files, check mydumper parameters and bwlist config")
+		return errors.New("no available `-schema-create.sql` files, check mydumper parameter matches black-white-list in task config")
 	}
 
 	return nil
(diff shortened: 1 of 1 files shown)


loader/loader.go
```diff
@@ -680,9 +682,12 @@ func (l *Loader) prepareDbFiles(files map[string]struct{}) error {
 		}
 	}
 
-	if len(l.db2Tables) == 0 {
+	if schemaFileCount == 0 {
 		return errors.New("invalid mydumper files for there are no `-schema-create.sql` files found")
 	}
+	if len(l.db2Tables) == 0 {
+		return errors.New("no available `-schema-create.sql` files, check mydumper parameters and bwlist config")
```
csuzhangxc: use name in _task.yaml_ rather than `bwlist`?

amyangfei: changed to `no available `-schema-create.sql` files, check mydumper parameter matches black-white-list in task config`

PR #185, opened 2019-06-25
loader: fix break in forloop and some minor refine

amyangfei:

"break" statement terminates execution of the innermost "for", "switch" or "select" statement. We should cancel execution when `context.Done` is detected.

1. fix a break wrong usage
2. remove some unused code
3. use a range operator to simplify single case in `for { select {} }`

### Check List 

Tests 

 - Unit test
 - Integration test

### DIFF
--- a/loader/loader.go
+++ b/loader/loader.go
@@ -419,15 +419,9 @@ func (l *Loader) Process(ctx context.Context, pr chan pb.ProcessResult) {
 	wg.Add(1)
 	go func() {
 		defer wg.Done()
-		for {
-			select {
-			case err, ok := <-l.runFatalChan:
-				if !ok {
-					return
-				}
-				cancel() // cancel l.Restore
-				errs = append(errs, err)
-			}
+		for err := range l.runFatalChan {
+			cancel() // cancel l.Restore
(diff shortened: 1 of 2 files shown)


amyangfei:
PTAL @GregoryIan @csuzhangxc

loader/loader.go
```diff
@@ -1035,7 +1029,7 @@ func (l *Loader) restoreData(ctx context.Context) error {
 		select {
 		case <-ctx.Done():
 			log.Infof("stop dispatch data file job because %v", ctx.Err())
-			break
+			return nil
```
kennytm: No need to `l.closeFileJobQueue()`?

amyangfei: `l.closeFileJobQueue()` just set `fileJobQueueClosed` flag to `true` and close `fileJob` channel, in this scenario it is not needed, however calling it has no side effect.

We have not made the resource recycle regulation when we meet the context cancelation.
For one thing, if we always check the context before any resource access, we can always avoid invalid resource access. On the other side, active close channel, reset resource to nil is always ok.

WangXiangUSTC:
LGTM

PR #186, opened 2019-06-25
loader: generate db-schema-create.sql file if not exists

WangXiangUSTC:
mydumper will not dump schema create sql sometimes, and then loader will return error.
fix issue https://internal.pingcap.net/jira/browse/TOOL-1316 

create schema create sql file if not find

### Check List 

Tests 

 - Unit test
 - Manual test

Related changes

### DIFF
--- a/loader/loader.go
+++ b/loader/loader.go
@@ -683,10 +683,10 @@ func (l *Loader) prepareDbFiles(files map[string]struct{}) error {
 	}
 
 	if schemaFileCount == 0 {
-		return errors.New("invalid mydumper files for there are no `-schema-create.sql` files found")
+		log.Warn("invalid mydumper files for there are no `-schema-create.sql` files found, and will generate later")
 	}
 	if len(l.db2Tables) == 0 {
-		return errors.New("no available `-schema-create.sql` files, check mydumper parameter matches black-white-list in task config")
+		log.Warn("no available `-schema-create.sql` files, check mydumper parameter matches black-white-list in task config, will generate later")
 	}
 
 	return nil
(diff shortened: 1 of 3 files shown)


loader/util.go
```diff
@@ -73,3 +76,14 @@ func shortSha1(s string) string {
...
 }
+
+func generateSchemaCreateFile(dir string, schema string) error {
+	file, err := os.Create(path.Join(dir, fmt.Sprintf("%s-schema-create.sql", schema)))
+	if err != nil {
+		return errors.Trace(err)
+	}
+	defer file.Close()
+
+	file.WriteString(fmt.Sprintf("CREATE DATABASE `%s`;\n", schema))
```
kennytm: Use `fmt.Fprintf`?

What if `schema` contains a `` ` ``?

loader/loader.go
```diff
@@ -713,7 +713,13 @@ func (l *Loader) prepareTableFiles(files map[string]struct{}) error {
 		}
 		tables, ok := l.db2Tables[db]
 		if !ok {
-			return errors.Errorf("invalid table schema file, cannot find db - %s", file)
+			log.Warnf("cann't find schema create file for db %s, will generate one", db)
```
kennytm:
```suggestion
			log.Warnf("can't find schema create file for db %s, will generate one", db)
```

loader/util.go
amyangfei: also need to check error returned from `fmt.Fprintf` or `file.WriteString`

WangXiangUSTC: addressed in https://github.com/pingcap/dm/pull/186/commits/25659eee5dfb4a70f92960b721fe9d7fac819f15

WangXiangUSTC: do you have any suggest when schema contains ```? @kennytm

kennytm: There should be a function somewhere which replaces `` ` `` with ``` `` ```

WangXiangUSTC: done in https://github.com/pingcap/dm/pull/186/commits/d87c9591b3e8676567cb2997f23cde56b9918707
PTAL again

PR #187, opened 2019-06-25
*: add an upgrade mechanism and fix task meta encoding

csuzhangxc:

restart DM-worker can be failed when having more than 256 task operations

- add an upgrade mechanism
- change encoding for task meta from `LittleEndian` to `BigEndian`
	- when upgrading from an older version, clear previous saved meta

### Check List 

Tests 

 - Unit test
 - Integration test
	 - use an older version of DM-worker to start some tasks
	 - stop DM-worker
	 - replace DM-worker build from this PR
	 - observe whether DM-worker started, task meta cleared and no task auto running

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity


dm/worker/log.go
```diff
@@ -505,3 +525,36 @@ func CloneTaskLog(log *pb.TaskLog) *pb.TaskLog {
 func whetherNil(handler interface{}) bool {
 	return handler == nil || reflect.ValueOf(handler).IsNil()
 }
+
+// clearByPrefix clears all keys with the specified prefix.
+func clearByPrefix(txn *leveldb.Transaction, prefix []byte) error {
```
GregoryIan: why not use `Deleter`

csuzhangxc: `Deleter` can't `write` by _batch_, change to use `Deleter` and remove _batch_?

csuzhangxc:
@GregoryIan @amyangfei PTAL

dm/worker/log.go
GregoryIan: If so, I suggest that you try to unify as much as possible andyou can change the definition of interface. I'm not sure what leveldb's performance would be in bulk when deleting most of the data in a transaction.

GregoryIan:
good job!

dm/worker/log.go
```diff
@@ -340,7 +351,8 @@ func (logger *Logger) doGC(db *leveldb.DB, id int64) {
 	iter.Release()
 	err := iter.Error()
 	if err != nil {
-		log.Errorf("[task log gc] query logs from meta error %v", err)
+		log.Errorf("[task log gc] query logs from meta error %v in range [%s(% X), %s(% X))",
```
GregoryIan: what is the use of this log?

csuzhangxc: just log the _range_ if _iter_ error for debugging.

dm/worker/log.go
```diff
@@ -505,3 +527,36 @@ func CloneTaskLog(log *pb.TaskLog) *pb.TaskLog {
...
+				return errors.Annotatef(err, "delete kv with prefix % X until % X", prefix, iter.Key())
+			}
+			log.Infof("[worker log] delete kv with prefix % X until % X", prefix, iter.Key())
+			batch.Reset()
+		}
+	}
+	iter.Release()
+	err = iter.Error()
+	if err != nil {
+		return errors.Annotatef(err, "iterate kv with prefix % X", prefix)
```
GregoryIan: as log above, add range information?

dm/worker/log_test.go
```diff
@@ -69,6 +88,11 @@ func (t *testLog) TestTaskLogKey(c *C) {
 
 	_, err = DecodeTaskLogKey([]byte("xx"))
 	c.Assert(err, ErrorMatches, ".*not valid length data as.*")
+
+	// test compare
+	b23 := EncodeTaskLogKey(23)
+	b534 := EncodeTaskLogKey(534)
```
GregoryIan: :smile_cat:

dm/worker/meta_test.go
```diff
@@ -69,6 +76,16 @@ func testSetUpDB(c *C) (*leveldb.DB, string) {
 		Task:  testTask2Bytes,
 	}
 
+	testTask3Str, err := testTask3.Toml()
```
GregoryIan: why add it?

dm/worker/worker.go
```diff
@@ -80,8 +80,14 @@ func NewWorker(cfg *Config) (*Worker, error) {
 	}
 	w.relayPurger = purger
 
-	// open kv db
+	// try upgrade from an older version
 	dbDir := path.Join(w.cfg.MetaDir, "kv")
+	err = tryUpgrade(dbDir)
+	if err != nil {
+		return nil, errors.Annotatef(err, "try to upgrade from any older version to %d", currentWorkerVersion)
```
GregoryIan: would output null or older version?

dm/worker/upgrade.go
```diff
@@ -0,0 +1,235 @@
...
+	"github.com/pingcap/errors"
+	"github.com/syndtr/goleveldb/leveldb"
+
+	"github.com/pingcap/dm/pkg/log"
+)
+
+const (
+	// The current internal version of DM-worker used when upgrading from an older version, and it's different from the release version.
+	// NOTE: +1 when an incompatible problem is introduced.
+	currentWorkerVersion uint64 = 1
```
GregoryIan: can we use our version like `1.0.1`? we can only support support to upgrade between official version

dm/worker/upgrade.go
```diff
@@ -0,0 +1,235 @@
...
+}
+
+// fromUint64 restores the version from an uint64 value.
+func (v *internalVersion) fromUint64(verNo uint64) {
+	v.no = verNo
+}
+
+// toUint64 converts the version to an uint64 value.
+func (v *internalVersion) toUint64() uint64 {
+	return uint64(v.no)
```
GregoryIan:
```suggestion
	return v.no
```

dm/worker/upgrade.go
```diff
@@ -0,0 +1,235 @@
...
+
+	"github.com/pingcap/dm/pkg/log"
+)
+
+const (
+	// The current internal version of DM-worker used when upgrading from an older version, and it's different from the release version.
+	// NOTE: +1 when an incompatible problem is introduced.
+	currentWorkerVersion uint64 = 1
+	// The default previous internal version of DM-worker if no valid internal version exists in DB before the upgrade.
+	defaultPreviousWorkerVersion uint64 = 0
```
GregoryIan: In fact, our version may not be linear, more like a forked tree, now you should be able to understand why you don't want to do a version upgrade.

dm/worker/meta_test.go
csuzhangxc: add one more task meta to test _delete by batch_ when clearing task meta.

dm/worker/worker.go
csuzhangxc: `currentWorkerVersion` is the hardcoded version of the current binary. old version can't output this log, but new version always has a valid value.

dm/worker/upgrade.go
csuzhangxc: Then after each incompatible problem introduced, a new version needs to be released?
If not, users upgrade from an older version to _latest_ will encounter the problem.
By the way, TiDB is also using this strategy, see https://github.com/pingcap/tidb/blob/cf5f42b8ece0fad2b3f93c6753747b392f33f4c8/session/bootstrap.go#L313.

dm/worker/log.go
csuzhangxc: we can't know the range here, we are clearing all keys with the specified _prefix_, but we don't know the start-key and end-key.

dm/worker/upgrade.go
csuzhangxc: added a _release version_ in https://github.com/pingcap/dm/pull/187/commits/11d9b85209462289794102b09ac643411d990951.

dm/worker/log.go
GregoryIan: what about it?

dm/worker/upgrade.go
```diff
@@ -0,0 +1,232 @@
...
+	currentWorkerInternalNo uint64 = 1
+)
+
+var (
+	// The key used when saving the version of DM-worker
+	dmWorkerVersionKey = []byte("!DM-worker!version")
+	// The default previous version of DM-worker if no valid version exists in DB before the upgrade.
+	defaultPreviousWorkerVersion = version{InternalNo: 0, ReleaseVersion: "None"}
+	// all versions exists in the history.
+	workerVersion1 = version{InternalNo: 1, ReleaseVersion: "v1.0.0-alpha"}
```
GregoryIan:
```suggestion
	workerVersion1 = version{InternalNo: 1, ReleaseVersion: &util.ReleaseVersion}
```
how about it?

dm/worker/upgrade.go
```diff
@@ -0,0 +1,232 @@
...
+)
+
+// The version of DM-worker used when upgrading from an older version.
+type version struct {
+	InternalNo     uint64 `json:"internal-no"`     // internal version number
+	ReleaseVersion string `json:"release-version"` // release version, like `v1.0.0`
```
GregoryIan:
```suggestion
	ReleaseVersion *string `json:"release-version"` // release version, like `v1.0.0`
```

dm/worker/upgrade.go
csuzhangxc: `workerVersion1` should never be updated later, but `&util.ReleaseVersion` always changed for every release.

dm/worker/upgrade.go
csuzhangxc: to save space or what?

dm/worker/upgrade.go
```diff
@@ -0,0 +1,232 @@
...
+	}
+
+	err = h.Put(dmWorkerVersionKey, data, nil)
+	return errors.Annotatef(err, "save version %v into levelDB with key %s", ver, dmWorkerVersionKey)
+}
+
+// tryUpgrade tries to upgrade from an older version.
+func tryUpgrade(dbDir string) error {
+	// 1. check the DB directory
+	fs, err := os.Stat(dbDir)
```
GregoryIan:
```
notExist := false
fs, err := os.Stat(dbDir)
if err != nil {
	if os.IsNotExist(err) {
        notExist = true
	}
	return errors.Annotatef(err, "get stat for %s", dbDir)
} else if !fs.IsDir() { // should be a directory
	return errors.NotValidf("directory %s for DB", dbDir)
}

// 2. open the kv DB
db, err := openDB(dbDir, defaultKVConfig)
if err != nil {
	return errors.Annotatef(err, "open DB for %s", dbDir)
}
defer func() {
	err = db.Close()
	if err != nil {
		log.Errorf("[worker upgrade] close DB fail %v", err)
	}
}()

if notExist {
  ...
}
...
```

dm/worker/worker.go
```diff
@@ -80,8 +80,14 @@ func NewWorker(cfg *Config) (*Worker, error) {
 	}
 	w.relayPurger = purger
 
-	// open kv db
+	// try upgrade from an older version
 	dbDir := path.Join(w.cfg.MetaDir, "kv")
+	err = tryUpgrade(dbDir)
+	if err != nil {
+		return nil, errors.Annotatef(err, "try to upgrade from any older version to %v", newCurrentVersion())
```
GregoryIan: error message or return values can contain upgrade from which version to which, or do we have a log about elsewhere

GregoryIan:
Rest LGTM

dm/worker/log.go
csuzhangxc: added two methods (`Write`, `NewIterator`) into `Putter`, `Getter` and `Deleter` in https://github.com/pingcap/dm/pull/187/commits/1cf143fd09ddd4d274a8b20a4656d31b71ae881b.

dm/worker/worker.go
csuzhangxc: older version information (which one/whether exist) is included in `err`.

dm/worker/upgrade.go
csuzhangxc: done in https://github.com/pingcap/dm/pull/187/commits/4de40dee717712be8411e05d89c3cde50f441ee1.

csuzhangxc:
@GregoryIan PTAL again

dm/worker/upgrade.go
```diff
@@ -0,0 +1,228 @@
...
+// loadVersion loads the version of DM-worker from the levelDB.
+func loadVersion(h Getter) (ver version, err error) {
+	if whetherNil(h) {
+		return ver, errors.Trace(ErrInValidHandler)
+	}
+
+	data, err := h.Get(dmWorkerVersionKey, nil)
+	if err != nil {
+		if err == leveldb.ErrNotFound {
+			return defaultPreviousWorkerVersion, nil
```
amyangfei: Is it needed to add a log here?
DM-worker may `saveVersion` failed but leave a valid levelDB directory.

dm/worker/upgrade.go
```diff
@@ -0,0 +1,228 @@
...
+	log.Infof("[worker upgrade] the previous version is %v", prevVer)
+
+	// 4. check needing to upgrade
+	currVer := newCurrentVersion()
+	if prevVer.compare(currVer) == 0 {
+		log.Infof("[worker upgrade] the previous and current versions both are %v, no need to upgrade", prevVer)
+		return nil
+	} else if prevVer.compare(currVer) > 0 {
+		log.Warnf("[worker upgrade] the previous version %v is newer than current %v, automatic downgrade is not supported now", prevVer, currVer)
+		return nil
```
amyangfei: is it better to return error here?

dm/worker/upgrade.go
```diff
@@ -0,0 +1,228 @@
...
+// upgradeToVer1 upgrades from version 0 to version 1.
+// before this version, we use `LittleEndian` to encode/decode operation log ID, but it's not correct when scanning operation log by log ID.
+// so, if upgrading from previous version to this one, we need to:
+//  1. remove all operation log in the levelDB
+//  2. reset handled pointer
+//  3. remove all task meta in the levelDB
+// and let user to restart all necessary tasks.
+func upgradeToVer1(db *leveldb.DB) error {
+	log.Infof("[worker upgrade] upgrading to version %v", workerVersion1)
+	txn, err := db.OpenTransaction()
```
GregoryIan: don't use transaction here, use db directly

dm/worker/upgrade.go
csuzhangxc: added in https://github.com/pingcap/dm/pull/187/commits/10281b801665c53de0201cd2cb9a1a82dc2ccec7.

dm/worker/upgrade.go
csuzhangxc: changed in https://github.com/pingcap/dm/pull/187/commits/10281b801665c53de0201cd2cb9a1a82dc2ccec7.

dm/worker/upgrade.go
csuzhangxc: changed in https://github.com/pingcap/dm/pull/187/commits/10281b801665c53de0201cd2cb9a1a82dc2ccec7.

GregoryIan:
LGTM

dm/worker/upgrade.go
```diff
@@ -0,0 +1,214 @@
...
+// loadVersion loads the version of DM-worker from the levelDB.
+func loadVersion(h dbOperator) (ver version, err error) {
+	if whetherNil(h) {
+		return ver, errors.Trace(ErrInValidHandler)
+	}
+
+	data, err := h.Get(dmWorkerVersionKey, nil)
+	if err != nil {
+		if err == leveldb.ErrNotFound {
+			log.Warnf("[worker upgrade] no version found in levelDB, default %v used", defaultPreviousWorkerVersion)
```
amyangfei:
```suggestion
			log.Warnf("[worker upgrade] no version found in levelDB, default %s used", defaultPreviousWorkerVersion)
```

dm/worker/upgrade.go
```diff
@@ -0,0 +1,214 @@
...
+		return errors.Trace(ErrInValidHandler)
+	}
+
+	data, err := ver.MarshalBinary()
+	if err != nil {
+		return errors.Annotatef(err, "marshal version %v to binary data", ver)
+	}
+
+	err = h.Put(dmWorkerVersionKey, data, nil)
+	return errors.Annotatef(err, "save version %v into levelDB with key %s", ver, dmWorkerVersionKey)
```
amyangfei:
```suggestion
	return errors.Annotatef(err, "save version %s into levelDB with key %s", ver, dmWorkerVersionKey)
```
we can change placeholder for `version` to `%s` as it has implemented `Stringer.String` everwhere, although `%v` also finds it implements `Stringer.String` and has the same result as `%s`.
Later changed to zap log we can use `zap.Stringer` directly without any knowlege about what is version

dm/worker/log_test.go
```diff
@@ -188,6 +212,17 @@ func (t *testLog) TestTaskLog(c *C) {
 	c.Assert(logs, DeepEquals, []*pb.TaskLog{taskLog3})
 	c.Assert(logger.handledPointer.Location, Equals, int64(2))
 	c.Assert(logger.endPointer.Location, Equals, int64(4))
+
+	// clear operation log
+	txn, err := db.OpenTransaction()
+	c.Assert(err, IsNil)
+	c.Assert(ClearOperationLog(txn), IsNil)
```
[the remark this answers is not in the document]
csuzhangxc: changed in https://github.com/pingcap/dm/pull/187/commits/eee019ab94b15899638a6efa4e36c4731e294511.

dm/worker/log_test.go
```diff
@@ -309,4 +344,27 @@ func (t *testLog) TestTaskMeta(c *C) {
...
+	// add some task meta again
+	c.Assert(SetTaskMeta(db, testTask1Meta), IsNil)
+	c.Assert(SetTaskMeta(db, testTask2Meta), IsNil)
+	c.Assert(SetTaskMeta(db, testTask3Meta), IsNil)
+
+	// clear task meta
+	GCBatchSize = 2 // < 3
+	txn, err := db.OpenTransaction()
+	c.Assert(err, IsNil)
+	c.Assert(ClearTaskMeta(txn), IsNil)
```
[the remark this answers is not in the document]
csuzhangxc: changed in https://github.com/pingcap/dm/pull/187/commits/eee019ab94b15899638a6efa4e36c4731e294511.

dm/worker/upgrade.go
csuzhangxc: done in https://github.com/pingcap/dm/pull/187/commits/eee019ab94b15899638a6efa4e36c4731e294511.

dm/worker/upgrade.go
csuzhangxc: done in https://github.com/pingcap/dm/pull/187/commits/eee019ab94b15899638a6efa4e36c4731e294511.

dm/worker/upgrade.go
```diff
@@ -0,0 +1,214 @@
...
+	if v.InternalNo < other.InternalNo {
+		return -1
+	} else if v.InternalNo == other.InternalNo {
+		return 0
+	}
+	return 1
+}
+
+// String implements Stringer.String.
+func (v version) String() string {
```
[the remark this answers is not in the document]
csuzhangxc: still, keep using _value receiver_, but we should think more about _value receiver_ and _pointer receiver_ later.

Issue #181, opened 2019-06-21
Not working when use uppercase event in binlog event filter


ixiaohei:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do? If possible, provide a recipe for reproducing the error.
I use dm binlog event filter to filer some binlog event.But i find that it is not working when configuration the uppercase event; Such as use “ALL DML” is not working. But us "all dml" is working
2. What did you expect to see?
Both use "ALL DML" and "all dml" are all working
3. What did you see instead?
I just use "all dml" instead "ALL DML"
4. Versions of the cluster

    - DM version (run `dmctl -V` or `dm-worker -V` or `dm-master -V`):

        ```
Release Version: v1.0.0-alpha-70-g8bfa3e0
Git Commit Hash: 8bfa3e0e99b1bb1d59d9efd6320d9a86fa468217
Git Branch: master
UTC Build Time: 2019-05-06 03:05:07
Go Version: go version go1.12 linux/amd64
        ```

    - Upstream MySQL/MariaDB server version:

        ```
        5.7.20-log
        ```

    - Downstream TiDB cluster version (execute `SELECT tidb_version();` in a MySQL client):

        ```
Release Version: v2.1.5-1-g0bd6b1b
Git Commit Hash: 0bd6b1b60816bf8a2022e657d7a472311548d82e
Git Branch: release-2.1
UTC Build Time: 2019-02-28 08:17:03
GoVersion: go version go1.11.2 linux/amd64
Race Enabled: false
TiKV Min Version: 2.1.0-alpha.1-ff3dd160846b7d1aed9079c389fc188f7f5ea13e
Check Table Before Drop: false
        ```

    - How did you deploy DM: DM-Ansible or manually?

        ```
        DM-Ansible
        ```

amyangfei:
thanks for your report. Currently, DM supports lower case binlog event only and we are considering to make it case insensitive.

amyangfei:
we have supported case insensitive binlog event after this pr https://github.com/pingcap/dm/pull/188. Feel free to try the latest version of DM.

PR #188, opened 2019-06-26
syncer: support case insensitive binlog event in filter

amyangfei:
support case insensitive binlog event in binlog event filter. fix issue https://github.com/pingcap/dm/issues/181

upgrade the version of tidb-tools to this commit https://github.com/pingcap/tidb-tools/tree/16523947c844b1b1aa2996f56dbdc4251bb5c452

### Check List 

Tests 

 - Unit test


WangXiangUSTC:
LGTM

PR #176, opened 2019-06-14
log refactor: dm-worker

GregoryIan:
### What problem does this PR solve? 
refactor log of dm-worker

### What is changed and how it works?
use pingcap/log

PR #194, opened 2019-07-02
tests: Use log.Fatal instead of self made util function

suzaku:

The behavior of `ExitWithError` is very similar to `log.Fatal`.

Remove `ExitWithError` and just use `log.Fatal`.

### Check List 

Tests 

 - Manual test
    I use go run to check the output of the two affected program.
The only difference is that the log level changed from `error` to `fatal`.

Code changes

Side effects

Related changes


kennytm:
IMO the error should be logged as a field using `zap.Error`

suzaku:
> IMO the error should be logged as a field using `zap.Error`

#176 did just that.

kennytm:
🤔 You mean #176 already changed these into `log.Fatal`?

suzaku:
> 🤔 You mean #176 already changed these into `log.Fatal`?

I mean, #176 is the PR that introduce `zap` into `dm`. The `log.Fatal` I'm calling here doesn't support `zap-style` fields.

GregoryIan:
@suzaku please resolve conflicts

tests/_dmctl_tools/check_master_online.go
```diff
@@ -21,20 +21,21 @@ import (
...
 	if err != nil {
-		utils.ExitWithError(err)
+		log.Fatal(errors.ErrorStack(err))
 	}
 	cli := pb.NewMasterClient(conn)
 	req := &pb.ShowDDLLocksRequest{}
 	_, err = cli.ShowDDLLocks(context.Background(), req)
 	if err != nil {
-		utils.ExitWithError(err)
+		log.Fatal(errors.ErrorStack(err))
```
amyangfei: `"github.com/pingcap/dm/pkg/log"` has no `Fatal` method?

PR #196, opened 2019-07-03
Add dm-worker parameters

liubo0127:
Add new parameters for dm-worker

Only the configuration of dm-worker has been modified, which has no effect on user usage.

@csuzhangxc PTAL

### DIFF
--- a/dm/dm-ansible/roles/dm-worker/templates/dm-worker.toml.j2
+++ b/dm/dm-ansible/roles/dm-worker/templates/dm-worker.toml.j2
@@ -20,5 +20,5 @@ port = {{ mysql_port }}
 #relay log purge strategy
 [purge]
 interval = 3600
-expires = 24
+expires = 0
 remain-space = 15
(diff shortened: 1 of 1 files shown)


suzaku:
LGTM

dm/dm-ansible/roles/dm-worker/templates/dm-worker.toml.j2
```diff
@@ -7,9 +7,18 @@ meta-file = "relay.meta"
...
 [from]
 host = "{{ mysql_host }}"
 user = "{{ mysql_user }}"
 password = "{{ mysql_password }}"
 port = {{ mysql_port }}
+
+#relay log purge strategy
+[purge]
+interval = 3600
+expires = 24
```
csuzhangxc: can we set this default value (unit: hour) greater (because we have set `remain-space`)? maybe `1440` (60 * 24)

GregoryIan: ping @liubo0127

liubo0127:
```suggestion
expires = 0
```

Issue #70, opened 2019-03-08
increase DM unit test/integration test code coverage 


amyangfei:
## Feature Request

**Describe the feature you'd like:**

We want to cover most import logic path in DM and increase code coverage as high as possible

**Teachability Strategy:**

We want to finish this task in several steps
- add tests to cover most import logic path, following are tests not include in DM currently 
  - [ ] online DDL test
  - [ ] relay log correctness test
  - [ ] master slave switch test
  - [ ] dmctl API, including many APIs and also some error handling scenario, such as sql-skip/sql-replace, sharding DDL error handling etc.
  - [ ] syncer heartbeat
  - [ ] metric API
  - [ ] periodicity status, such as syncer PrintStatus, loader PrintStatus

- increase code coverage mainly based on coverage of each source file

amyangfei:
integration test not stable case
- [ ] https://internal.pingcap.net/idc-jenkins/blue/organizations/jenkins/dm_ghpr_test/detail/dm_ghpr_test/1810/pipeline

PR #198, opened 2019-07-09
syncer: refine syncer test to trigger casuality

WangXiangUSTC:
issue: https://github.com/pingcap/dm/issues/70
detect casuality conflict and generate flush job test case

refine test, run some sqls which will trigger casuality 

### Check List 

Tests 

 - Unit test

### DIFF
--- a/syncer/syncer_test.go
+++ b/syncer/syncer_test.go
@@ -1128,33 +1128,19 @@ func (s *testSyncerSuite) TestRun(c *C) {
 			"",
 			nil,
 		}, {
+			// in first 5 minutes, safe mode is true, will split update to delete + replace
 			update,
 			"DELETE FROM `test_1`.`t_1` WHERE `id` = ? LIMIT 1;",
 			[]interface{}{int64(580981944116838402)},
 		}, {
+			// in first 5 minutes, , safe mode is true, will split update to delete + replace
 			update,
 			"REPLACE INTO `test_1`.`t_1` (`id`,`name`) VALUES (?,?);",
 			[]interface{}{int64(580981944116838401), "b"},
 		},
 	}
 
-	for _, sql := range sqls1 {
-		c.Log("exec sql: ", sql)
-		_, err := s.db.Exec(sql)
-		c.Assert(err, IsNil)
-	}
-
-	for i := 0; i < 10; i++ {
-		time.Sleep(time.Second)
-
-		testJobs.RLock()
-		jobNum := len(testJobs.jobs)
-		testJobs.RUnlock()
-
-		if jobNum >= len(expectJobs1) {
-			break
-		}
-	}
+	executeSQLAndWait(c, s.db, sqls1, len(expectJobs1))
 
 	testJobs.Lock()
 	checkJobs(c, testJobs.jobs, expectJobs1)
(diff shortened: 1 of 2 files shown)


syncer/syncer_test.go
```diff
@@ -1203,26 +1235,28 @@ func (s *testSyncerSuite) TestRun(c *C) {
...
+
+		c.Assert(job.tp, Equals, expectJobs2[i].tp)
+		if job.tp == ddl {
+			c.Assert(job.ddls[0], Equals, expectJobs2[i].sqlInJob)
 		} else {
-			c.Assert(testJobs.jobs[i].sql, Equals, testCase.sqlInJob)
-			c.Assert(testJobs.jobs[i].args[0], Equals, testCase.arg)
+			c.Assert(job.sql, Equals, expectJobs2[i].sqlInJob)
+			c.Assert(job.args, DeepEquals, expectJobs2[i].args)
 		}
```
suzaku: It seems the execution of SQLs and the validation afterwards are almost identical for sql1 and sql2. Maybe we can extract a function or two to make the test simpler.

WangXiangUSTC: updated https://github.com/pingcap/dm/pull/198/commits/47740ebba6850394b74fddb1558349c147e359b8

suzaku: Can we do the same for the execution and wait part?

syncer/syncer_test.go
```diff
@@ -1074,53 +1084,63 @@ func (s *testSyncerSuite) TestRun(c *C) {
...
-	testCases1 := []struct {
-		sql      string
-		tp       opType
-		sqlInJob string
-		arg      interface{}
-	}{
+	sqls1 := []string{
+		"create database if not exists test_1",
+		"create table if not exists test_1.t_1(id int primary key, name varchar(24))",
+		"create table if not exists test_1.t_2(id int primary key, name varchar(24))",
```
GregoryIan: what does need another table  to test ?

syncer/syncer_test.go
```diff
@@ -1074,53 +1084,63 @@ func (s *testSyncerSuite) TestRun(c *C) {
...
 			"DELETE FROM `test_1`.`t_1` WHERE `id` = ? LIMIT 1;",
-			int64(580981944116838401),
+			[]interface{}{int64(580981944116838401)},
+		}, {
+			flush,
+			"",
+			nil,
+		}, {
+			update,
+			"DELETE FROM `test_1`.`t_1` WHERE `id` = ? LIMIT 1;",
```
GregoryIan: safe mode?

GregoryIan:
Rest LGTM

syncer/syncer_test.go
WangXiangUSTC: yes, I add a comment `in first 5 minutes, safe mode is true, will split update to delete + replace`

syncer/syncer_test.go
WangXiangUSTC: add a new function `executeSQLAndWait`

syncer/syncer_test.go
WangXiangUSTC: because I add `RouteRules` to route `t_1` to `t_2` when update config, so create table `t_2` here

WangXiangUSTC:
@GregoryIan @suzaku PTAL again

PR #199, opened 2019-07-10
*: Use insert for non safe-mode instead of replace

july2993:
*: Use insert for non safe-mode instead of replace

check safe-mode as update event, and use insert for non safe-mode instead of replace

### Check List 

Tests 

Code changes

Side effects

 Related changes


amyangfei:
PTAL @csuzhangxc

PR #203, opened 2019-07-15
Add syncer sharding test

3pointer:
1. Add syncer sharding test to increase syncer test coverage
2. Update tidb-tool go module
3. Update safe_mode/sequence_safe_mode mydumper config in integration test

Add TestSharding function to test
Simulate two tables(t1,t2) in one dm-worker syncer to one target table
```t1 insert -> t2 insert -> t1 add column -> t2 insert -> t2 add column -> t1, t2 insert (new schema)```
and check mock result as wish

### Check List 

Tests 

 - Unit test
 - Integration test

### DIFF
--- a/syncer/syncer_test.go
+++ b/syncer/syncer_test.go
@@ -19,7 +19,6 @@ import (
 	"database/sql/driver"
 	"fmt"
 	"github.com/DATA-DOG/go-sqlmock"
-	"github.com/siddontang/go-mysql/mysql"
 	"strings"
 	"sync"
 	"testing"
@@ -1045,7 +1044,6 @@ func (s *testSyncerSuite) TestSharding(c *C) {
 	testCases := []struct {
 		testSQLs []string
 		expectSQLS []struct{sql string;args []driver.Value}
-		globalCheckPoint mysql.Position
 	} {
 		// case 1:
 		// upstream binlog events:
@@ -1091,7 +1089,6 @@ func (s *testSyncerSuite) TestSharding(c *C) {
 					[]driver.Value{6, 6, "test"},
 				},
 			},
-			mysql.Position{Pos: 11317},
 		},
 		// case 2:
 		// upstream binlog events:
@@ -1142,7 +1139,6 @@ func (s *testSyncerSuite) TestSharding(c *C) {
 					[]driver.Value{6, 6, "test"},
 				},
 			},
-			mysql.Position{Pos: 7006},
 		},
 	}
 
@@ -1260,8 +1256,29 @@ func (s *testSyncerSuite) TestSharding(c *C) {
 		case <-time.After(2 * time.Second):
 		}
 		cancel()
+		// wait for flush finish in Process.Run()
+		// cancel function only closed done channel asynchronously
+		// other goroutine(s.streamer.GetEvent()) received done msg then trigger flush
+		// so the flush action may not execute before we check due to goroutine schedule
+		time.Sleep(time.Second)
 
-		c.Assert(syncer.checkpoint.GlobalPoint().Pos, Equals, _case.globalCheckPoint.Pos)
+		// Xid Event has 31 bytes
+		// check whether last event before flushed globalPoint is Xid event
+		q, err := s.db.Query(fmt.Sprintf("SHOW BINLOG EVENTS FROM %d LIMIT 1", syncer.checkpoint.GlobalPoint().Pos - 31))
+		c.Assert(err, IsNil)
+
+		var pos, endLogPos uint32
+		var eventType string
+		var unusedVal interface{}
+		unused := &unusedVal
+
+		for q.Next() {
+			q.Scan(&unused, &pos, &eventType, &unused, &endLogPos, &unused)
+			break
+		}
+		c.Assert(syncer.checkpoint.GlobalPoint().Pos, Equals, endLogPos)
+		c.Assert(eventType, Equals, "Xid")
+
 		// check expectations for mock db
 		if err := mock.ExpectationsWereMet(); err != nil {
 			c.Errorf("there were unfulfilled expectations: %s", err)
--- a/syncer/syncer_test.go
+++ b/syncer/syncer_test.go
@@ -1046,12 +1046,12 @@ func (s *testSyncerSuite) TestSharding(c *C) {
 		"CREATE TABLE IF NOT EXISTS `stest_1`.`st_2` (id INT, age INT)",
 	}
 
-	testSQLs := []struct{
-		rawSql string
-		expectSql string
-		args []driver.Value
-		skip bool  // skip ddl in sharding
-	} {
+	testSQLs := []struct {
+		rawSQL    string
+		expectSQL string
+		args      []driver.Value
+		skip      bool // skip ddl in sharding
+	}{
 		{
 			"INSERT INTO `stest_1`.`st_1`(id, age) VALUES (1, 1)",
 			"REPLACE INTO",
@@ -1069,7 +1069,7 @@ func (s *testSyncerSuite) TestSharding(c *C) {
 			"ALTER TABLE `stest_1`.`st_1` ADD COLUMN NAME VARCHAR(30)",
 			"ALTER TABLE",
 			[]driver.Value{},
-			true, // skip not active ddl
+			true, // skip not active ddl in mock db
 		},
 		{
 
@@ -1128,7 +1128,7 @@ func (s *testSyncerSuite) TestSharding(c *C) {
 			TargetTable:   "st",
 		},
 	}
-	// easy to mock
+	// set batch to 1 is easy to mock
 	s.cfg.Batch = 1
 	s.cfg.WorkerCount = 1
 	s.cfg.MaxRetry = 1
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Comment by GregoryIan** [comment]
run `go mod tidy`

```diff
diff --git a/go.sum b/go.sum
index a53e85f..7de7852 100644
--- a/go.sum
+++ b/go.sum
@@ -221,12 +221,8 @@ github.com/pingcap/pd v2.1.12+incompatible/go.mod h1:nD3+EoYes4+aNNODO99ES59V83M
 github.com/pingcap/tidb v0.0.0-20190613131440-c59a108c28b7 h1:WzzifUZ1nRsZMijRrLznb2TKh6JYGi/9pgvVT7mnl2w=
 github.com/pingcap/tidb v0.0.0-20190613131440-c59a108c28b7/go.mod h1:c4qUUwEF9VSWePAPB7qFEVdaytJBcc2Tf/yuKKhCErA=
 github.com/pingcap/tidb-tools v2.1.3-0.20190321065848-1e8b48f5c168+incompatible/go.mod h1:XGdcy9+yqlDSEMTpOXnwf3hiTeqrV6MN/u1se9N8yIM=
-github.com/pingcap/tidb-tools v3.0.0-beta.1.0.20190626065910-16523947c844+incompatible h1:kiYtLakPd2GKNHiYnfa6XCNRoRLRLndavuGhA7kBslA=
-github.com/pingcap/tidb-tools v3.0.0-beta.1.0.20190626065910-16523947c844+incompatible/go.mod h1:XGdcy9+yqlDSEMTpOXnwf3hiTeqrV6MN/u1se9N8yIM=
 github.com/pingcap/tidb-tools v3.0.0-beta.1.0.20190709045846-a3fe9b9aaf6f+incompatible h1:+Nr9JhbkVEcEiQYchHex/UlMcurFLsq9P1Ocqe1ItAc=
 github.com/pingcap/tidb-tools v3.0.0-beta.1.0.20190709045846-a3fe9b9aaf6f+incompatible/go.mod h1:XGdcy9+yqlDSEMTpOXnwf3hiTeqrV6MN/u1se9N8yIM=
-github.com/pingcap/tidb-tools v3.0.0+incompatible h1:LbnRRZ9Dyrvopn3aC4HTbqMOWjMOfpZlsAdyz3+VrGk=
-github.com/pingcap/tidb-tools v3.0.0+incompatible/go.mod h1:XGdcy9+yqlDSEMTpOXnwf3hiTeqrV6MN/u1se9N8yIM=
 github.com/pingcap/tipb v0.0.0-20190428032612-535e1abaa330 h1:rRMLMjIMFulCX9sGKZ1hoov/iROMsKyC8Snc02nSukw=
 github.com/pingcap/tipb v0.0.0-20190428032612-535e1abaa330/go.mod h1:RtkHW8WbcNxj8lsbzjaILci01CtYnYbIkQhjyZWrWVI=
 github.com/pkg/errors v0.8.0/go.mod h1:bwawxfHBFNV+L2hUp1rHADufV3IMtnDRdf1r5NINEl0
```

syncer/syncer_test.go
```diff
@@ -1031,6 +1034,196 @@ func (s *testSyncerSuite) TestCasuality(c *C) {
 	wg.Wait()
 }
 
+func (s *testSyncerSuite) TestSharding(c *C) {
+	db, mock, err := sqlmock.New()
+	if err != nil {
+		c.Fatalf("an error '%s' was not expected when opening a stub database connection", err)
+	}
+
+	createSQLs := []string{
```
GregoryIan: clear environment  at first?   `drop database` and `reset master`?

GregoryIan: Ok. I saw them

syncer/syncer_test.go
```diff
@@ -1031,6 +1034,196 @@ func (s *testSyncerSuite) TestCasuality(c *C) {
...
+	mock.ExpectQuery("SHOW COLUMNS").WillReturnRows(
+		sqlmock.NewRows([]string{"Field", "Type", "Null", "Key", "Default", "Extra"}).AddRow("id", "int", "NO", "PRI", null, "").AddRow("age", "int", "NO", "", null, ""))
+	mock.ExpectQuery("SHOW INDEX").WillReturnRows(
+		sqlmock.NewRows([]string{"Table", "Non_unique", "Key_name", "Seq_in_index", "Column_name",
+			"Collation", "Cardinality", "Sub_part", "Packed", "Null", "Index_type", "Comment", "Index_comment"},
+		).AddRow("st", 0, "PRIMARY", 1, "id", "A", 0, null, null, null, "BTREE", "", ""))
+
+	// mock downstream test sql
+	for i, sql := range testSQLs {
+		if sql.skip {
```
GregoryIan: do we need it?

syncer/syncer_test.go
```diff
@@ -1031,6 +1035,245 @@ func (s *testSyncerSuite) TestCasuality(c *C) {
...
+		}()
+
+		select {
+		case r := <-resultCh:
+			c.Assert(len(r.Errors), Equals, 0)
+		case <-time.After(2 * time.Second):
+		}
+		cancel()
+
+		c.Assert(syncer.checkpoint.GlobalPoint().Pos, Equals, _case.globalCheckPoint.Pos)
```
3pointer: @amyangfei After I check the code, I found three ways of flushing the globalPoint
```
1. get flush job
2. get ddl job
3. time.Now() - last globalSavePointTime > 30 * time.Second
```

And in current test case syncer.globalPoint didn't flush after this context canceled when blocking in s.streamer.GetEvent(), so this flushedCheckPoint may not equal to globalPoint. So I only checked globalPoint.Pos.
I'm not sure this is the right way to check the point? Any suggest for check the globalPoint?

amyangfei: 1. when `syncer` unit pauses or stops, it will dispatch flush job and flush checkpoints, we can use this mechanism.
2. we can add some failpoint injection variable to control the flush period.

Besides we can get the binlog position of last XID from the upstream database and compare it with the global point instead of the hard code position

syncer/syncer_test.go
```diff
@@ -1035,68 +1036,113 @@ func (s *testSyncerSuite) TestCasuality(c *C) {
...
 		{
-			"INSERT INTO `stest_1`.`st_1`(id, age) VALUES (1, 1)",
-			"REPLACE INTO",
-			[]driver.Value{1, 1},
-			false,
+			[]string {
+				"INSERT INTO `stest_1`.`st_1`(id, age) VALUES (1, 1)",
+				"INSERT INTO `stest_1`.`st_2`(id, age) VALUES (2, 2)",
+				"ALTER TABLE `stest_1`.`st_1` ADD COLUMN NAME VARCHAR(30)",
+				"INSERT INTO `stest_1`.`st_2`(id, age) VALUES (4, 4)",
```
GregoryIan: add a more case `INSERT INTO stest_1.st_1(id, age) VALUES (?, ?)`

GregoryIan:
Good job! there are only one problem now `how to assert checkpoint`, Rest LGTM

syncer/syncer_test.go
3pointer: What is the proper way to get last XID from upstream database? I tried the `show master status` , retrived the `		syncer.syncer.GetNextPosition()` which is greater than flushedGlobalPoint

amyangfei: It is not easy to get last XID position from upstream directly, but we can query `show binlog events from xxx` from upstream and make sure the global point equals to the last XID position, where `xxx` should be a valid position earlier than `global point`

syncer/syncer_test.go
```diff
@@ -1031,6 +1034,279 @@ func (s *testSyncerSuite) TestCasuality(c *C) {
...
+		var pos, endLogPos uint32
+		var name, eventType string
+		var unusedVal interface{}
+		unused := &unusedVal
+
+		for r.Next() {
+			r.Scan(&name, &pos, &eventType, &unused, &endLogPos, &unused)
+		}
+
+		c.Assert(eventType, Equals, "Xid")
```
3pointer: @amyangfei here check last event before globalPoint is Xid event, please review it

syncer/syncer_test.go
```diff
@@ -1031,6 +1034,279 @@ func (s *testSyncerSuite) TestCasuality(c *C) {
...
+
+		// check expectations for mock db
+		if err := mock.ExpectationsWereMet(); err != nil {
+			c.Errorf("there were unfulfilled expectations: %s", err)
+		}
+		runSQL(dropSQLs)
+		c.Assert(syncer.isClosed(), IsTrue)
```
amyangfei: move this line after `syncer.Close()`

syncer/syncer_test.go
GregoryIan: I think you misunderstand it, do we need to check checkpoint here @amyangfei

amyangfei: we have got the flushed global point, which aspect do you want to check? @GregoryIan

syncer/syncer_test.go
3pointer: Done

GregoryIan:
LGTM

PR #204, opened 2019-07-17
loader: return error if parse mydumper meta data failed

WangXiangUSTC:
fix issue: https://internal.pingcap.net/jira/browse/TOOL-1425
should return error in load unit if parse mydumper metadata failed, otherwise will get error "invalid mysql position string: " in sync unit, this will make user confused.

handle the error for `getMydumpMetadata`

### DIFF
--- a/loader/loader.go
+++ b/loader/loader.go
@@ -421,7 +421,12 @@ func (l *Loader) Process(ctx context.Context, pr chan pb.ProcessResult) {
 	defer cancel()
 
 	l.newFileJobQueue()
-	l.getMydumpMetadata()
+	if err := l.getMydumpMetadata(); err != nil {
+		pr <- pb.ProcessResult{
+			Errors: []*pb.ProcessError{unit.NewProcessError(pb.ErrorType_UnknownError, errors.ErrorStack(err))},
+		}
+		return
+	}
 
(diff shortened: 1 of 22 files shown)


loader/loader.go
```diff
@@ -421,7 +421,12 @@ func (l *Loader) Process(ctx context.Context, pr chan pb.ProcessResult) {
 	defer cancel()
 
 	l.newFileJobQueue()
-	l.getMydumpMetadata()
+	if err := l.getMydumpMetadata(); err != nil {
+		pr <- pb.ProcessResult{
```
amyangfei: `loaderExitWithErrorCounter.WithLabelValues(l.cfg.Name).Inc()`
increase `loaderExitWithErrorCounter` here?

loader/loader.go
```diff
@@ -421,7 +421,12 @@ func (l *Loader) Process(ctx context.Context, pr chan pb.ProcessResult) {
 	defer cancel()
 
 	l.newFileJobQueue()
-	l.getMydumpMetadata()
+	if err := l.getMydumpMetadata(); err != nil {
```
amyangfei: `loaderExitWithErrorCounter.WithLabelValues(l.cfg.Name).Inc()`
increase `loaderExitWithErrorCounter` in this code block?

WangXiangUSTC: updated

PR #216, opened 2019-07-29
*: refactor error system

amyangfei:
### What problem does this PR solve? 

implement error system based on RFC introduced in https://github.com/pingcap/dm/pull/211

### What is changed and how it works?
1. add new error library
2. classify errors and use new error library for all error handling, classification is working in progress:

   - [x] checker
   - [x] relay unit
   - [x] dump unit
   - [x] load unit
   - [x] sync unit
   - [x] dm master
   - [x] dm worker
   - [x] dm pkg

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

 - Has exported variable/fields change

Side effects

Related changes

 - Need to update the documentation
 - Need to be included in the release note

PR #221, opened 2019-07-31
Resolved duplicate entry conflict

rogerszy:
### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes


ericsyh:
@rogerszy Pls run `make fmt` then push the commit.

loader/db.go
```diff
@@ -16,11 +16,14 @@ package loader
...
-	"time"
-
 	"github.com/pingcap/dm/dm/config"
 	tcontext "github.com/pingcap/dm/pkg/context"
 	"github.com/pingcap/dm/pkg/log"
+	"os"
+	"regexp"
+	"strconv"
+	"strings"
+	"time"
```
WangXiangUSTC: we split import to three part:
1. the standard pkg, include these "time", "strings" and so on
2. pkg in dm, like the "github.com/pingcap/dm/dm/config"
3. and other third party pkg

please format this import

WangXiangUSTC:
and need solve the conflicts

WangXiangUSTC:
can you add some details about this pr, for examples, what problem this pr resolved, and how to resolved?

rogerszy:
No details added, just resolved the conflict with other people‘s PR

loader/db.go
rogerszy: "regexp"，"strconv" and "strings" exists in the previous version，I just add a "regexp"

rogerszy:
For this PR, the main problem is about transfering data to TiDB from multiple MySQL instance, an error may occur(ERROR 1062: the Duplicate entry 'XXX' for key 'XXX'), in order to solve the problem, PR processing logic is analyze from SQL file, and delete duplicate data pair that duplicate entry belongs in SQL, then executing SQL again, finally, duplicate data pairs are written into a file for later analyze

rogerszy:
For the logic of deleting duplicate data pair from SQL, first, conflict value and conflict key are obtained from error information, and conflict key is obtained by regular expression, then column position of conflict value is obtained according to conflict key and table structure, finally, the SQL are compared character by character until conflict value is found at corresponding position

loader/db.go
WangXiangUSTC: I means we can move these five to the top, together with "fmt"

loader/db.go
```diff
@@ -155,6 +157,21 @@ func (conn *Conn) executeSQLCustomRetry(ctx *tcontext.Context, sqls []string, en
 
 		if err != nil {
 			tidbExecutionErrorCounter.WithLabelValues(conn.cfg.Name).Inc()
+			dupErr := errors.Cause(err) // Modify the SQL and try again for ErrDupEntry
+			if isMySQLError(dupErr, tmysql.ErrDupEntry) {
+				newSQL, modErr := modifySQL(conn.db, err, sqls[1])
+				if modErr != nil {
+					ctx.L().Info("execute statement", zap.String("modErr", fmt.Sprint(modErr)))
```
WangXiangUSTC:
```suggestion
					ctx.L().Info("execute statement", zap.Err(modErr))
```

loader/db.go
```diff
@@ -273,3 +290,213 @@ func isMySQLError(err error, code uint16) bool {
 	e, ok := err.(*mysql.MySQLError)
 	return ok && e.Number == code
 }
+
+func modifySQL(db *sql.DB, err error, sql string) (string, error) {
+	dupData := ""
+	var table string
```
WangXiangUSTC: we can use same way to define dupData and table

loader/db.go
```diff
@@ -273,3 +290,213 @@ func isMySQLError(err error, code uint16) bool {
...
+	reg := regexp.MustCompile(`^INSERT INTO .*?VALUES`)
+	if reg.MatchString(sql) {
+		tmpTable := reg.FindAllString(sql, 1)
+		table = tmpTable[0][13 : len(tmpTable[0])-8]
+	} else {
+		return sql, errors.New("Can't get table name")
```
WangXiangUSTC: add  sql in the error? will make it easy to debug

loader/db.go
```diff
@@ -273,3 +290,213 @@ func isMySQLError(err error, code uint16) bool {
...
+	logErr := errorLog(errorData, table, dupData, columnName, strconv.Itoa(pos)) // Write dupdata to the errorlog
+	if logErr != nil {
+		return sql, logErr
+	}
+	return newSQL, nil
+}
+
+func getDupEntry(err error) (string, string, error) {
+	var keyName string
+	if len(fmt.Sprint(err)) < 40 {
```
WangXiangUSTC: could you please  add a comment, why need less than 40

WangXiangUSTC:
this pr is used to resolve this situations:
load more than one mysql's data into TiDB, and mysqls may have same key, so will execute insert sql failed.
is it right?

rogerszy:
Yes, it is

rogerszy:
All Suggestions have been adopted and modified

loader/db.go
```diff
@@ -273,3 +291,215 @@ func isMySQLError(err error, code uint16) bool {
 	e, ok := err.(*mysql.MySQLError)
 	return ok && e.Number == code
 }
+
+func modifySQL(db *sql.DB, err error, sql string) (string, error) {
```
WangXiangUSTC: can you provide an example, the origin sql and the modified sql?

loader/db.go
```diff
@@ -273,3 +291,215 @@ func isMySQLError(err error, code uint16) bool {
...
+					return delRes, dupDataPair, nil
+				}
+			}
+		}
+	}
+	return delRes, "", errors.New("Can't find dup Error")
+}
+
+func errorLog(dupDataSQL, errorTable, errorData, errorKey, errorPos string) error {
+	path := "./dupDataFile.txt"
```
WangXiangUSTC: how to use this file?

rogerszy: This file will appear in the catalog of each dm-worker, and we can use the catalog to confirm the file comes from which mysql instance, and file records the conflict value, conflict key, the column location of the conflict value, and an insert SQL
*********************
Key: b
Data: repqwd
Pos: 2
INSERT INTO `t3` VALUES('123456','repqwd','66668888',2);

loader/db.go
rogerszy: for example
source sql is
INSERT INTO `t3` VALUES
('aaa',',trx\',\'\'/13,.x','rrr',7), 
('bbb','qqq','ttt',6),
('123456','repqwd','66668888',2),
('ddd','eee','`~!@#?%&*()+',11);

if we find the 'repqwd' is a confilct value, we can get the pos is 2
after we delete it
modified sql is
INSERT INTO `t3` VALUES
('aaa',',trx\',\'\'/13,.x','rrr',7), 
('bbb','qqq','ttt',6),
('ddd','eee','`~!@#?%&*()+',11);

and duplicate data pair will be written into file like this
Key: b
Data: repqwd
Pos: 2
INSERT INTO t3 VALUES('123456','repqwd','66668888',2);

tips:if all data pair are duplication, we will get final SQL is 
INSERT INTO `t3` VALUES;
this is a wrong SQL, so we added a code to process it
else if isMySQLError(dupErr, tmysql.ErrParse) && dupErrFlag {
	ctx.L().Warn("execute statement", zap.String("Modified", "All data is dupError"))
	return nil // All data is dupError
}

WangXiangUSTC:
can you add some test, unit test or integration?

rogerszy:
I tried to change the code and then changed it back. This time, the same code test passed, so I think the problem of the previous test should be the problem of the test script, my code should be ok

rogerszy:
What problem does this PR solve?
	this pr is used to resolve this situation:
		load more than one mysql's data into TiDB, and mysqls may have same key, so will execute insert sql failed.

rogerszy:
What is changed and how it works?
	This pr determines whether to turn on ignore by adding a parameter in the configuration file, i.e., change the "INSERT INTO..." SQL to "INSERT IGNORE INTO..." for dupEntry error, and get the error data after the insert through the "SHOW WARNINGS" SQL, for example
	+---------+------+--------------------------------------------+
	| Level   | Code | Message                                    |
	+---------+------+--------------------------------------------+
	| Warning | 1062 | Duplicate entry 'aaa' for key 'PRIMARY'    |
	| Warning | 1062 | Duplicate entry '`~!@#?%&*()+' for key 'c' |
	+---------+------+--------------------------------------------+
	Then write a complete line of data from the SQL to the file based on the error message, the file content is
	*********************
	Time: 2019-8-8 21:52:41
	Key: PRIMARY
	Data: aaa
	Pos: 1
	INSERT INTO `t3` VALUES('aaa','bbb','ccc',1);
	*********************
	Time: 2019-8-8 21:52:41
	Key: c
	Data: `~!@#?%&*()+
	Pos: 3
	INSERT INTO `t3` VALUES('1111','2222','`~!@#?%&*()+',2);
	*********************
	The main change is to insert "IGNORE" after the "INSERT", and insert an analysis function in the transaction execution

rogerszy:
Manual test (add detailed scripts or steps below)
	For the functional testing of the code, you can prepare the same table in multiple MySQL instance, contain the same primary key and the same value, then start the task, you can view dupDataFile.txt at the dm-worker directory, observe whether all of the same primary key values are written into the file, each worker is likely to generate the file directory, depends on whether the worker met dupEntry error, at the same time to check the worker's log, if there is an error message

rogerszy:
Related contents
	According to "SHOW WARNINGS" for the complete line error data logic, There may be unconsidered corner case, in this case, the error data will not be written to the file, and returns two errors to the log at the same time, the first error will show which error data it is, such as [error = "Duplicate entry 'repqwd for key' PRIMARY '"], the second error will show where the parsing process problems, For example, [error="Can't find dup error"]

rogerszy:
Need to update the documentation
	Because this mode may not be suitable for all users, a parameter should be added to the configuration file to decide whether to turn the feature on or not. If the parameter "is-ignore:true" is added to the configuration file, it is turned on

loader/db.go
```diff
@@ -273,3 +294,258 @@ func isMySQLError(err error, code uint16) bool {
 	e, ok := err.(*mysql.MySQLError)
 	return ok && e.Number == code
 }
+
+func processErrSQL(txn *sql.Tx, err error, sql string) []error {
```
WangXiangUSTC: why return []error? seems these error just used to print warn log.

loader/db.go
WangXiangUSTC: can we just save these information into log?

dm/config/subtask.go
```diff
@@ -94,6 +94,7 @@ type SubTaskConfig struct {
 	// so we should not export flagSet
 	flagSet *flag.FlagSet
 
+	IsIgnore bool `toml:"is-ignore" json:"is-ignore"`
```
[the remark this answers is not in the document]
csuzhangxc: It can only take effect for _loader_, do we need to include text like `load` in the name or move it into `LoaderConfig`?

dm/dm-ansible/conf/task.yaml.example
```diff
@@ -1,6 +1,7 @@
 ---
 name: test # global unique
 task-mode: all  # full/incremental/all
+is-ignore: true # Whether skip dupEntry error and write the complete row of data to a file
```
csuzhangxc: Its default value is `false`, should we change it to `false` here?

dm/master/task.yaml
```diff
@@ -1,6 +1,7 @@
 ---
 name: test # global unique
 task-mode: all  # full/incremental/all
+is-ignore: true # Whether skip dupEntry error and write the complete row of data to a file
```
csuzhangxc: also, change to `false`?

csuzhangxc:
@rogerszy We have introduced a new error system in https://github.com/pingcap/dm/pull/216. Will you update some error handing to use it?

pkg/utils/db.go
```diff
@@ -292,3 +298,27 @@ func IsErrDupEntry(err error) bool {
...
+// "SHOW WARNINGS;" must follow the previous SQL statement when use "SHOW WARNINGS;" to view the warnings produced by the previous SQL statement
+// so the two SQL statements must be in the same transaction
+func ShowWarnings(txn *sql.Tx) ([]Warns, error) {
+	var tmpWarns []Warns
+	rows, err := txn.Query("SHOW WARNINGS;")
+	if err != nil {
+		return nil, terror.DBErrorAdapt(err, terror.ErrDBQueryFailed, "SHOW WARNINGS;")
+	}
+	defer rows.Close()
+	for rows.Next() {
```
WangXiangUSTC: need handle `rows.Error()`

loader/db_test.go
```diff
@@ -0,0 +1,154 @@
...
+	c.Assert(err, IsNil)
+
+	pos = 3
+	data, err = findDataPair(target, dupData, pos)
+	c.Assert(data, Equals, "")
+	c.Assert(err, NotNil)
+}
+
+func (t *testdDbSuite) TestDupEntryLog(c *C) {
+	path := "./dupDataFile.txt"
```
WangXiangUSTC: we can use c.MkDir() to create a test dir, and the file is under this dir. this dir will removed automatic

loader/db.go
```diff
@@ -278,3 +296,258 @@ func isMySQLError(err error, code uint16) bool {
...
+	}
+
+	warns, err := tutils.ShowWarnings(txn)
+	if err != nil {
+		ctx.L().Warn("get duplicate entry", zap.Error(err))
+		return
+	}
+
+	for i := 0; i < len(warns); i++ {
+		if warns[i].Level == "Warning" && warns[i].Code == "1062" {
```
WangXiangUSTC: it is better to use mysql error code https://github.com/pingcap/parser/blob/2bb2c0e8f34025aa5c5ceb713ec5233936e3d069/mysql/errcode.go#L82

loader/db_test.go
```diff
@@ -0,0 +1,154 @@
...
+
+	_, err = txn.Exec(sql)
+	c.Assert(err, IsNil)
+
+	err = txn.Commit()
+	c.Assert(err, IsNil)
+}
+
+func (t *testdDbSuite) TestGetDupEntry(c *C) {
+	mes := "Duplicate entry 'aaa' for key 'PRIMARY'"
```
WangXiangUSTC: can we add some case with special character, like `a``a`, `a\"a` and so on

csuzhangxc:
we will support to resolve skip duplicate entry through one of the two below later:

1. replace load unit with TiDB Lightning _tidb_ backend
2. support skip some errors according to error codes like [slave-skip-errors](https://dev.mysql.com/doc/refman/5.7/en/replication-options-replica.html#option_mysqld_slave-skip-errors).

so, close this PR now.

PR #222, opened 2019-07-31
Syncer ignore columns

3pointer:
Sometimes user want to add extra columns in downstream table, this is not allowed because syncer handle row events need to align with downstream column. This PR solve the problem that user may add ignore columns config so syncer can ignore specify column and align with other columns.

Add syncer config `ignore-columns`
Ignore column when handle row events
Add integration test

### Check List 

Tests 

 - Unit test
 - Integration test


csuzhangxc:
We can support to ignore columns after having _schema tracker_ later, so I'll close this PR now.

PR #225, opened 2019-08-05
worker: fix panic in handle task

amyangfei:

Fix a panic bug in dm task hander 
We can reproduce the panic with steps as follows:
1. setup a DM cluster, start a task
2. stop one of DM-worker, delete the dm_worker_meta  directory of this node
3. start this DM-worker, send `stop-task`, `pause-task` or `resume-task` from DM-master

If task meta is not stored in DM-worker's memory when we want to use it, log a warning message instead of using it. 

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes

### DIFF
--- a/dm/worker/worker.go
+++ b/dm/worker/worker.go
@@ -712,7 +712,7 @@ func (w *Worker) handleTask() {
 	failpoint.Inject("handleTaskInternal", func(val failpoint.Value) {
 		if milliseconds, ok := val.(int); ok {
 			handleTaskInterval = time.Duration(milliseconds) * time.Millisecond
-			w.l.Info("set handleTaskInterval", zap.Int("value", milliseconds))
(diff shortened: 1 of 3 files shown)


GregoryIan:
Good Job! LGTM

dm/worker/worker.go
```diff
@@ -707,7 +708,14 @@ func (w *Worker) restoreSubTask() error {
 var maxRetryCount = 10
 
 func (w *Worker) handleTask() {
-	ticker := time.NewTicker(time.Second)
+	var handleTaskInterval = time.Second
+	failpoint.Inject("handleTaskInternal", func(val failpoint.Value) {
+		if milliseconds, ok := val.(int); ok {
+			handleTaskInterval = time.Duration(milliseconds) * time.Millisecond
+			w.l.Info("set handleTaskInterval", zap.Int("value", milliseconds))
```
csuzhangxc: log a field with `failpoint` key like other failpoint inject?

PR #227, opened 2019-08-09
dm-operator: add types definition for CRD used by dm-operator

DanielZhangQD:
Support dm on k8s - Add types definition for CRD used by dm-operator

Add types definition for CRD used by dm-operator

### Check List 

Tests 


operator/pkg/apis/pingcap.com/v1alpha1/types.go
```diff
@@ -0,0 +1,376 @@
...
+	Timezone         string                               `json:"timezone,omitempty"`
+	Replicas         int32                                `json:"replicas,omitempty"`
+	Affinity         *corev1.Affinity                     `json:"affinity,omitempty"`
+	NodeSelector     map[string]string                    `json:"nodeSelector,omitempty"`
+	StorageClassName string                               `json:"storageClassName,omitempty"`
+	Tolerations      []corev1.Toleration                  `json:"tolerations,omitempty"`
+	Annotations      map[string]string                    `json:"annotations,omitempty"`
+	PVReclaimPolicy  corev1.PersistentVolumeReclaimPolicy `json:"pvReclaimPolicy,omitempty"`
+	Source           DMSource                             `json:"source,omitempty"`
+	NodeConfig       DMWorkerConfig                       `json:"nodeConfig,omitempty"`
```
DanielZhangQD: The DMWorkerConfig, DMMasterConfig, DMTask, etc. types definition should reuse the ones in dm code, however, the deepcopy functions cannot be generated if just import them in this file, may need to check further if deepcopy issue can be fixed so that no need to re-define these types here again to reduce maintain effort.

GregoryIan:
Thanks @DanielZhangQD , I will close thie pr now, and we can open it while we need it later

PR #228, opened 2019-08-09
dm-operator: add auto generated clients for CRD

DanielZhangQD:
Support dm on k8s

Add auto-generated clients for CRD operation

### Check List 

Tests 


GregoryIan:
thanks @DanielZhangQD ,  I will close thie pr now, and we can open it while we need it later

PR #231, opened 2019-08-09
dm-operator: add DMNode sync logics

DanielZhangQD:
Support dm on k8s

Sync service, headless service, configmap, statefulset for DM Master CRD and DM Worker CRD

### Check List 

Tests 


GregoryIan:
thanks @DanielZhangQD , I will close thie pr now, and we can open it while we need it later

PR #236, opened 2019-08-12
dm-worker: refine db operation

3pointer:
1. Unify db operation in Syncer and Loader into one pkg 
2. Remove syncer.maxRetry config

1. Use baseConn struct to connect db, both Syncer and Loader hold the baseConn, which can take same retry policy in the future.
2. Remove syncer.maxRetry config, In my option, if we can define error type in the future, when we met error we need to retry, we should retry. user may not realize the error we met, and don't know how many retry times need to be set, So make the user controlling the maxRetry times is a bit improperly.

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code


pkg/utils/db.go
```diff
@@ -277,35 +277,140 @@ func IsErrBinlogPurged(err error) bool {
...
+	}
+	conn.DB = db
+	return nil
+}
+
+// NormalRetryOperation will retry 100 times
+func (conn *BaseConn) NormalRetryOperation(ctx *tcontext.Context, operateFn func(*tcontext.Context, int) (interface{}, error), retryFn func(int, error) bool) (interface{}, error) {
+	var err error
+	var ret interface{}
+	for i := 0; i < 100; i++ {
```
GregoryIan: please don't use hard code

GregoryIan: just let `retryFn` to control when to exit?

GregoryIan: maybe `retryFn` may need a `tcontext` argument

pkg/utils/db.go
```diff
@@ -277,35 +277,140 @@ func IsErrBinlogPurged(err error) bool {
...
+				continue
+			}
 		}
+		break
 	}
+	return ret, err
+}
+
+// InfinityRetryOperation retry operation until succeed
+func (conn *BaseConn) InfinityRetryOperation(operateFn func() (interface{}, error), retryFn func(error) bool) (interface{}, error) {
```
GregoryIan: do `operateFn` and `retryFn` need `tcontext` argument?

pkg/utils/db.go
```diff
@@ -277,35 +277,140 @@ func IsErrBinlogPurged(err error) bool {
...
+			}
+		}
+		break
+	}
+	return ret, err
+}
+
+// QuerySQL query sql
+func (conn *BaseConn) QuerySQL(tctx *tcontext.Context, query string) (*sql.Rows, error) {
+	if conn == nil || conn.DB == nil {
```
GregoryIan: should we also check it in `NormalRetryOperation` and `InfinityRetryOperation`？

pkg/utils/db.go
```diff
@@ -277,35 +277,140 @@ func IsErrBinlogPurged(err error) bool {
...
+
+		_, err = txn.ExecContext(tctx.Context(), sqls[i].Query, sqls[i].Args...)
+		if err != nil {
+			tctx.L().Error("execute statement failed", zap.String("Query", sqls[i].Query), zap.Reflect("argument", sqls[i].Args), log.ShortError(err))
+			rerr := txn.Rollback()
+			if rerr != nil {
+				tctx.L().Error("rollback failed", zap.String("Query", sqls[i].Query), zap.Reflect("argument", sqls[i].Args), log.ShortError(rerr))
+			}
+			// we should return the exec err, instead of the rollback rerr.
+			return i, errors.Trace(err)
```
GregoryIan: should we use the new error system in #216

syncer/db.go
```diff
@@ -16,16 +16,14 @@ package syncer
...
-	"time"
-
-	"github.com/pingcap/errors"
-	"github.com/siddontang/go-mysql/mysql"
-	"go.uber.org/zap"
-
 	"github.com/pingcap/dm/dm/config"
 	tcontext "github.com/pingcap/dm/pkg/context"
 	"github.com/pingcap/dm/pkg/log"
+	"github.com/pingcap/dm/pkg/utils"
```
GregoryIan: We divide the import package path into three sections
* built-in packages
* dm packages
* other packags

pkg/utils/db.go
GregoryIan: should we sleep some time?

syncer/relay.go
```diff
@@ -60,7 +60,7 @@ func (s *Syncer) setInitActiveRelayLog() error {
 		}
 	} else {
 		// start from dumper or loader, get current pos from master
-		pos, _, err = utils.GetMasterStatus(s.fromDB.db, s.cfg.Flavor)
+		pos, _, err = utils.GetMasterStatus(s.fromDB.baseConn.DB, s.cfg.Flavor)
```
GregoryIan: wrap a function to make it like `s.fromDB.GetMasterStatus`?

loader/db.go
```diff
@@ -26,216 +28,140 @@ import (
...
+		return nil, errors.Annotatef(err, "query statement, sql [%s] failed", query)
 	}
-	return nil, errors.Annotatef(err, "query statement, sql [%s] args [%v] failed", query, args)
+	return ret.(*sql.Rows), nil
 }
 
-func (conn *Conn) executeSQL2(ctx *tcontext.Context, stmt string, maxRetry int, args ...interface{}) error {
-	if conn == nil || conn.db == nil {
-		return errors.NotValidf("database connection")
+func (conn *Conn) executeSQL(ctx *tcontext.Context, ddl bool, queries []string, args ...[]interface{}) error {
```
GregoryIan: can we replace `ddl bool` argument by other argument like `retryFn`

GregoryIan:
how to reset all connections in one unit to handle error `invalid connection`/`bad connection`?

syncer/db.go
```diff
@@ -66,208 +68,87 @@ type binlogSize struct {
...
-
-	return nil, errors.Errorf("query sql[%s] failed", query)
+	return ret.(*sql.Rows), nil
 }
 
-// Note: keep it for later use?
-func (conn *Conn) executeSQL(tctx *tcontext.Context, sqls []string, args [][]interface{}, maxRetry int) error {
-	if len(sqls) == 0 {
-		return nil
+func (conn *Conn) executeSQL(tctx *tcontext.Context, queries []string, args [][]interface{}) (int, error) {
```
WangXiangUSTC: some function use sqls, some use queries, I think it is better to uniform them

GregoryIan:
one important question: under what circumstances we will receive an invalid connection error？you can learn from code of `go-mysql`.
And I think we need to have a strategy to avoid the situation that cannot do replication because of frequent `invalid connection`. We prefer slow replication than going into the error recovery process frequently.

pkg/utils/db.go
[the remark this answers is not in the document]
3pointer: yes, its useless, removed already

pkg/utils/db.go
```diff
@@ -277,35 +277,140 @@ func IsErrBinlogPurged(err error) bool {
...
+	}
+
+	var (
+		err  error
+		rows *sql.Rows
+	)
+
+	tctx.L().Debug("Query statement", zap.String("Query", query))
+
+	rows, err = conn.DB.QueryContext(tctx.Context(), query)
```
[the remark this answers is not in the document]
3pointer: Done

pkg/utils/db.go
```diff
@@ -277,35 +277,140 @@ func IsErrBinlogPurged(err error) bool {
...
+	}
+
+	if affected, err := conn.executeSQLImp(tctx, sqls); err != nil {
+		return affected, err
+	}
+	return 0, nil
+}
+
+func (conn *BaseConn) executeSQLImp(tctx *tcontext.Context, sqls []SQL) (int, error) {
+	if conn == nil || conn.DB == nil {
```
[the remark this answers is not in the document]
3pointer: Done

pkg/utils/db.go
```diff
@@ -277,35 +277,140 @@ func IsErrBinlogPurged(err error) bool {
...
+		return 0, errors.NotValidf("database connection")
+	}
+
+	if affected, err := conn.executeSQLImp(tctx, sqls); err != nil {
+		return affected, err
+	}
+	return 0, nil
+}
+
+func (conn *BaseConn) executeSQLImp(tctx *tcontext.Context, sqls []SQL) (int, error) {
```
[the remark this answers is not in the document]
3pointer: already added comments

pkg/utils/db.go
3pointer: this PR need rebase master after #216 merged

syncer/db.go
3pointer: Done

syncer/db.go
3pointer: `sqls` and `queries` are different
`sqls` is an array of `query with argument`
`queries` is an array of `query without argument`

pkg/utils/db.go
```diff
@@ -278,17 +279,109 @@ func IsErrBinlogPurged(err error) bool {
...
-	return IsMySQLError(err, tmysql.ErrNoSuchTable)
+// IsNoSuchThreadError checks whether err is NoSuchThreadError
+func IsNoSuchThreadError(err error) bool {
+	return IsMySQLError(err, tmysql.ErrNoSuchThread)
 }
 
-// IsErrDupEntry checks whether err is DupEntry error
-func IsErrDupEntry(err error) bool {
-	return IsMySQLError(err, tmysql.ErrDupEntry)
+// BaseConn Connect to Database
```
csuzhangxc: one unimportant question, which one is better?

1. add a new pkg to support base DB connection operations
2. keep the base DB connection operations in this `utils` pkg

pkg/utils/db.go
```diff
@@ -278,17 +279,109 @@ func IsErrBinlogPurged(err error) bool {
...
+	if len(sqls) == 0 {
+		return 0, nil
+	}
+	if conn == nil || conn.DB == nil {
+		return 0, terror.ErrDBUnExpect.Generate("database connection not valid")
+	}
+
+	txn, err := conn.DB.Begin()
+	if err != nil {
+		return 0, terror.DBErrorAdapt(err, terror.ErrDBUnExpect)
```
csuzhangxc: Is it needed to include `sqls` in the error here?

BTW, one argument is needed for `ErrDBUnExpect`.

pkg/retry/strategy.go
```diff
@@ -0,0 +1,39 @@
...
+
+package retry
+
+import (
+	tcontext "github.com/pingcap/dm/pkg/context"
+	"time"
+)
+
+// Speed represents enum of retry speed
+type Speed uint8
```
csuzhangxc: any meaningful name?

pkg/retry/strategy.go
```diff
@@ -0,0 +1,39 @@
...
+
+const (
+	// SpeedSlow represents slow retry speed, every retry should wait more time depends on increasing retry times
+	SpeedSlow Speed = iota + 1
```
csuzhangxc: any words like `linear` in the variable name to make it clearer?

or different `Strategy` can have different implementation for the same `Speed`?

pkg/retry/strategy.go
```diff
@@ -0,0 +1,39 @@
...
+)
+
+// Strategy define different kind of retry strategy
+type Strategy interface {
+
+	// FiniteRetryStrategy will retry `retryCount` times when failed to operate DB.
+	// it will wait `firstRetryDuration` before it starts first retry, and then rest of retries wait time depends on retrySpeed.
+	FiniteRetryStrategy(ctx *tcontext.Context, retryCount int, firstRetryDuration time.Duration, retrySpeed Speed,
+		operateFn func(*tcontext.Context, int) (interface{}, error),
+		retryFn func(int, error) bool) (interface{}, error)
```
csuzhangxc: any comments or variable names to explain the arguments and return values of `operateFn` and `retryFn`?

pkg/utils/base_conn.go
```diff
@@ -0,0 +1,170 @@
...
+
+// BaseConn wraps a connection to DB
+type BaseConn struct {
+	DB *sql.DB
+
+	// for reset
+	DSN string
+}
+
+var _ retry.Strategy = new(BaseConn)
```
csuzhangxc: What is the meaning of this line?
let the compiler to check `BaseConn` implement `Strategy` _interface_?

We never use `Strategy` _interface_ directly, is it necessary to define such an interface?

pkg/baseconn/conn.go
```diff
@@ -0,0 +1,148 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package baseconn
+
+import (
+	"database/sql"
+	tcontext "github.com/pingcap/dm/pkg/context"
```
csuzhangxc: add a blank line between L17 and L18.

pkg/baseconn/conn.go
```diff
@@ -0,0 +1,148 @@
...
+	if err != nil {
+		return nil, terror.ErrDBDriverError.Delegate(err)
+	}
+	return &BaseConn{db, dbDSN, nil}, nil
+}
+
+// SetRetryStrategy set retry strategy for baseConn
+func (conn *BaseConn) SetRetryStrategy(strategy retry.Strategy) error {
+	if conn == nil {
+		return terror.ErrDBDriverError.Generate("no valid connection")
```
csuzhangxc: how about using `terror.ErrDBUnExpect.Generate("database connection not valid")` as we did in other places (also in L67)?

pkg/baseconn/conn.go
```diff
@@ -0,0 +1,148 @@
...
+func NewBaseConn(dbDSN string) (*BaseConn, error) {
+	db, err := sql.Open("mysql", dbDSN)
+	if err != nil {
+		return nil, terror.ErrDBDriverError.Delegate(err)
+	}
+	err = db.Ping()
+	if err != nil {
+		return nil, terror.ErrDBDriverError.Delegate(err)
+	}
+	return &BaseConn{db, dbDSN, nil}, nil
```
csuzhangxc: can we set a strategy when `NewBaseConn`?

pkg/baseconn/conn.go
```diff
@@ -0,0 +1,148 @@
...
+		return 0, nil
+	}
+	if conn == nil || conn.DB == nil {
+		return 0, terror.ErrDBUnExpect.Generate("database connection not valid")
+	}
+
+	txn, err := conn.DB.Begin()
+
+	if err != nil {
+		return 0, terror.ErrDBExecuteFailed.Delegate(err)
```
csuzhangxc: one formatted argument (`%s`) needed for `ErrDBExecuteFailed` (also in L136).

pkg/baseconn/conn.go
```diff
@@ -0,0 +1,148 @@
...
+	return l, nil
+}
+
+// Close release DB resource
+func (conn *BaseConn) Close() error {
+	if conn == nil || conn.DB == nil {
+		return nil
+	}
+
+	return terror.DBErrorAdapt(conn.DB.Close(), terror.ErrDBUnExpect, "close")
```
csuzhangxc: using `terror.ErrDBUnExpect.Generate` directly?

pkg/baseconn/conn_test.go
```diff
@@ -0,0 +1,92 @@
...
+	"github.com/pingcap/dm/pkg/terror"
+
+	"github.com/DATA-DOG/go-sqlmock"
+	. "github.com/pingcap/check"
+)
+
+var _ = Suite(&testBaseConnSuite{})
+
+type testBaseConnSuite struct {
+}
```
csuzhangxc: not enabled for this test file now, something like
```go
func TestT(t *testing.T) {
	check.TestingT(t)
}
```
is needed.

pkg/baseconn/conn.go
```diff
@@ -0,0 +1,148 @@
...
+
+// NewBaseConn builds BaseConn to connect real DB
+func NewBaseConn(dbDSN string, strategy retry.Strategy) (*BaseConn, error) {
+	db, err := sql.Open("mysql", dbDSN)
+	if err != nil {
+		return nil, terror.ErrDBDriverError.Delegate(err)
+	}
+	err = db.Ping()
+	if err != nil {
+		return nil, terror.ErrDBDriverError.Delegate(err)
```
csuzhangxc: Should we need to call `db.Close()` here?
If so, same in L76.

pkg/baseconn/conn.go
```diff
@@ -0,0 +1,148 @@
...
+	db, err := sql.Open("mysql", conn.DSN)
+	if err != nil {
+		return terror.ErrDBDriverError.Delegate(err)
+	}
+	err = db.Ping()
+	if err != nil {
+		return terror.ErrDBDriverError.Delegate(err)
+	}
+	if conn.DB != nil {
+		conn.DB.Close()
```
csuzhangxc: if returning an error, should we need to log it?

pkg/baseconn/conn.go
```diff
@@ -0,0 +1,148 @@
...
+func NewBaseConn(dbDSN string, strategy retry.Strategy) (*BaseConn, error) {
+	db, err := sql.Open("mysql", dbDSN)
+	if err != nil {
+		return nil, terror.ErrDBDriverError.Delegate(err)
+	}
+	err = db.Ping()
+	if err != nil {
+		return nil, terror.ErrDBDriverError.Delegate(err)
+	}
+	return &BaseConn{db, dbDSN, strategy}, nil
```
csuzhangxc: if passing a `nil` strategy, should we need to give a default strategy to it?

pkg/baseconn/conn.go
3pointer: Delegate may be better

csuzhangxc: `Delegate` the returned error before log it?

pkg/baseconn/conn.go
3pointer: Done

pkg/baseconn/conn.go
3pointer: you are right log it and continue is better

pkg/baseconn/conn.go
3pointer: Done

pkg/retry/strategy.go
```diff
@@ -0,0 +1,103 @@
...
+type FiniteRetryStrategy struct {
+}
+
+// DefaultRetryStrategy wait `FirstRetryDuration` before it starts first retry, and then rest of retries wait time depends on RetryInterval.
+// ErrInvalidConn is a special error, need a public retry strategy, so return it to up layer continue retry.
+func (*FiniteRetryStrategy) DefaultRetryStrategy(ctx *tcontext.Context, params DefaultRetryParams,
+	operateFn func(*tcontext.Context, int) (interface{}, error)) (interface{}, int, error) {
+	var err error
+	var ret interface{}
+	i := 0
```
csuzhangxc: we can use a single `var` block to hole these all three variables, like
```go
var (
	err error
	ret interface{}
	i   = 0
)
```
or even you can change the function signature to
```go
func (*FiniteRetryStrategy) DefaultRetryStrategy(ctx *tcontext.Context, params DefaultRetryParams,
	operateFn func(*tcontext.Context, int) (interface{}, error)) (ret interface{}, i int, err error) {
```

pkg/retry/strategy.go
```diff
@@ -0,0 +1,103 @@
...
+	for ; i < params.RetryCount; i++ {
+		ret, err = operateFn(ctx, i)
+		if err != nil {
+			if IsInvalidConnError(err) {
```
csuzhangxc: can we only let `params.IsRetryableFn` to do the check?

@amyangfei what do you think?

pkg/retry/strategy_test.go
```diff
@@ -0,0 +1,86 @@
...
+			return false
+		},
+	}
+	ctx := tcontext.Background()
+
+	operateFn := func(*tcontext.Context, int) (interface{}, error) {
+		return nil, terror.ErrDBDriverError.Generate("test database error")
+	}
+
+	_, opCount, err := strategy.DefaultRetryStrategy(ctx, params, operateFn)
```
csuzhangxc: should we need to _assert_ the first returned value (`_`)?

pkg/baseconn/conn_test.go
```diff
@@ -0,0 +1,104 @@
...
+
+var _ = Suite(&testBaseConnSuite{})
+
+type testBaseConnSuite struct {
+}
+
+func (t *testBaseConnSuite) TestBaseConn(c *C) {
+	baseConn, err := NewBaseConn("error dsn", nil)
+	c.Assert(err.(*terror.Error).Code(), Equals, terror.ErrCode(10001))
```
amyangfei:
```suggestion
	c.Assert(terror.ErrDBDriverError.Equal(err), IsTrue)
```
we can check with specific `*Error` directly, and the same with following error check

pkg/retry/errors.go
```diff
@@ -0,0 +1,90 @@
...
+		default:
+			return false
+		}
+	}
+
+	return false
+}
+
+// IsInvalidConnError tells whether it's a mysql connection error
+func IsInvalidConnError(err error) bool {
```
amyangfei: this will be more simple
```go
func IsInvalidConnError(err error) bool
    return errors.Cause(err) == mysql.ErrInvalidConn
}
```

pkg/retry/strategy.go
```diff
@@ -0,0 +1,103 @@
...
+
+// Strategy define different kind of retry strategy
+type Strategy interface {
+
+	// DefaultRetryStrategy define retry strategy
+	// params: (retry parameters for this strategy, a normal operation)
+	// return: (result of operation, number of retry, error of operation)
+	DefaultRetryStrategy(ctx *tcontext.Context,
+		params DefaultRetryParams,
+		// TODO: remove `number of retry` in operateFn, we need it now because loader.ExecuteSQL operation need it
```
amyangfei: maybe we can use the `n-time injection` feature of failpoint to remove `number of retry

dm/config/task.go
```diff
@@ -184,7 +183,6 @@ type SyncerConfig struct {
 	MetaFile    string `yaml:"meta-file" toml:"meta-file" json:"meta-file"` // meta filename, used only when load SubConfig directly
 	WorkerCount int    `yaml:"worker-count" toml:"worker-count" json:"worker-count"`
 	Batch       int    `yaml:"batch" toml:"batch" json:"batch"`
-	MaxRetry    int    `yaml:"max-retry" toml:"max-retry" json:"max-retry"`
```
GregoryIan: Does it affect the process start and stop, because some users may have already set this parameter？

pkg/retry/strategy.go
amyangfei: such as injection with
```bash
export GO_FAILPOINTS='github.com/pingcap/dm/loader/LoadExecCreateTableFailed=3*return("1152")'
```
- `1152,2` the second part is not needed any more
- fail three times because there exists one `create schema` and two failed `create table`

pkg/retry/errors.go
```diff
@@ -0,0 +1,90 @@
...
+func IsSyncerRetryableError(err error) bool {
+	if err == nil {
+		return false
+	}
+	err = errors.Cause(err) // check the original error
+	mysqlErr, ok := err.(*mysql.MySQLError)
+	if ok {
+		switch mysqlErr.Number {
+		// ER_LOCK_DEADLOCK can retry to commit while meet deadlock
+		case tmysql.ErrUnknown, gmysql.ER_LOCK_DEADLOCK, tmysql.ErrPDServerTimeout, tmysql.ErrTiKVServerTimeout, tmysql.ErrTiKVServerBusy, tmysql.ErrResolveLockTimeout, tmysql.ErrRegionUnavailable:
```
GregoryIan: The same applies to the loader

pkg/retry/strategy.go
```diff
@@ -0,0 +1,103 @@
...
+)
+
+// retryInterval represents enum of retry wait interval
+type retryInterval uint8
+
+const (
+	// LinearIncrease represents increase time wait retry policy, every retry should wait more time depends on increasing retry times
+	LinearIncrease retryInterval = iota + 1
+	// Stable represents fixed time wait retry policy, every retry should wait a fixed time
+	Stable
```
GregoryIan: should `Stable` be zero? it would be simpler

pkg/retry/strategy.go
GregoryIan: why `loader.ExecuteSQL` need it?

amyangfei: The failpoint injection fails for the first two execution.

pkg/retry/strategy.go
```diff
@@ -0,0 +1,103 @@
...
+	IsRetryableFn func(int, error) bool
+}
+
+// Strategy define different kind of retry strategy
+type Strategy interface {
+
+	// DefaultRetryStrategy define retry strategy
+	// params: (retry parameters for this strategy, a normal operation)
+	// return: (result of operation, number of retry, error of operation)
+	DefaultRetryStrategy(ctx *tcontext.Context,
```
GregoryIan: is `Apply` better?

dm/config/task.go
3pointer: some users may not realize the meaning of `MaxRetry` , only care about whether data lost or not , and some internal error like `invalid connection` also counted as once retry, it's not proper for user to control retry count, and  i think it's a bit conflict with optimizing retry part, force remove this config is worth

pkg/retry/strategy.go
3pointer: good idea

pkg/baseconn/conn.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+	err = db.Ping()
+	if err != nil {
+		db.Close()
+		return terror.ErrDBDriverError.Delegate(err)
+	}
+	if conn.DB != nil {
+		err := conn.DB.Close()
+		if err != nil {
+			tctx.L().Warn("reset connection", log.ShortError(err))
```
GregoryIan: does it have connection/session ID information？

pkg/retry/errors.go
3pointer: good idea

pkg/retry/errors.go
3pointer: I checked the master branch code in loader/db.go, it didn't apply these errors, should we change it now? it may change the original logic.

pkg/retry/strategy_test.go
3pointer: yes, add a  check to assert first value is better

pkg/baseconn/conn_test.go
3pointer: Done

pkg/retry/strategy.go
3pointer: Done

pkg/retry/strategy.go
3pointer: Done

pkg/retry/strategy.go
3pointer: Done

pkg/baseconn/conn.go
3pointer: actually `DB` doesn't have ID information, we can add this after we have own conn manager

pkg/retry/strategy.go
amyangfei: seems `params.IsRetryableFn` already filters the `ErrInvalidConn` and this check is duplicated

3pointer: some loader execute/query operations don't filter `ErrInvalidConn` for now

loader/db.go
```diff
@@ -34,213 +36,151 @@ import (
...
+			sqls = append(sqls, baseconn.SQL{Query: query, Args: []interface{}{}})
+		} else {
+			sqls = append(sqls, baseconn.SQL{Query: query, Args: args[i]})
 		}
-
-		return nil
 	}
 
-	return terror.DBErrorAdapt(err, terror.ErrDBExecuteFailed, strings.Join(sqls, ";"))
+	return conn.executeSQLCustomRetry(ctx, sqls, retry.IsLoaderRetryableError)
```
csuzhangxc: only this line is different from `executeDDL`? can we refactor it to decrease duplicate code?

loader/db.go
```diff
@@ -34,213 +36,151 @@ import (
...
+		BackoffStrategy:    retry.LinearIncrease,
+		IsRetryableFn: func(retryTime int, err error) bool {
+			ctx.L().Debug("execute statement", zap.Int("retry", retryTime), zap.String("sqls", fmt.Sprintf("%-.200v", sqls)))
+			tidbExecutionErrorCounter.WithLabelValues(conn.cfg.Name).Inc()
+			if retryFn(err) {
+				return true
 			}
-			if row != 1 {
-				ctx.L().Warn("update checkpoint", zap.Int64("affected rows", row))
+			return false
```
csuzhangxc: Is `return retryFn(err)` simpler?

loader/db.go
```diff
@@ -34,213 +36,151 @@ import (
...
+
+				if len(sqls) == 1 && strings.Contains(sqls[0].Query, "CREATE TABLE") {
+					err = tmysql.NewErr(uint16(errCode))
+					ctx.L().Warn("executeSQLCustomRetry failed", zap.String("failpoint", "LoadExecCreateTableFailed"), zap.Error(err))
+				}
+			})
+			cost := time.Since(startTime)
+			txnHistogram.WithLabelValues(conn.cfg.Name).Observe(cost.Seconds())
+			if cost > 1 {
+				ctx.L().Warn("transaction execute successfully", zap.Duration("cost time", cost))
```
csuzhangxc: always `successfully`? whether needing to check `err`?

pkg/retry/errors.go
```diff
@@ -20,6 +20,21 @@ import (
 	gmysql "github.com/siddontang/go-mysql/mysql"
 )
 
+func isRetryWhiteListError(err error) bool {
```
csuzhangxc: `isCommonRetryableError`?

loader/db.go
```diff
@@ -34,213 +36,151 @@ import (
...
-	var (
-		err  error
-		rows *sql.Rows
-	)
-
-	ctx.L().Debug("query statement", zap.String("sql", query), zap.Reflect("arguments", args))
 	startTime := time.Now()
+
+	params := retry.Params{
+		RetryCount:         10,
```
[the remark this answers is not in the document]
3pointer: I think set by the user is not properly, because users don't know real meaning of retryCount, it's usually face to internal errors such as `TiKVServerTooBusy`, if users set this retryCount too small or too big, it will bring some problems, let internal retry logic to deal with internal error is better.

loader/db.go
```diff
@@ -20,12 +20,14 @@ import (
...
 
-	"github.com/pingcap/dm/dm/config"
+	"github.com/pingcap/dm/pkg/baseconn"
 	tcontext "github.com/pingcap/dm/pkg/context"
 	"github.com/pingcap/dm/pkg/log"
+	"github.com/pingcap/dm/pkg/retry"
 	"github.com/pingcap/dm/pkg/terror"
 
 	"github.com/go-sql-driver/mysql"
+	"github.com/pingcap/dm/dm/config"
```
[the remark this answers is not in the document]
3pointer: Done in https://github.com/pingcap/dm/pull/236/commits/59dacaa6bceb5fd19f6360808d5875fdb48a01c5#diff-5f5bb90e4d87dad0cb25c377bb10679bR23

loader/db.go
```diff
@@ -34,213 +36,151 @@ import (
...
+		IsRetryableFn: func(retryTime int, err error) bool {
+			if retry.IsLoaderRetryableError(err) {
+				ctx.L().Warn("query statement", zap.Int("retry", retryTime), zap.String("sql", query), log.ShortError(err))
+				return true
+			}
+			return false
+		},
+	}
+
+	ret, _, err := conn.baseConn.RetryStrategy.Apply(
```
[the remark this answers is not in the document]
3pointer: ok, done in https://github.com/pingcap/dm/pull/236/commits/59dacaa6bceb5fd19f6360808d5875fdb48a01c5#diff-5f5bb90e4d87dad0cb25c377bb10679bR63

loader/db.go
```diff
@@ -34,213 +36,151 @@ import (
...
+		ctx,
+		params,
+		func(ctx *tcontext.Context) (interface{}, error) {
+			rows, err := conn.baseConn.QuerySQL(ctx, query)
+			return rows, err
```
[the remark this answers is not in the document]
3pointer: done with other same issue

loader/db.go
3pointer: done in https://github.com/pingcap/dm/pull/236/commits/59dacaa6bceb5fd19f6360808d5875fdb48a01c5#diff-5f5bb90e4d87dad0cb25c377bb10679bR63

loader/db.go
3pointer: done in 
 https://github.com/pingcap/dm/pull/236/commits/59dacaa6bceb5fd19f6360808d5875fdb48a01c5#diff-5f5bb90e4d87dad0cb25c377bb10679bR121

pkg/retry/errors.go
3pointer: https://github.com/pingcap/dm/pull/236/commits/59dacaa6bceb5fd19f6360808d5875fdb48a01c5#diff-17b57264f15f5001c686d1232127256fR23

loader/db.go
3pointer: actually I didn't get what this mean, I just copy from old code

pkg/retry/errors.go
GregoryIan: if right, we can change it and unify the logic

pkg/retry/errors.go
```diff
@@ -0,0 +1,42 @@
...
+		default:
+			return false
+		}
+	}
+	return false
+}
+
+// IsInvalidConnError tells whether it's a mysql connection error
+func IsInvalidConnError(err error) bool {
+	return errors.Cause(err) == mysql.ErrInvalidConn
```
GregoryIan: should we check whether it's bad connection?  invalid ~ bad connection

pkg/retry/strategy.go
GregoryIan: what about now? @3pointer

pkg/baseconn/conn.go
GregoryIan: ok, I just want output some more clear information, like which connection, to where

GregoryIan:
really cool, good job @3pointer ,  LGTM now

pkg/baseconn/conn.go
```diff
@@ -0,0 +1,162 @@
...
+	conn.DB = db
+	return nil
+}
+
+// QuerySQL defines query statement, and connect to real DB
+func (conn *BaseConn) QuerySQL(tctx *tcontext.Context, query string, args ...interface{}) (*sql.Rows, error) {
+	if conn == nil || conn.DB == nil {
+		return nil, terror.ErrDBUnExpect.Generate("database connection not valid")
+	}
+	tctx.L().Debug("query statement", zap.String("query", query))
```
amyangfei: better to log args here

pkg/retry/strategy.go
amyangfei: ErrInvalidConn is used as blacklist here, our whitelist retryable error doesn’t contain this error, so it is automatically filtered?

amyangfei: but blacklist does have higher priority here and bumps the specific error here

pkg/baseconn/conn_test.go
```diff
@@ -0,0 +1,104 @@
...
+	err = baseConn.SetRetryStrategy(nil)
+	c.Assert(terror.ErrDBUnExpect.Equal(err), IsTrue)
+
+	_, err = baseConn.QuerySQL(tctx, "select 1")
+	c.Assert(terror.ErrDBUnExpect.Equal(err), IsTrue)
+
+	_, err = baseConn.ExecuteSQL(tctx, []SQL{{"", nil}})
+	c.Assert(terror.ErrDBUnExpect.Equal(err), IsTrue)
+
+	db, mock, err := sqlmock.New()
```
amyangfei: check err is nil also

pkg/retry/strategy.go
3pointer: it seems we can remove this logic, or rename `IsInvalidConnError`  to `NotRetryableError`,so we can add blacklist error in this function, may be more clearly?

amyangfei: I prefer to remove it for the following reasons
1. The blacklist logic is covered by whitelist logic, it’s duplicated from the logic perspective
2. the invalid conn error will be recorded in the upper level code, the addition warning log here doesn’t carry more information

3pointer: ok, I'll remove it, if we need blacklist in the future, we can add it back

pkg/baseconn/conn_test.go
3pointer: fixed

loader/db.go
csuzhangxc: @3pointer but old code check `err` first, see https://github.com/pingcap/dm/blob/cb56783af3dfbee6682b6e45147078fe4bd5d5c5/loader/db.go#L161

syncer/syncer_test.go
```diff
@@ -841,9 +843,9 @@ func (s *testSyncerSuite) TestTimezone(c *C) {
 		s.resetBinlogSyncer()
 
 		// we should not use `sql.DB.Exec` to do query which depends on session variables
-		// because `sql.DB.Exec` will choose a underlying Conn for every query from the connection pool
-		// and different Conn using different session
-		// ref: `sql.DB.Conn`
+		// because `sql.DB.Exec` will choose a underlying BaseConn for every query from the connection pool
+		// and different BaseConn using different session
+		// ref: `sql.DB.BaseConn`
```
csuzhangxc: these three lines should still be `Conn` not `BaseConn`.

syncer/syncer_test.go
```diff
@@ -1307,14 +1307,14 @@ func (s *testSyncerSuite) TestSharding(c *C) {
...
 
-		// make syncer write to mock db
-		syncer.toDBs = []*Conn{{cfg: s.cfg, db: db}}
-		syncer.ddlDB = &Conn{cfg: s.cfg, db: db}
+		// make syncer write to mock baseConn
+		syncer.toDBs = []*Conn{{cfg: s.cfg, baseConn: &baseconn.BaseConn{db, s.dbAddr, &retry.FiniteRetryStrategy{}}}}
+		syncer.ddlDB = &Conn{cfg: s.cfg, baseConn: &baseconn.BaseConn{db, s.dbAddr, &retry.FiniteRetryStrategy{}}}
 
-		// run sql on upstream db to generate binlog event
+		// run sql on upstream baseConn to generate binlog event
```
csuzhangxc: `db` is more accurate here.

tests/load_interrupt/run.sh
```diff
@@ -60,7 +60,7 @@ function run() {
     check_row_count 2
 
     # only failed at the first two time, will retry later and success
-    export GO_FAILPOINTS='github.com/pingcap/dm/loader/LoadExecCreateTableFailed=return("1152,2")'
+    export GO_FAILPOINTS='github.com/pingcap/dm/loader/LoadExecCreateTableFailed=3*return("1105")'
```
csuzhangxc: why do not inject two variables now?

pkg/baseconn/conn.go
```diff
@@ -0,0 +1,162 @@
...
+		if err != nil {
+			tctx.L().Warn("reset connection", log.ShortError(err))
+		}
+	}
+	conn.DB = db
+	return nil
+}
+
+// QuerySQL defines query statement, and connect to real DB
+func (conn *BaseConn) QuerySQL(tctx *tcontext.Context, query string, args ...interface{}) (*sql.Rows, error) {
```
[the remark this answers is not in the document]
3pointer: ExecuteSQL has multiple queries and args, but QuerySQL has only one query and variable arguments, if we change `query` and `args` to `sql`, it may lost variable arguments(...args)

loader/db.go
```diff
@@ -34,213 +36,126 @@ import (
...
+	if conn == nil || conn.baseConn == nil {
 		return nil, terror.ErrDBUnExpect.Generate("database connection not valid")
 	}
 
-	var (
-		err  error
-		rows *sql.Rows
-	)
-
-	ctx.L().Debug("query statement", zap.String("sql", query), zap.Reflect("arguments", args))
```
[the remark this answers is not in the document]
3pointer: this Debug log  moved to baseConn.QuerySQL

loader/db.go
```diff
@@ -34,213 +36,126 @@ import (
...
+			return conn.baseConn.QuerySQL(ctx, query, args...)
+		})
+
 	defer func() {
 		if err == nil {
 			cost := time.Since(startTime)
 			queryHistogram.WithLabelValues(conn.cfg.Name).Observe(cost.Seconds())
 			if cost > 1 {
-				ctx.L().Warn("query statement", zap.String("sql", query), zap.Reflect("arguments", args), zap.Duration("cost time", cost))
+				ctx.L().Warn("query statement", zap.String("sql", query), zap.Duration("cost time", cost))
```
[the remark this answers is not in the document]
3pointer: 1. yes
2. Is Observe add cost.Seconds at same time? and if retry happened, can we get actual executing time?

loader/db.go
3pointer: oh I see, I fixed it in https://github.com/pingcap/dm/pull/236/files#diff-5f5bb90e4d87dad0cb25c377bb10679bR132

pkg/baseconn/conn.go
```diff
@@ -0,0 +1,162 @@
...
+	if conn == nil || conn.DB == nil {
+		return nil, terror.ErrDBUnExpect.Generate("database connection not valid")
+	}
+	tctx.L().Debug("query statement", zap.String("query", query), zap.Reflect("argument", args))
+
+	rows, err := conn.DB.QueryContext(tctx.Context(), query, args...)
+
+	if err != nil {
+		tctx.L().Error("query statement failed", zap.String("query", query), log.ShortError(err))
```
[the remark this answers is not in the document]
3pointer: all log in `loader/db.go`  `syncer/db.go` `baseconn/db.go` has been formatted. you can check it now.
1. Debug has moved from loader/syncer to baseconn
2. Warn log in retryFn
3. Error log after all retries failed

pkg/retry/strategy.go
```diff
@@ -0,0 +1,94 @@
...
+		operateFn func(*tcontext.Context) (interface{}, error),
+	) (interface{}, int, error)
+}
+
+// FiniteRetryStrategy will retry `RetryCount` times when failed to operate DB.
+type FiniteRetryStrategy struct {
+}
+
+// Apply for FiniteRetryStrategy, it wait `FirstRetryDuration` before it starts first retry, and then rest of retries wait time depends on BackoffStrategy.
+// ErrInvalidConn is a special error, need a public retry strategy, so return it to up layer continue retry.
```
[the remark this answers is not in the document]
3pointer: Done in https://github.com/pingcap/dm/pull/236/files#diff-738c9f8c5ac61da20a2cbec3d15d5d4eR67

pkg/retry/strategy_test.go
```diff
@@ -0,0 +1,94 @@
...
+
+	params.IsRetryableFn = func(int, error) bool {
+		return true
+	}
+	_, opCount, err = strategy.Apply(ctx, params, operateFn)
+	c.Assert(opCount, Equals, params.RetryCount)
+	c.Assert(terror.ErrDBDriverError.Equal(err), IsTrue)
+
+	params.RetryCount = 3
+	operateFn = func(ctx *tcontext.Context) (interface{}, error) {
```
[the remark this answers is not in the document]
3pointer: I fixed it

loader/db.go
```diff
@@ -34,213 +36,126 @@ import (
...
+		IsRetryableFn: func(retryTime int, err error) bool {
+			if retry.IsRetryableError(err) {
+				ctx.L().Warn("query statement", zap.Int("retry", retryTime), zap.String("sql", query), log.ShortError(err))
+				return true
+			}
+			return false
+		},
+	}
+
+	ret, _, err := conn.baseConn.ApplyRetryStrategy(
```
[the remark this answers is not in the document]
3pointer: Done in https://github.com/pingcap/dm/pull/236/files#diff-5f5bb90e4d87dad0cb25c377bb10679bR63

pkg/baseconn/conn.go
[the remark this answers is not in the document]
csuzhangxc: if so, how about let all of them use `query` and `args` directly? Is `SQL` struct necessary?

syncer/db.go
```diff
@@ -62,229 +69,163 @@ type binlogSize struct {
...
-
 	return dbs, nil
 }
 
-func closeDBs(tctx *tcontext.Context, dbs ...*Conn) {
-	for _, db := range dbs {
-		err := db.close()
+func closeConns(tctx *tcontext.Context, conns ...*Conn) {
+	for _, conn := range conns {
+		err := conn.baseConn.Close()
```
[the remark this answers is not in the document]
3pointer: Yes, it's intential, after we optimize  connection, it will be better

loader/db.go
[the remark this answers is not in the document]
csuzhangxc: IMO, let _cost_ as the time of executing a SQL statement only once (and only if successfully) **without** any retry time is better.

tests/load_interrupt/run.sh
3pointer: @amyangfei told me about this, I think may be second variable is useless

syncer/syncer_test.go
3pointer: I change some comments in syncer_test, you can check it now

tests/load_interrupt/run.sh
amyangfei: the second is used as failure time, which includes one `create schema` and two `create table` failure

pkg/baseconn/conn.go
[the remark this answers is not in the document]
3pointer: you are right, I removed `SQL`, it's seems better now

loader/db.go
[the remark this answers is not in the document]
3pointer: Ok, I fixed it, you can check it now

loader/db.go
```diff
@@ -34,213 +36,126 @@ import (
...
-	ctx.L().Debug("query statement", zap.String("sql", query), zap.Reflect("arguments", args))
 	startTime := time.Now()
+
+	params := retry.Params{
+		RetryCount:         10,
+		FirstRetryDuration: time.Second,
+		BackoffStrategy:    retry.Stable,
+		IsRetryableFn: func(retryTime int, err error) bool {
+			if retry.IsRetryableError(err) {
+				ctx.L().Warn("query statement", zap.Int("retry", retryTime), zap.String("sql", query), log.ShortError(err))
```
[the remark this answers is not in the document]
3pointer: ditto https://github.com/pingcap/dm/pull/236#discussion_r317358331

loader/db.go
```diff
@@ -34,213 +36,126 @@ import (
...
-				continue
-			}
-			if row != 1 {
-				ctx.L().Warn("update checkpoint", zap.Int64("affected rows", row))
+	params := retry.Params{
+		RetryCount:         10,
+		FirstRetryDuration: 2 * time.Second,
+		BackoffStrategy:    retry.LinearIncrease,
+		IsRetryableFn: func(retryTime int, err error) bool {
+			ctx.L().Debug("execute statement", zap.Int("retry", retryTime), zap.String("sqls", fmt.Sprintf("%-.200v", sqls)))
```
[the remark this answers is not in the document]
3pointer: ditto https://github.com/pingcap/dm/pull/236#discussion_r317358331

syncer/db.go
```diff
@@ -62,229 +69,163 @@ type binlogSize struct {
...
-func (conn *Conn) querySQL(tctx *tcontext.Context, query string, maxRetry int) (*sql.Rows, error) {
-	if conn == nil || conn.db == nil {
-		return nil, terror.ErrDBUnExpect.Generate("database connection not valid")
-	}
-
-	var (
-		err  error
-		rows *sql.Rows
-	)
-	tctx.L().Debug("query statement", zap.String("sql", query))
```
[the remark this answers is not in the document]
3pointer: ditto https://github.com/pingcap/dm/pull/236#discussion_r317358331

syncer/db.go
```diff
@@ -62,229 +69,163 @@ type binlogSize struct {
...
-	for i := 0; i < maxRetry; i++ {
-		if i > 0 {
-			sqlRetriesTotal.WithLabelValues("query", conn.cfg.Name).Add(1)
-			tctx.L().Warn("query statement", zap.Int("retry", i), zap.String("sql", query))
```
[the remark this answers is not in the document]
3pointer: ditto https://github.com/pingcap/dm/pull/236#discussion_r317358331

syncer/db.go
```diff
@@ -62,229 +69,163 @@ type binlogSize struct {
...
-
-	if conn == nil || conn.db == nil {
-		return terror.ErrDBUnExpect.Generate("database connection not valid")
-	}
-
-	var err error
-	for i := 0; i < maxRetry; i++ {
-		if i > 0 {
-			sqlRetriesTotal.WithLabelValues("stmt_exec", conn.cfg.Name).Add(1)
-			tctx.L().Warn("execute statements", zap.Int("retry", i), zap.Strings("sqls", sqls), zap.Reflect("arguments", args))
```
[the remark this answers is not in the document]
3pointer: ditto https://github.com/pingcap/dm/pull/236#discussion_r317358331

syncer/db.go
```diff
@@ -62,229 +69,163 @@ type binlogSize struct {
...
-		}
-
-		return rows, nil
+// ResetConn reset baseConn.*DB's connection pool
+func (conn *Conn) ResetConn(tctx *tcontext.Context) error {
+	if conn.baseConn == nil {
+		return terror.ErrDBUnExpect.Generate("database base connection not valid")
 	}
-
-	tctx.L().Error("query statement failed", zap.String("sql", query), log.ShortError(err))
```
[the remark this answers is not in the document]
3pointer: ditto https://github.com/pingcap/dm/pull/236#discussion_r317358331

syncer/db.go
```diff
@@ -62,229 +69,163 @@ type binlogSize struct {
...
+	return utils.GetServerUUID(conn.baseConn.DB, flavor)
+}
 
-		return nil
-	}
+func (conn *Conn) getParser(ansiQuotesMode bool) (*parser.Parser, error) {
+	return utils.GetParser(conn.baseConn.DB, ansiQuotesMode)
+}
 
-	tctx.L().Error("exec statements failed", zap.Strings("sqls", sqls), zap.Reflect("arguments", args), log.ShortError(err))
```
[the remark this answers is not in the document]
3pointer: ditto https://github.com/pingcap/dm/pull/236#discussion_r317358331

loader/db.go
```diff
@@ -34,213 +36,126 @@ import (
...
-			if isRetryableError(err) {
-				ctx.L().Warn("query statement", zap.Int("retry", i), zap.String("sql", query), zap.Reflect("arguments", args), log.ShortError(err))
-				continue
-			}
-			ctx.L().Error("query statement", zap.String("sql", query), zap.Reflect("arguments", args), log.ShortError(err))
-			return nil, terror.DBErrorAdapt(err, terror.ErrDBQueryFailed, query)
-		}
-		return rows, nil
-	}
-	ctx.L().Error("query statement", zap.String("sql", query), zap.Reflect("arguments", args), log.ShortError(err))
```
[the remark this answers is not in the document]
3pointer: ditto https://github.com/pingcap/dm/pull/236#discussion_r317358331

loader/db.go
```diff
@@ -34,213 +36,126 @@ import (
...
-				continue
-			}
-			ctx.L().Error("execute statement", zap.String("sql", stmt), zap.Reflect("arguments", args), log.ShortError(err))
-			return terror.DBErrorAdapt(err, terror.ErrDBExecuteFailed, stmt)
-		}
-		return nil
+	if err != nil {
+		return nil, err
 	}
-	ctx.L().Error("execute statement", zap.String("sql", stmt), zap.Reflect("arguments", args), log.ShortError(err))
```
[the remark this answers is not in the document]
3pointer: ditto https://github.com/pingcap/dm/pull/236#discussion_r317358331

syncer/db.go
```diff
@@ -62,229 +69,163 @@ type binlogSize struct {
...
-	if conn == nil || conn.db == nil {
-		return terror.ErrDBUnExpect.Generate("database connection not valid")
-	}
+func (conn *Conn) fetchAllDoTables(bw *filter.Filter) (map[string][]string, error) {
+	return utils.FetchAllDoTables(conn.baseConn.DB, bw)
+}
 
-	startTime := time.Now()
-	defer func() {
-		txnHistogram.WithLabelValues(conn.cfg.Name).Observe(time.Since(startTime).Seconds())
```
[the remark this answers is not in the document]
3pointer: fix it in https://github.com/pingcap/dm/pull/236/files#diff-c3cc6fccd12f8e1679a575d1aaf8fd14R169

syncer/syncer_test.go
3pointer: ok, I updated some same comments

syncer/db.go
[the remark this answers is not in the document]
csuzhangxc: seems still not added back

csuzhangxc:
how about adding slow log (`if cost > 1`) for _syncer/db.go_?

3pointer:
> how about adding slow log (`if cost > 1`) for _syncer/db.go_?

you mean log as same as loader/d.go such like `ctx.L().Warn("transaction execute successfully", zap.Duration("cost time", cost))`?

tests/load_interrupt/run.sh
csuzhangxc: so, is it still needed?

csuzhangxc:
> > how about adding slow log (`if cost > 1`) for _syncer/db.go_?
> 
> you mean log as same as loader/d.go such like `ctx.L().Warn("transaction execute successfully", zap.Duration("cost time", cost))`?

YES, it is.

3pointer:
> > > how about adding slow log (`if cost > 1`) for _syncer/db.go_?
> > 
> > 
> > you mean log as same as loader/d.go such like `ctx.L().Warn("transaction execute successfully", zap.Duration("cost time", cost))`?
> 
> YES, it is.

ok, fixed

tests/load_interrupt/run.sh
amyangfei: not needed

syncer/db.go
```diff
@@ -62,229 +71,171 @@ type binlogSize struct {
...
+			return conn.baseConn.QuerySQL(ctx, query, args...)
+		},
+	)
 
-	txn, err := conn.db.Begin()
 	if err != nil {
-		tctx.L().Error("begin transaction", zap.Strings("sqls", sqls), zap.Reflect("arguments", args), zap.Error(err))
-		return terror.DBErrorAdapt(err, terror.ErrDBExecuteFailed, strings.Join(sqls, ";"))
+		tctx.L().Error("query statement failed after retry", zap.String("query", query), zap.Reflect("argument", args), log.ShortError(err))
+		return nil, terror.ErrDBQueryFailed.Delegate(err, query)
```
amyangfei: `err` is already a `*terror.Error` object, we should return `err` instead of using `Delegate`.

syncer/db.go
```diff
@@ -62,229 +71,171 @@ type binlogSize struct {
...
+				}
+			}
+			return ret, err
+		})
+
 	if err != nil {
-		tctx.L().Error("commit in executing job", zap.Error(err))
-		return &ExecErrorContext{err: terror.DBErrorAdapt(err, terror.ErrDBDriverError), pos: jobs[0].currentPos, jobs: fmt.Sprintf("%v", jobs)}
+		tctx.L().Error("execute statements failed after retry", zap.Strings("queries", queries), zap.Reflect("arguments", args), log.ShortError(err))
+		return ret.(int), terror.ErrDBExecuteFailed.Delegate(err, queries)
```
amyangfei: ditto

syncer/db.go
3pointer: ok, Done

syncer/db.go
3pointer: Done

PR #240, opened 2019-08-19
worker: fix subtask initialized state

amyangfei:

A subtask can transform from `paused` to `running` state by `resume-task`, but we should not allow a subtask to `resume` in running state.

WX20190819-145932@2x

When we create a new subtask, its state is `New`, and then we call `Init` of this task to initialize all the units of it, then the `initialized` flag will be set to true.
In the old logic, assuming there is a subtask with mode `all` and is running under sync unit, we stop it and restart from task config. Then it will not correctly set the `initialized` value.

Use a variable to record where subtask initialization is successful.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

Related changes

### DIFF
--- a/dm/worker/subtask.go
+++ b/dm/worker/subtask.go
@@ -109,6 +109,7 @@ func (st *SubTask) Init() error {
 
 	st.DDLInfo = make(chan *pb.DDLInfo, 1)
 
+	initializeUnitSuccess := true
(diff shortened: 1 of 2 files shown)


GregoryIan:
LGTM. Good Job!

PR #251, opened 2019-08-26
*: fix some error scope in task units

amyangfei:
some db operation misses error scope field.

add error scope when the scope is determined

### Check List 

Tests 

 - Integration test

Code changes

Side effects

Related changes


WangXiangUSTC:
LGTM

PR #252, opened 2019-08-27
syncer: ignore error may skip execute some ddls

WangXiangUSTC:
for example:
execute ddl in upstream `alter table t1 add column a int, add column b int`, syncer will split it to two ddls:
ddl1 `alter table t1 add column a int`
ddl2 `alter table t1 add column b int`

syncer will execute sqls ddl1 and ddl2 together, and if execute ddl1 meet error and this error can be ignored, will skip execute ddl2. 

fix it, judge whether error can be ignored when execute sql, if can ignore will continue execute next sql.

### Check List 

Tests 

 - Unit test

Code changes


csuzhangxc:
@3pointer PTAL

PR #254, opened 2019-08-28
syncer: reset db connections during resume

amyangfei:

As DM already has the task auto recovery mechanism and a new database management system, we can reset the database connection to downstream after a subtask is resumed

Reset `toDBs`, `ddlDB` and RemoteCheckpoint db connections in syncer unit when a subtask is resumed 

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

Related changes


dm/config/subtask.go
```diff
@@ -116,6 +116,7 @@ type SubTaskConfig struct {
 	HeartbeatUpdateInterval int    `toml:"heartbeat-update-interval" json:"heartbeat-update-interval"`
 	HeartbeatReportInterval int    `toml:"heartbeat-report-interval" json:"heartbeat-report-interval"`
 	EnableHeartbeat         bool   `toml:"enable-heartbeat" json:"enable-heartbeat"`
+	ResumeResetConn         bool   `toml:"resume-reset-conn" json:"resume-reset-conn"`
```
WangXiangUSTC: I think we can remove this config, and reset connection as default.

csuzhangxc: agree

PR #256, opened 2019-08-29
pkg/retry: add some retryable error

WangXiangUSTC:
some tidb error can be resolved  by retry, add some retryable error type

add some error

ErrQueryInterrupted: query execution was interrupted by user
ErrWriteConflictInTiDB: commit meets an write conflict error
ErrWriteConflict:  similar with ErrWriteConflictInTiDB
ErrTableLocked:  the table was locked by other session, can retry success when other session release the lock.

### DIFF
--- a/pkg/retry/errors.go
+++ b/pkg/retry/errors.go
@@ -27,7 +27,7 @@ func IsRetryableError(err error) bool {
 	if ok {
 		switch mysqlErr.Number {
 		// ER_LOCK_DEADLOCK can retry to commit while meet deadlock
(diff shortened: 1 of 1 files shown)


csuzhangxc:
@WangXiangUSTC Add any description about these newly added error codes?

WangXiangUSTC:
@csuzhangxc  description added

WangXiangUSTC:
@amyangfei this error occurred when some one kill the query

amyangfei:
> @amyangfei this error occurred when some one kill the query

maybe a query/execute longer than timeout will cause this error in our use scenario, otherwise our db op may not cancel a query (except for context canceled, then retry is not needed yet)

PR #257, opened 2019-08-29
*: limit the error message and log length; fix query cost check

csuzhangxc:

1. limit the error message and log length for DB operations
2. fix query cost check

1. truncate queries and arguments for log and error message
2. fix `if cost > 1` check

### Check List 

Tests 

 - Unit test

### DIFF
--- a/loader/db.go
+++ b/loader/db.go
@@ -20,18 +20,23 @@ import (
 	"strings"
 	"time"
 
+	"github.com/go-sql-driver/mysql"
+	"github.com/pingcap/errors"
+	"github.com/pingcap/failpoint"
+	tmysql "github.com/pingcap/parser/mysql"
+	"go.uber.org/zap"
+
 	"github.com/pingcap/dm/dm/config"
 	"github.com/pingcap/dm/pkg/baseconn"
 	tcontext "github.com/pingcap/dm/pkg/context"
 	"github.com/pingcap/dm/pkg/log"
 	"github.com/pingcap/dm/pkg/retry"
 	"github.com/pingcap/dm/pkg/terror"
+	"github.com/pingcap/dm/pkg/utils"
+)
 
-	"github.com/go-sql-driver/mysql"
-	"github.com/pingcap/errors"
-	"github.com/pingcap/failpoint"
-	tmysql "github.com/pingcap/parser/mysql"
-	"go.uber.org/zap"
+const (
+	stringLenLimit = 1024
 )
 
 // Conn represents a live DB connection
@@ -51,7 +56,10 @@ func (conn *Conn) querySQL(ctx *tcontext.Context, query string, args ...interfac
 		BackoffStrategy:    retry.Stable,
 		IsRetryableFn: func(retryTime int, err error) bool {
 			if retry.IsRetryableError(err) {
-				ctx.L().Warn("query statement", zap.Int("retry", retryTime), zap.String("query", query), zap.Reflect("argument", args), log.ShortError(err))
+				ctx.L().Warn("query statement", zap.Int("retry", retryTime),
+					zap.String("query", utils.TruncateString(query, stringLenLimit)),
+					zap.Reflect("argument", utils.TruncateInterface(args, stringLenLimit)),
+					log.ShortError(err))
 				return true
 			}
 			return false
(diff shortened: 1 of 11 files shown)


csuzhangxc:
@amyangfei @WangXiangUSTC PTAL

pkg/baseconn/conn.go
```diff
@@ -89,13 +94,18 @@ func (conn *BaseConn) QuerySQL(tctx *tcontext.Context, query string, args ...int
...
 	rows, err := conn.DB.QueryContext(tctx.Context(), query, args...)
 
 	if err != nil {
-		tctx.L().Error("query statement failed", zap.String("query", query), zap.Reflect("argument", args), log.ShortError(err))
-		return nil, terror.ErrDBQueryFailed.Delegate(err, query)
+		tctx.L().Error("query statement failed",
+			zap.String("query", utils.TruncateString(query, stringLenLimit)),
+			zap.Reflect("argument", utils.TruncateInterface(args, stringLenLimit)),
+			log.ShortError(err))
+		return nil, terror.ErrDBQueryFailed.Delegate(err, utils.TruncateString(query, stringLenLimit))
```
WangXiangUSTC: how about remain the whole error message, only do TruncateString when need print it

loader/db.go
```diff
@@ -67,14 +75,20 @@ func (conn *Conn) querySQL(ctx *tcontext.Context, query string, args ...interfac
 			if err == nil {
 				cost := time.Since(startTime)
 				queryHistogram.WithLabelValues(conn.cfg.Name).Observe(cost.Seconds())
-				if cost > 1 {
-					ctx.L().Warn("query statement", zap.String("query", query), zap.Reflect("argument", args), zap.Duration("cost time", cost))
+				if cost.Seconds() > 1 {
+					ctx.L().Warn("query statement",
+						zap.String("query", utils.TruncateString(query, stringLenLimit)),
+						zap.Reflect("argument", utils.TruncateInterface(args, stringLenLimit)),
```
WangXiangUSTC: seems we can use zap.String

pkg/utils/common.go
```diff
@@ -33,8 +33,9 @@ import (
 // move to tidb-tools later
 
 const (
-	maxRetryCount = 3
-	retryTimeout  = 3 * time.Second
+	maxRetryCount  = 3
+	retryTimeout   = 3 * time.Second
+	stringLenLimit = 1024
```
WangXiangUSTC: can we only define it in untils/string.go, and use `-1` as argument in `TruncateString` if use default stringLenLimit?

pkg/baseconn/conn.go
csuzhangxc: the error message can be logged/sent by RPC in many places, it's hard to find out all of them, and ensure truncate it in later usage.

pkg/utils/string_test.go
```diff
@@ -0,0 +1,75 @@
...
+		{v: 123, n: 0, expect: "..."},
+		{v: 123, n: 3, expect: "123"},
+		{v: compose{b: true, i: 123, s: "abc"}, n: 3, expect: "{b:..."},
+		{v: compose{b: true, i: 123, s: "abc"}, n: 9, expect: "{b:true i..."},
+		{v: compose{b: true, i: 123, s: "abc"}, n: 99, expect: "{b:true i:123 s:abc}"},
+		{v: []string{"abc", "123"}, n: 3, expect: "[ab..."},
+		{v: []string{"abc", "123"}, n: 9, expect: "[abc 123]"},
+		{v: []compose{{b: true, i: 123, s: "abc"}, {b: false, i: 456, s: "def"}}, n: 3, expect: "[{b..."},
+		{v: []compose{{b: true, i: 123, s: "abc"}, {b: false, i: 456, s: "def"}}, n: 99, expect: "[{b:true i:123 s:abc} {b:false i:456 s:def}]"},
+	}
```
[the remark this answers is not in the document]
csuzhangxc: added in https://github.com/pingcap/dm/pull/257/commits/02aa7c3d02ce604be1d3f8ec03e59aefbe60d7c2.

pkg/baseconn/conn.go
csuzhangxc: any better suggestion?

loader/db.go
csuzhangxc: done in https://github.com/pingcap/dm/pull/257/commits/02aa7c3d02ce604be1d3f8ec03e59aefbe60d7c2.

pkg/utils/common.go
csuzhangxc: done in https://github.com/pingcap/dm/pull/257/commits/02aa7c3d02ce604be1d3f8ec03e59aefbe60d7c2.

csuzhangxc:
@amyangfei @WangXiangUSTC PTAL again

pkg/baseconn/conn.go
WangXiangUSTC: ok, no better suggestion

WangXiangUSTC:
LGTM

PR #259, opened 2019-08-30
DM-worker: refine lock usage

csuzhangxc:

1. get lock only if needed
2. deep copy some protobuf message

1. only get the lock when access `subTasks` in `Worker. handleTask`
2. change `Lock` to `RLock` in `Worker.Status`
3. use `proto.Clone` to deep copy a protobuf message

### Check List 

Tests 

 - Unit test
 	1. modify Load/Flush source code for checkpoint to sleep some time
 	2. use `start-task`/`stop-task` to operate the subtask (these commands will return with RPC timeout)
 	3. use `query-status` to query the status (should without RPC timeout)


csuzhangxc:
@GregoryIan @amyangfei PTAL

GregoryIan:
I have two suggestions
* we can use fine-grained locks here
* clear protection scope of the lock, and it's better to use code pattern like
```
struct {
sync.Mutex
object
}
```

GregoryIan:
Good Job！

csuzhangxc:
@GregoryIan PTAL again

dm/worker/status.go
```diff
@@ -50,25 +50,27 @@ func (w *Worker) Status(stName string) []*pb.SubTaskStatus {
 	w.RLock()
```
GregoryIan: can we remove them now?

dm/worker/status.go
```diff
@@ -129,25 +131,27 @@ func (w *Worker) StatusJSON(stName string) string {
 func (w *Worker) Error(stName string) []*pb.SubTaskError {
 	w.Lock()
```
GregoryIan: same with `w.Status`

dm/worker/status.go
csuzhangxc: exclusive with `Close`.

csuzhangxc: removed in https://github.com/pingcap/dm/pull/259/commits/01f693ba86374689365439155952183820910bb5.

dm/worker/status.go
csuzhangxc: removed in https://github.com/pingcap/dm/pull/259/commits/01f693ba86374689365439155952183820910bb5.

GregoryIan:
LGTM

csuzhangxc:
@amyangfei PTAL

dm/worker/worker.go
```diff
@@ -406,12 +407,12 @@ func (w *Worker) doFetchDDLInfo(ctx context.Context, ch chan<- *pb.DDLInfo) {
 		return // should not go here
 	}
 
-	w.RLock()
-	w.subTasks[v.Task].SaveDDLInfo(v)
```
amyangfei: does this mean if v.Task not in `w.subTasks`, it will panic in the old code, do we need to add a log if `st` is nil

dm/worker/subtask_holder.go
```diff
@@ -0,0 +1,75 @@
...
+
+import (
+	"sync"
+)
+
+// subTaskHolder holds subtask instances.
+type subTaskHolder struct {
+	// protect concurrent access of `subTasks`
+	// the internal operation of the subtask should be protected by any mechanism in the subtask itself.
+	mu       sync.RWMutex
```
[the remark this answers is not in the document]
csuzhangxc: I personally tend to use this style, because of the methods from `sync.RWMutex` will not be exported to the outer space.

In this case, `Worker` -> `subTaskHolder` -> `mu`, then `Worker` can not call `mu`'s method, like `w.subTaskHolder.Lock`.

BTW, seems the builtin pkgs prefer this style? maybe we can define our style later?

dm/worker/worker.go
csuzhangxc: YES, in the old code, if calling this method when closing the Worker, it may panic.

added a log in https://github.com/pingcap/dm/pull/259/commits/6905cdd2a1123408668f196076e33e8f2ede80e1.

PR #260, opened 2019-08-30
*: add some necessary error check

amyangfei:

Some error can't be ignored

Add necessary error check

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

Related changes


syncer/sharding_group.go
```diff
@@ -431,7 +431,10 @@ func (k *ShardingGroupKeeper) AddGroup(targetSchema, targetTable string, sourceI
 	if schemaGroup, ok := k.groups[schemaID]; !ok {
 		k.groups[schemaID] = NewShardingGroup(k.cfg.SourceID, k.shardMetaSchema, k.shardMetaTable, sourceIDs, meta, true)
 	} else {
-		schemaGroup.Merge(sourceIDs)
+		needShardingHandle, synced, remain, err = schemaGroup.Merge(sourceIDs)
```
WangXiangUSTC: is this a bug

csuzhangxc: Has any side effect with setting `needShardingHandle, synced, remain, err` here?

amyangfei: no side effect, but  `_, _, _, err = schemaGroup.Merge(sourceIDs)` seems better?

csuzhangxc: `_, _, _, err = schemaGroup.Merge(sourceIDs)` seems better

PR #264, opened 2019-09-02
dm-worker: fix lock acquire starvation

amyangfei:

DM-worker must be shutdown gracefully after receiving `kill` or `kill -hup`, but some condition exists lock acquire starvation and `worker.Close()` will not end.

one scenario is as follows:

`worker.Close()`, w acquired a lock -> `w.relayPurger.Close()` -> relayPurger.wg.Wait(), wait relayPurger.run ends
`relayPurger.run` -> ticker trigger, tryPurge -> interceptor, aka `*Worker` call `ForbidPurge()`, which will try to acquire `worker.RLock`

we can wrap Mutex with some atomic operation to achieve `TryLock` feature, but I think it is over-optimization for this scenario.

remove lock protection in `ForbidPurge`

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

Related changes


GregoryIan:
LGTM

PR #265, opened 2019-09-02
relay: retry for upstream bad connection

csuzhangxc:

- retry to read binlog events from upstream when encountering bad connection error

- check return error from binlog reader
- use _pkg/backoff_ to trigger the retry

### Check List 

Tests 

 - Unit test
 	- run DM-worker with fetching binlog events
 	- `KILL` connection for the relay unit
 	- observe whether relay retry to fetch binlog events again

Side effects

 - Increased code complexity


csuzhangxc:
@GregoryIan @amyangfei PTAL

relay/relay.go
```diff
@@ -273,11 +275,34 @@ func (r *Relay) process(parentCtx context.Context) error {
...
 	transformer2 := transformer.NewTransformer(parser2)
 
 	go r.doIntervalOps(parentCtx)
 
-	return r.handleEvents(parentCtx, reader2, transformer2, writer2)
+	// handles binlog events with retry mechanism.
+	// it only do the retry for some binlog reader error now.
+	for {
+		err := r.handleEvents(parentCtx, reader2, transformer2, writer2)
+		if !readerRetry.Check(parentCtx, err) {
```
amyangfei: check whether `err` is nil before `readerRetry.Check`, and return directly if `err` is nil?

csuzhangxc: if `err` is `nil`, I think we should return directly (by cancel)

csuzhangxc: return directly if `err` is `nil` in https://github.com/pingcap/dm/pull/265/commits/6f8de211a7a8435c473053577f898db21e495489.

pkg/parser/common_test.go
```diff
@@ -83,11 +83,7 @@ func (t *testParserSuite) TestParser(c *C) {
 	}
 
 	unsupportedSQLs := []string{
-		"alter table bar ADD FULLTEXT INDEX `fulltext` (`name`) WITH PARSER ngram",
```
GregoryIan: remove two supported cases handy?

csuzhangxc: YES, because parser also updated when executing
```
GO111MODULE=on go get -u github.com/siddontang/go-mysql@f52d30c9fcb7824f4b6f38e3e4ed1037a1a05acd
```

relay/relay.go
```diff
@@ -273,11 +275,36 @@ func (r *Relay) process(parentCtx context.Context) error {
...
+		}
+
+		r.tctx.L().Info("receive retryable error for binlog reader", log.ShortError(err))
+		err = reader2.Close() // close the previous reader
+		if err != nil {
+			r.tctx.L().Error("fail to close binlog event reader", zap.Error(err))
+		}
+		reader2, err = r.setUpReader() // setup a new one
+		if err != nil {
+			return err
```
GregoryIan: does need it `errors.Trace`?

relay/retry/reader.go
```diff
@@ -0,0 +1,89 @@
...
+	}, nil
+}
+
+// Check checks whether should retry for the error.
+func (rr *ReaderRetry) Check(ctx context.Context, err error) bool {
+	if !isReaderRetryableError(err) {
+		return false
+	}
+
+	for lrt := rr.lastRetryTime; time.Since(lrt) > rr.cfg.BackoffRollback; lrt = lrt.Add(rr.cfg.BackoffRollback) {
```
GregoryIan: can we this logic code into backoff pkg? but not now

relay/relay.go
csuzhangxc: the returned error is a `terror.Error` with `code`, `class`, `scope`, etc.
so, is it still need to use `errors.Trace`?

relay/relay.go
```diff
@@ -273,11 +275,36 @@ func (r *Relay) process(parentCtx context.Context) error {
...
+	for {
+		err := r.handleEvents(parentCtx, reader2, transformer2, writer2)
+		if err == nil {
+			return nil
+		} else if !readerRetry.Check(parentCtx, err) {
+			return err
+		}
+
+		r.tctx.L().Info("receive retryable error for binlog reader", log.ShortError(err))
+		err = reader2.Close() // close the previous reader
```
GregoryIan: can we close `reader2` twice?

relay/relay.go
amyangfei: it is not allowed to wrap `*terror.Error` with `errors.Trace`, which will lose error trace stack

relay/relay.go
csuzhangxc: YES, we can close `Reader` any times, but it will return a `can not close` error except the first one.

In this place, we only close a `reader2` instance once. this L298 close the previous one, and in the `defer` in L260 ~ L265, it may close the first one (without retry) or the latest one (with retry).

relay/retry/reader.go
csuzhangxc: and a function like `func (b *Backoff) Adjust()` to adjust/rollback `cwnd`?
If so, maybe too many things `backoff` can do?
what your opinion? @amyangfei

GregoryIan:
Rest LGTM

relay/retry/reader.go
amyangfei: I prefer not added this logic into `backoff` pkg, but maybe we can add a function like `RollbackN`?

relay/relay.go
csuzhangxc: check `reader2 != nil` in https://github.com/pingcap/dm/pull/265/commits/c6baaabfd39e63573a1d61ccbd8e1ebb1582e380.

GregoryIan:
LGTM

dm/worker/relay.go
```diff
@@ -96,6 +97,13 @@ func NewRealRelayHolder(cfg *Config) RelayHolder {
 		},
 		BinLogName: clone.RelayBinLogName,
 		BinlogGTID: clone.RelayBinlogGTID,
+		ReaderRetry: rr.ReaderRetryConfig{ // we use config for TaskChecker now
```
amyangfei:
```suggestion
		ReaderRetry: rr.ReaderRetryConfig{ // we use config from TaskChecker now
```

dm/worker/relay.go
```diff
@@ -320,6 +328,13 @@ func (h *realRelayHolder) Update(ctx context.Context, cfg *Config) error {
 			User:     cfg.From.User,
 			Password: cfg.From.Password,
 		},
+		ReaderRetry: rr.ReaderRetryConfig{ // we use config for TaskChecker now
```
amyangfei: ditto

PR #269, opened 2019-09-05
loader: fix data values lost in error message

csuzhangxc:

when column mapping returned with error in _load_ unit, the original data values will be lost in the error message like

```
[code=34004:class=load-unit:scope=internal:level=high] mapping row data [] for table &{sourceSchema:test_parse_rows_values sourceTable:tbl_1 targetSchema:test_parse_rows_values targetTable:tbl_1 columnNameList:[c1] insertHeadStmt:}: id must less than 17592186044416, greater than or equal to 0, but got 585520728116297738 not valid
```

overwrite the original data values when no error returned, then the error message will like

```
[code=34004:class=load-unit:scope=internal:level=high] mapping row data [585520728116297738] for table &{sourceSchema:test_parse_rows_values sourceTable:tbl_1 targetSchema:test_parse_rows_values targetTable:tbl_1 columnNameList:[c1] insertHeadStmt:}: id must less than 17592186044416, greater than or equal to 0, but got 585520728116297738 not valid
```

### Check List 

Tests 

 - Unit test


loader/convert_data_test.go
```diff
@@ -196,3 +196,32 @@ func (t *testConvertDataSuite) TestParseTableWithGeneratedColumn(c *C) {
...
+				Arguments:     []string{"1", "", ""},
+			},
+		}
+	)
+
+	columnMapping, err := cm.NewMapping(false, rules)
+	c.Assert(err, IsNil)
+
+	values, err := parseRowValues(data, ti, columnMapping)
+	c.Assert(err, ErrorMatches, ".*585520728116297738.*")
```
amyangfei:
```suggestion
	c.Assert(err, ErrorMatches, ".*mapping row data \\[585520728116297738\\] for table.*")
```
The error messages contains two `585520728116297738`

csuzhangxc:
@WangXiangUSTC PTAL

loader/convert_data.go
```diff
@@ -158,10 +158,11 @@ func parseRowValues(str []byte, table *tableInfo, columnMapping *cm.Mapping) ([]
 
 	if columnMapping != nil {
 		var err error
```
WangXiangUSTC: can remove this line

csuzhangxc: removed in https://github.com/pingcap/dm/pull/269/commits/a629c67e673437a169ae8304f1c8d45562f065eb.

PR #280, opened 2019-09-10
conn: add max idle conn config

amyangfei:

currently we use a single `sql.DB` connection pool for multiple concurrent executors in single DM-worker, each time one a txn is committed, the connection will be tried to put back to the idle connection pool. Too small idle connection pool size will lead to rejection of the put-back and database too many connections re-create.

set database MaxIdleConns according to worker-count

### Check List 

Tests 

 - Unit test
 - Integration test
 

Code changes

Side effects

Related changes

   - Fix a bug that may cause database connection re-establish too frequent

### DIFF
--- a/syncer/db.go
+++ b/syncer/db.go
@@ -210,14 +210,8 @@ func createBaseConn(dbCfg config.DBConfig, timeout string, rawDBCfg *baseconn.Ra
 	return baseConn, nil
 }
 
-func getRawDBConfig(cfg *config.SubTaskConfig) *baseconn.RawDBConfig {
-	return &baseconn.RawDBConfig{
-		MaxIdleConns: cfg.SyncerConfig.WorkerCount,
-	}
-}
-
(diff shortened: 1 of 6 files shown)


syncer/db.go
```diff
@@ -210,14 +210,8 @@ func createBaseConn(dbCfg config.DBConfig, timeout string, rawDBCfg *baseconn.Ra
...
 
-func getRawDBConfig(cfg *config.SubTaskConfig) *baseconn.RawDBConfig {
-	return &baseconn.RawDBConfig{
-		MaxIdleConns: cfg.SyncerConfig.WorkerCount,
-	}
-}
-
 func createConn(cfg *config.SubTaskConfig, dbCfg config.DBConfig, timeout string) (*Conn, error) {
-	baseConn, err := createBaseConn(dbCfg, timeout, getRawDBConfig(cfg))
+	baseConn, err := createBaseConn(dbCfg, timeout, &baseconn.RawDBConfig{MaxIdleConns: 2})
```
csuzhangxc: use `baseconn.DefaultRawDBConfig()`?

PR #286, opened 2019-09-16
syncer: fix in-memory tps count

amyangfei:

in-memory job counter counts on both added jobs and finished jobs.

count finished jobs only.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

Related changes


amyangfei:
PTAL @GregoryIan @csuzhangxc

PR #289, opened 2019-09-17
dm-worker: split DDL sync from sync function

3pointer:
Optimize sync function 

Split sync DDL from `sync()` because sync DDL and sync DML don't have to many common logic.
Split them to make the code clear.

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

### DIFF
--- a/syncer/syncer.go
+++ b/syncer/syncer.go
@@ -902,12 +902,12 @@ func (s *Syncer) syncDDL(ctx *tcontext.Context, queueBucket string, db *Conn, dd
 			}
 			s.ddlExecInfo.ClearBlockingDDL()
 		}
+		s.jobWg.Done()
 		if err != nil {
 			s.execErrorDetected.Set(true)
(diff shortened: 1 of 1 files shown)


csuzhangxc:
@amyangfei @csuzhangxc PTAL

syncer/syncer.go
```diff
@@ -856,6 +856,62 @@ func (s *Syncer) flushCheckPoints() error {
...
+				sqlJob.ddlExecItem.resp <- err
+			}
+			s.ddlExecInfo.ClearBlockingDDL()
+		}
+		if err != nil {
+			s.execErrorDetected.Set(true)
+			s.runFatalChan <- unit.NewProcessError(pb.ErrorType_ExecSQL, errors.ErrorStack(err))
+			continue
+		}
+		s.jobWg.Done()
```
csuzhangxc: Do we need to call `s.jobWg.Done()` even `err != nil`?

3pointer: Yes, we need to call

PR #304, opened 2019-09-30
relay: begin with the latest binlog file if relay position is not set

WangXiangUSTC:
relay begin with the latest binlog file if relay position is not set

### Check List 

Tests 

 - Manual test (don't set relay log's config, and check the binlog file's start name under relay log's dir)

Related changes


relay/relay.go
```diff
@@ -498,14 +498,21 @@ func (r *Relay) reSetupMeta() error {
 		return err
 	}
 
+	lastPos, _, err := utils.GetMasterStatus(r.db, r.cfg.Flavor)
```
csuzhangxc: do not support fetching binlog from the newest GTID sets?

Is any better name than `lastPos`? like `latestPos` or `newest`?

WangXiangUSTC: my mistake, fixed in https://github.com/pingcap/dm/pull/304/commits/36603239528f65c8a27f7cd1c9660cc920d2a65b

csuzhangxc:
can you and some test cases to cover this logic? like mock `r.db` for `SHOW MASTER STATUS`?

csuzhangxc:
> can you and some test cases to cover this logic? like mock `r.db` for `SHOW MASTER STATUS`?

Oh, maybe cases in _realy/meta_test.go_ are enough.

relay/relay.go
```diff
@@ -498,14 +498,21 @@ func (r *Relay) reSetupMeta() error {
 		return err
 	}
 
+	latestPos, latestGTID, err := utils.GetMasterStatus(r.db, r.cfg.Flavor)
```
csuzhangxc: should we always call `GetMasterStatus` or only when not pos/GTIDs are given?

WangXiangUSTC: update in https://github.com/pingcap/dm/pull/304/commits/cd77953171a8b31e7a69f0fcb14c8fec8e0be5cd

WangXiangUSTC:
> > can you and some test cases to cover this logic? like mock `r.db` for `SHOW MASTER STATUS`?
> 
> Oh, maybe cases in _realy/meta_test.go_ are enough.
> or add cases with both `binlogName` and `latestBinlogName` are not empty.

update in https://github.com/pingcap/dm/pull/304/commits/0909418297c854c33e853c7631dd3048df699c2e

csuzhangxc:
@amyangfei PTAL

relay/meta.go
```diff
@@ -165,30 +165,30 @@ func (lm *LocalMeta) AdjustWithStartPos(binlogName string, binlogGTID string, en
...
 		gset, err = gtid.ParserGTID(lm.flavor, binlogGTID)
 		if err != nil {
 			return false, terror.Annotatef(err, "relay-binlog-gtid %s", binlogGTID)
 		}
-	}
-
-	// verified, update them
-	if enableGTID {
-		lm.BinLogName = minCheckpoint.Name
+		lm.BinLogName = ""
```
amyangfei: If gtid is enabled in the relay module, the `BinlogName` is useless when pulling relay log, but the relay module still saves a `BinlogName` based on the binlog position each time when flushing relay meta. Shall we unify the usage of `BinlogName` in relay meta?

csuzhangxc: if we get GTID from the upstream master server, we can also get BinlogName at the same time. but if we get GTID from the startup config file, BinlogName is often empty.

But we may still assign BinlogName to meta.

WangXiangUSTC: so, should we set `BinlogName` here  when enable gtid? @csuzhangxc @amyangfei

amyangfei: I prefer to set `BinlogName`

WangXiangUSTC: update in https://github.com/pingcap/dm/pull/304/commits/e6d94862de9f2676332fc2f0a4f36a261b6e8b35, do you think it's ok @csuzhangxc @amyangfei

PR #305, opened 2019-10-08
worker: fix potential panic when restoring subtask from meta

csuzhangxc:

```golang
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x53b07e]

goroutine 47 [running]:
context.WithDeadline(0x0, 0x0, 0xbf5be448e73dc52d, 0x463dd73766, 0x2ad5a60, 0x0, 0x0, 0x0)
/usr/local/go/src/context/context.go:384 +0x4e
context.WithTimeout(0x0, 0x0, 0x45d964b800, 0xc001974200, 0x2, 0x2)
/usr/local/go/src/context/context.go:451 +0x6b
github.com/pingcap/dm/dm/worker.(*SubTask).unitTransWaitCondition(0xc000372e40, 0x0, 0x0)
/home/jenkins/workspace/build_dm_master/go/src/github.com/pingcap/dm/dm/worker/subtask.go:627 +0x3ca
github.com/pingcap/dm/dm/worker.(*SubTask).run(0xc000372e40)
/home/jenkins/workspace/build_dm_master/go/src/github.com/pingcap/dm/dm/worker/subtask.go:182 +0x5c
github.com/pingcap/dm/dm/worker.(*SubTask).fetchResult(0xc000372e40, 0xc0004a7560)
/home/jenkins/workspace/build_dm_master/go/src/github.com/pingcap/dm/dm/worker/subtask.go:248 +0xb28
created by github.com/pingcap/dm/dm/worker.(*SubTask).run
/home/jenkins/workspace/build_dm_master/go/src/github.com/pingcap/dm/dm/worker/subtask.go:196 +0x4a8
```

before this PR, we have the following order in _worker.go_:
1. `InitConditionHub` registers `Worker` instances
2. `restoreSubTask` restores subtasks from meta
3. `w.ctx, w.cancel = context.WithCancel(context.Background())` inits `w.ctx` and `w.cancel`

but when restoring subtasks from meta, `w.ctx` may be used, then `nil` pointer will be accessed and panicked.

in this PR, we make sure have the following order:
1. `w.ctx, w.cancel = context.WithCancel(context.Background())` inits `w.ctx` and `w.cancel`
2. `InitConditionHub` registers `Worker` instances
3. `restoreSubTask` restores subtasks from meta

### Check List 

Tests 

 - Unit test
 - Integration test

Related changes

### DIFF
--- a/dm/worker/worker.go
+++ b/dm/worker/worker.go
@@ -82,7 +82,7 @@ func NewWorker(cfg *Config) (w *Worker, err error) {
 	w.ctx, w.cancel = context.WithCancel(context.Background())
 
 	defer func(w2 *Worker) {
-		if err != nil { // when err !=nil, `w` will become nil in this func, so we pass `w` in defer.
+		if err != nil { // when err != nil, `w` will become nil in this func, so we pass `w` in defer.
 			// release resources, NOTE: we need to refactor New/Init/Start/Close for components later.
 			w2.cancel()
 			w2.subTaskHolder.closeAllSubTasks()
--- a/dm/worker/worker.go
+++ b/dm/worker/worker.go
@@ -71,14 +71,29 @@ type Worker struct {
 }
 
 // NewWorker creates a new Worker
-func NewWorker(cfg *Config) (*Worker, error) {
-	w := &Worker{
+func NewWorker(cfg *Config) (w *Worker, err error) {
+	w = &Worker{
 		cfg:           cfg,
 		relayHolder:   NewRelayHolder(cfg),
 		tracer:        tracing.InitTracerHub(cfg.Tracer),
 		subTaskHolder: newSubTaskHolder(),
 		l:             log.With(zap.String("component", "worker controller")),
 	}
+	w.ctx, w.cancel = context.WithCancel(context.Background())
+
+	defer func(w2 *Worker) {
+		if err != nil { // when err !=nil, `w` will become nil in this func, so we pass `w` in defer.
+			// release resources, NOTE: we need to refactor New/Init/Start/Close for components later.
+			w2.cancel()
+			w2.subTaskHolder.closeAllSubTasks()
+			if w2.meta != nil {
+				w2.meta.Close()
+			}
+			if w2.db != nil {
+				w2.db.Close()
+			}
+		}
+	}(w)
 
 	// initial relay holder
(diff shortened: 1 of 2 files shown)


csuzhangxc:
@amyangfei @WangXiangUSTC PTAL

dm/worker/worker.go
```diff
@@ -71,14 +71,29 @@ type Worker struct {
...
+		if err != nil { // when err !=nil, `w` will become nil in this func, so we pass `w` in defer.
+			// release resources, NOTE: we need to refactor New/Init/Start/Close for components later.
+			w2.cancel()
+			w2.subTaskHolder.closeAllSubTasks()
+			if w2.meta != nil {
+				w2.meta.Close()
+			}
+			if w2.db != nil {
+				w2.db.Close()
+			}
```
WangXiangUSTC: can we just use w2.Close(), and maybe need add some check in `Close` (close something when it is not nil)

dm/worker/worker.go
```diff
@@ -71,14 +71,29 @@ type Worker struct {
...
 		cfg:           cfg,
 		relayHolder:   NewRelayHolder(cfg),
 		tracer:        tracing.InitTracerHub(cfg.Tracer),
 		subTaskHolder: newSubTaskHolder(),
 		l:             log.With(zap.String("component", "worker controller")),
 	}
+	w.ctx, w.cancel = context.WithCancel(context.Background())
+
+	defer func(w2 *Worker) {
+		if err != nil { // when err !=nil, `w` will become nil in this func, so we pass `w` in defer.
```
WangXiangUSTC:
```suggestion
		if err != nil { // when err != nil, `w` will become nil in this func, so we pass `w` in defer.
```

dm/worker/worker.go
csuzhangxc: I think we can.
But I also think we should refactor the pattern of components' new/init/start/close later. 
like `new` only creates fields but does not acquire (IO) resources and start no goroutines, then when `new` failed, no operation needs to do and let GC to remove the objects is enough.

I think we need to do some refactoring when developing the HA version of DM.

what's your opinion?

WangXiangUSTC: OK，we can do this later

WangXiangUSTC:
LGTM

PR #299, opened 2019-09-25
.*: remove column mapping in example config file and help output

WangXiangUSTC:
### What problem does this PR solve? 
column mapping is not recommend to use, remove it in example config file and help output 

### What is changed and how it works?
remove it

Related changes

 - Need to cherry-pick to the release branch
 - Need to update the documentation

PR #308, opened 2019-10-09
.*: remove column mapping in example config file and help output (#299)

WangXiangUSTC:
column mapping is not recommend to use, remove it in example config file and help output

remove it
cherry pick from #299 

Related changes


csuzhangxc:
@amyangfei PTAL

PR #310, opened 2019-10-14
mydumper: generate tables needed to dump for mydumper automatically

lichunzhu:
[TOOL-1654](https://internal.pingcap.net/jira/browse/TOOL-1654)
Now users need to add `extra-args` for `mydumper` to help it decide tables needed to dump. When the situations are very complicated it's very hard for users to do that. This PR aims to generate the tables needed to be dumped automatically so that we can simplify the configuration process.

Generate mydumper `extra-args` through black-white list and router rules when it is empty. Before start mydumper will connect to `DB` to fetch all the tables from database, filter useful tables and generate `extra-args` in `--tables-list db1.tb1,db2.tb2..` format.
A unit-test is added to test whether it can generate right args when `extra-args` is empty. For integration_tests except "start_task" I expurge the `extra-args` to test whether it can run correctly.
### Check List 

Tests 

 - Unit test
 - Integration test

Related changes


mydumper/util.go
```diff
@@ -37,3 +55,41 @@ func trimOutQuotes(arg string) string {
 	}
 	return arg
 }
+
+func createBaseConn(dbCfg config.DBConfig, timeout string, rawDBCfg *baseconn.RawDBConfig) (*baseconn.BaseConn, error) {
```
WangXiangUSTC: already had one in syncer/db.go, how about move it to pkg and use same one?

mydumper/util.go
```diff
@@ -37,3 +55,41 @@ func trimOutQuotes(arg string) string {
...
+	if err != nil {
+		return "", err
+	}
+	var stringTables string
+	for _, tables := range sourceTables {
+		for _, table := range tables {
+			stringTables += "," + table.Schema + "." + table.Name
+		}
+	}
+	return stringTables[1:], nil // start from 1 to avoid first comma
```
WangXiangUSTC: how about use strings.Join

mydumper/mydumper.go
```diff
@@ -261,12 +262,18 @@ func (m *Mydumper) constructArgs() []string {
 	extraArgs := strings.Fields(cfg.ExtraArgs)
 	if len(extraArgs) > 0 {
 		ret = append(ret, ParseArgLikeBash(extraArgs)...)
+	} else {
```
WangXiangUSTC: if extraArgs is not empty but like "--skip-tz-utc", we should also generate `--tables-list` for it

lichunzhu:
@WangXiangUSTC Revised, PTAL.

lichunzhu:
@WangXiangUSTC Revised, PTAL.

dm/worker/worker_test.go
```diff
@@ -161,7 +162,7 @@ func (t *testServer) TestTaskAutoResume(c *C) {
 	// start task
 	cli := t.createClient(c, fmt.Sprintf("127.0.0.1:%d", port))
 	subtaskCfgBytes, err := ioutil.ReadFile("./subtask.toml")
-	_, err = cli.StartSubTask(context.Background(), &pb.StartSubTaskRequest{Task: string(subtaskCfgBytes)})
+	_, err = cli.StartSubTask(context.Background(), &pb.StartSubTaskRequest{Task: strings.Replace(string(subtaskCfgBytes), "#extra-args", "extra-args", 1)})
```
WangXiangUSTC: why update this?

mydumper/mydumper_test.go
```diff
@@ -59,6 +60,25 @@ func (m *testMydumperSuite) TestArgs(c *C) {
...
+	c.Assert(err, IsNil)
 	c.Assert(args, DeepEquals, expected)
 }
+
+func (m *testMydumperSuite) TestEmptyExtraArgs(c *C) {
+	expected := strings.Fields("--host 127.0.0.1 --port 3306 --user root " +
+		"--outputdir ./dumped_data --threads 4 --chunk-filesize 64 --skip-tz-utc " +
+		"--tables-list mockDatabase.mockTable1,mockDatabase.mockTable2 " +
+		"--password 123")
+	m.cfg.MydumperConfig.ExtraArgs = ""
```
WangXiangUSTC: also need add unit test for case:
1. ExtraArgs contain arg like "--skip-tz-utc", this case need generate table list
2. ExtraArgs contain arg like "-T" or || "--table-list" and so on, this case don't need generate table list

WangXiangUSTC:
@csuzhangxc PTAL

dm/worker/worker_test.go
lichunzhu: Because this test uses `github.com/pingcap/dm/dm/worker/subtask.toml` as input and in this file `extra-args` is commented. Without this mydumper will try to connect to the DB and generate `extra-args` which will cause failure.

WangXiangUSTC: ok, can add a comment for this

WangXiangUSTC:
LGTM

mydumper/mydumper_test.go
```diff
@@ -52,13 +53,54 @@ func (m *testMydumperSuite) SetUpSuite(c *C) {
...
+	c.Assert(failpoint.Enable("github.com/pingcap/dm/pkg/utils/mockSuccessfullyFetchTargetDoTables", "return(true)"), IsNil)
+	defer failpoint.Disable("github.com/pingcap/dm/pkg/utils/mockSuccessfullyFetchTargetDoTables")
+
+	expectedMockResult := "--tables-list mockDatabase.mockTable1,mockDatabase.mockTable2"
+	// empty extraArgs
+	generateArgsAndCompare(c, m, expectedMockResult, "")
+	// extraArgs doesn't contains -T/-B/-x args
+	m.cfg.MydumperConfig.SkipTzUTC = false
+	generateArgsAndCompare(c, m, expectedMockResult, "--skip-tz-utc")
+	m.cfg.MydumperConfig.SkipTzUTC = true
```
csuzhangxc: why do we assign `true` to `SkipTzUTC` here?

mydumper/util.go
```diff
@@ -37,3 +50,45 @@ func trimOutQuotes(arg string) string {
...
+	if err != nil {
+		return "", err
+	}
+	var filteredTables []string
+	for _, tables := range sourceTables {
+		for _, table := range tables {
+			filteredTables = append(filteredTables, table.Schema+"."+table.Name)
+		}
+	}
+	return strings.Join(filteredTables, ","), nil
```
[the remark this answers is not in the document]
WangXiangUSTC: and need add a test for this case

pkg/baseconn/conn.go
```diff
@@ -68,6 +71,22 @@ func NewBaseConn(dbDSN string, strategy retry.Strategy, rawDBCfg *RawDBConfig) (
 	return &BaseConn{db, dbDSN, strategy, rawDBCfg}, nil
 }
 
+// CreateBaseConn builds BaseConn through DBConfig to connect real DB
+func CreateBaseConn(dbCfg config.DBConfig, timeout string, rawDBCfg *RawDBConfig) (*BaseConn, error) {
+	failpoint.Inject("createEmptyBaseConn", func(_ failpoint.Value) {
```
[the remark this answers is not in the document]
WangXiangUSTC: ok, how about remove this function and still use `createBaseConn ` in syncer in this pr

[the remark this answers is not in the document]
csuzhangxc: > ok, how about remove this function and still use `createBaseConn ` in syncer in this pr

I think it's fine.

tests/sharding/conf/dm-task.yaml
```diff
@@ -71,7 +71,7 @@ mydumpers:
     threads: 4
     chunk-filesize: 64
     skip-tz-utc: true
-    extra-args: "--regex '^sharding.*'"
+    extra-args: ""
```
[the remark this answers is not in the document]
lichunzhu: Sure. Updated in https://github.com/pingcap/dm/pull/310/commits/2b941a2baba25c6771588989674584077c7566d4.

mydumper/util.go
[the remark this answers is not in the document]
lichunzhu: @csuzhangxc It seems that **` " '** and space are also valid char in MySQL's indentifer. How can we escape it?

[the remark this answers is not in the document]
csuzhangxc: you can ref <https://dev.mysql.com/doc/refman/5.7/en/identifiers.html>

pkg/utils/common.go
```diff
@@ -121,6 +121,20 @@ func FetchAllDoTables(db *sql.DB, bw *filter.Filter) (map[string][]string, error
 
 // FetchTargetDoTables returns all need to do tables after filtered and routed (fetches from upstream MySQL)
 func FetchTargetDoTables(db *sql.DB, bw *filter.Filter, router *router.Table) (map[string][]*filter.Table, error) {
+	failpoint.Inject("mockGenerateExtraArgs", func(_ failpoint.Value) {
```
csuzhangxc: can we move this `mockGenerateExtraArgs ` to the caller (`fetchMyDumperDoTables`) in _mydumper/util.go_?

mydumper/util.go
```diff
@@ -37,3 +51,47 @@ func trimOutQuotes(arg string) string {
 	}
 	return arg
 }
+
+// fetchMyDumperDoTables fetches and filters the tables that needed to be dumped through black-white list and route rules
+func fetchMyDumperDoTables(cfg *config.SubTaskConfig) (string, error) {
+	dbDSN := fmt.Sprintf("%s:%s@tcp(%s:%d)/?charset=utf8mb4&maxAllowedPacket=%d",
+		cfg.From.User, cfg.From.Password, cfg.From.Host, cfg.From.Port, cfg.From.MaxAllowedPacket)
```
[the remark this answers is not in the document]
lichunzhu: Revised in https://github.com/pingcap/dm/pull/310/commits/4ca1eec5bfc0395b69aaa6a5c214c4ecbb25595a.

mydumper/mydumper.go
```diff
@@ -262,11 +263,18 @@ func (m *Mydumper) constructArgs() []string {
 	if len(extraArgs) > 0 {
 		ret = append(ret, ParseArgLikeBash(extraArgs)...)
 	}
+	if needToGenerateDoTables(extraArgs) {
+		doTables, err := fetchMyDumperDoTables(cfg)
+		if err != nil {
+			return nil, err
+		}
+		ret = append(ret, "--tables-list", doTables)
```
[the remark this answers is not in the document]
lichunzhu: Revised in https://github.com/pingcap/dm/pull/310/commits/4ca1eec5bfc0395b69aaa6a5c214c4ecbb25595a.

mydumper/mydumper.go
```diff
@@ -263,11 +264,19 @@ func (m *Mydumper) constructArgs() []string {
 	if len(extraArgs) > 0 {
 		ret = append(ret, ParseArgLikeBash(extraArgs)...)
 	}
+	if needToGenerateDoTables(extraArgs) {
+		m.logger.Info("-T/-B/-x index is not given, now we will start to generate table list that mydumper needs to dump through black-white list and route rules from given fromDB")
```
csuzhangxc: 1. what is `index` mean in this message?
2. how about remove `and route rules` from the message?

mydumper/util.go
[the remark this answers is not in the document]
lichunzhu: I add a TODO comment since mydumper doesn't support quote `.

mydumper/mydumper.go
lichunzhu: What about:
```suggest
m.logger.Info("Tables needed to dump are not given, now we will start to generate table list that mydumper needs to dump through black-white list from given fromDB")
```

csuzhangxc: fine

PR #316, opened 2019-10-16
master: only fetch necessary DM-worker's config

csuzhangxc:

The previous version of DM-master always fetches all DM-workers' config when starting a task.

When having any un-accessible DM-worker, then we can't start any task.

Change to only fetches necessary DM-workers' config

### Check List 

Tests 

 - Unit test
 - Integration test
 	- start task with un-accessible DM-worker

Related changes


csuzhangxc:

WangXiangUSTC:
LGTM, and it is better to add a test

csuzhangxc:
> LGTM, and it is better to add a test

@WangXiangUSTC I modify _start_task_ test case in https://github.com/pingcap/dm/pull/316/commits/e13abaf612496e4b82ad0dfdeab1fe1c6d81c106, PTAL again.

WangXiangUSTC:
LGTM

PR #317, opened 2019-10-17
syncer: fast fail on some error in ddl execution

amyangfei:

syncer doesn't need to retry on the unsupported DDL execution error, such as `can't drop column with index`, `unsupported drop integer primary key`, etc. Pause task directly with error is enough when meeting these errors.

add error blacklist in db operation retryable decision

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

Related changes

    Fast fail on some TiDB unsupported DDL error in the binlog replication unit.

### DIFF
--- a/pkg/retry/errors.go
+++ b/pkg/retry/errors.go
@@ -14,10 +14,35 @@
 package retry
 
 import (
+	"strings"
+
 	"github.com/go-sql-driver/mysql"
 	"github.com/pingcap/errors"
 	tmysql "github.com/pingcap/parser/mysql"
 	gmysql "github.com/siddontang/go-mysql/mysql"
+)
+
+var (
+	// UnsupportedDDLMsgs list the error messages of some unsupported DDL in TiDB
+	UnsupportedDDLMsgs = []string{
+		"can't drop column with index",
+		"unsupported add column",
+		"unsupported modify column",
+		"unsupported modify",
+		"unsupported drop integer primary key",
+	}
+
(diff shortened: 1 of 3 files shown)


syncer/db.go
```diff
@@ -159,7 +159,7 @@ func (conn *Conn) executeSQLWithIgnore(tctx *tcontext.Context, ignoreError func(
 		FirstRetryDuration: retryTimeout,
 		BackoffStrategy:    retry.Stable,
 		IsRetryableFn: func(retryTime int, err error) bool {
-			if retry.IsRetryableError(err) {
+			if retry.IsRetryableErrorFastFailFilter(err) {
```
csuzhangxc: only needing to replace `IsRetryableError` with `IsRetryableErrorFastFailFilter` in sync's `execute` SQL, right?

amyangfei: yes. have effect on DDL execution in sync unit only

pkg/retry/errors.go
```diff
@@ -14,12 +14,37 @@
...
 	gmysql "github.com/siddontang/go-mysql/mysql"
 )
 
+var (
+	// UnsupportedDDLMsgs list the error messages of some unsupported DDL in TiDB
+	UnsupportedDDLMsgs = []string{
+		"can't drop column with index",
+		"unsupported add column",
+		"unsupported modify column",
+		"unsupported modify",
```
csuzhangxc: `unsupported modify` other things?

amyangfei: charset and collate

csuzhangxc:
@WangXiangUSTC PTAL

PR #323, opened 2019-10-17
worker: generate index flavor automatically by querying version info from DB

lichunzhu:
[TOOL-1166](https://internal.pingcap.net/jira/browse/TOOL-1166)
Now, we need to set flavor for DM-worker manually, but it may be set incorrectly.

use `dbutil.ShowVersion` and `check.IsMariaDB` from [tidb-tools](https://github.com/pingcap/tidb-tools) to detect upstream MySQL/MariaDB server type and set flavor automatically.

### Check List 

Tests 

 - Unit test
 - Integration test

Related changes


dm/worker/config.go
```diff
@@ -239,6 +247,46 @@ func (c *Config) configFromFile(path string) error {
...
+	}
+	dbDSN := fmt.Sprintf("%s:%s@tcp(%s:%d)/?charset=utf8mb4&maxAllowedPacket=%d",
+		clone.From.User, clone.From.Password, clone.From.Host, clone.From.Port, *clone.From.MaxAllowedPacket)
+	conn, err := newBaseConn(dbDSN, nil, baseconn.DefaultRawDBConfig())
+	if err != nil {
+		return err
+	}
+	defer conn.Close()
+
+	query := "SELECT @@version_comment"
```
WangXiangUSTC: how about  move these code to a function like `GetFlavor` in pkg,  I need this function to do something

lichunzhu: OK

csuzhangxc:
@lichunzhu I found `version_comment         | (Ubuntu)` in <https://stackoverflow.com/questions/8987679/how-to-retrieve-the-current-version-of-a-mysql-database/8987742#8987742>.

lichunzhu:
@csuzhangxc It works if we only support mariaDB and mysql.
When we run select version seperately in mysql and mariaDB, **mysql** might be like:
```
+------------+
| @@version  |
+------------+
| 5.7.26-log |
+------------+
```
mariaDB might be like:
```
+--------------------------------------+
| @@version                            |
+--------------------------------------+
| 10.4.8-MariaDB-1:10.4.8+maria~bionic |
+--------------------------------------+
```

csuzhangxc:
> If there is a kind of database which supports `select @@version` command but is not neither mysql nor mariaDB, it would be judged as **mysql**.

I think it's fine now, if neither `mysql` nor `mariadb`, it will fail later.

lichunzhu:
So shall we change `select @@version_comment` to `select @@version` and use `IsMariaDB`?

csuzhangxc:
> So shall we change `select @@version_comment` to `select @@version` and use `IsMariaDB`?

I think to change to `version` it better. @WangXiangUSTC what's your opinion?

dm/worker/config_test.go
```diff
@@ -22,6 +22,10 @@ import (
 	. "github.com/pingcap/check"
 
 	"github.com/pingcap/dm/dm/config"
+	"github.com/pingcap/dm/pkg/conn"
+
+	"github.com/DATA-DOG/go-sqlmock"
+	"github.com/siddontang/go-mysql/mysql"
```
csuzhangxc: you can put these two lines below L22.

dm/worker/config.go
```diff
@@ -239,6 +247,29 @@ func (c *Config) configFromFile(path string) error {
...
+		return err
+	}
+	fromDB, err := applyNewBaseDB(clone.From)
+	if err != nil {
+		return err
+	}
+	defer fromDB.Close()
+
+	c.Flavor, err = utils.GetFlavor(fromDB.DB)
+	return err
```
csuzhangxc: use `terror.WithScope` to annotate it?

pkg/utils/db.go
```diff
@@ -37,6 +37,26 @@ var (
 	domainServerIDSeparator = "-"
 )
 
+// GetFlavor gets flavor from DB
+func GetFlavor(db *sql.DB) (string, error) {
+	query := "SELECT @@version_comment"
+	row := db.QueryRow(query)
```
GregoryIan: where to close `row`?

dm/worker/config.go
```diff
@@ -239,6 +247,29 @@ func (c *Config) configFromFile(path string) error {
 	return c.verify()
 }
 
+// adjustFlavor adjusts flavor through querying from given database
+func (c *Config) adjustFlavor() error {
+	if c.Flavor != "" {
+		if c.Flavor != mysql.MariaDBFlavor && c.Flavor != mysql.MySQLFlavor {
```
GregoryIan: use `switch`

pkg/utils/db.go
```diff
@@ -37,6 +37,26 @@ var (
...
+		return "", terror.DBErrorAdapt(err, terror.ErrDBDriverError)
+	}
+	lowercaseVersionComment := strings.ToLower(versionComment)
+	switch {
+	case strings.Contains(lowercaseVersionComment, gmysql.MariaDBFlavor):
+		return gmysql.MariaDBFlavor, nil
+	case strings.Contains(lowercaseVersionComment, gmysql.MySQLFlavor):
+		return gmysql.MySQLFlavor, nil
+	default:
+		return "", terror.ErrNotSupportedFlavor.Generate(versionComment)
```
GregoryIan: consider that default return mysql

dm/worker/config.go
```diff
@@ -250,10 +250,14 @@ func (c *Config) configFromFile(path string) error {
 // adjustFlavor adjusts flavor through querying from given database
 func (c *Config) adjustFlavor() error {
 	if c.Flavor != "" {
-		if c.Flavor != mysql.MariaDBFlavor && c.Flavor != mysql.MySQLFlavor {
+		switch c.Flavor {
+		case mysql.MariaDBFlavor:
+			return nil
+		case mysql.MySQLFlavor:
+			return nil
```
csuzhangxc: you can combine these two `case` into one, like `case mysql.MariaDBFlavor, mysql.MySQLFlavor:`.

lichunzhu: Revised.

WangXiangUSTC:
> > So shall we change `select @@version_comment` to `select @@version` and use `IsMariaDB`?
> 
> I think to change to `version` it better. @WangXiangUSTC what's your opinion?

since we have already some pkg function is tidb-tools, I think it's better to resue it

dm/worker/config.go
```diff
@@ -239,6 +247,31 @@ func (c *Config) configFromFile(path string) error {
...
+	if err != nil {
+		return err
+	}
+	fromDB, err := applyNewBaseDB(clone.From)
+	if err != nil {
+		return terror.WithScope(err, terror.ScopeUpstream)
+	}
+	defer fromDB.Close()
+
+	c.Flavor, err = utils.GetFlavor(fromDB.DB)
```
csuzhangxc: Is there any timeout needed for the DB operation?

dm/worker/config.go
```diff
@@ -262,13 +271,19 @@ func (c *Config) adjustFlavor() error {
...
+	from := clone.From
+	from.RawDBCfg = config.DefaultRawDBConfig().SetReadTimeout(flavorReadTimeout)
+	fromDB, err := applyNewBaseDB(from)
 	if err != nil {
 		return terror.WithScope(err, terror.ScopeUpstream)
 	}
 	defer fromDB.Close()
 
-	c.Flavor, err = utils.GetFlavor(fromDB.DB)
+	ctx, _ := context.WithTimeout(context.Background(), flavorGetTimeout)
```
csuzhangxc: It's better to record `cancel` and add `defer cancel()`. comment from `context.WithTimeout`:
```
// Canceling this context releases resources associated with it, so code should
// call cancel as soon as the operations running in this Context **complete:**
```

dm/worker/config.go
```diff
@@ -262,13 +271,19 @@ func (c *Config) adjustFlavor() error {
...
+	c.Flavor, err = utils.GetFlavor(ctx, fromDB.DB)
+	if ctx.Err() != nil {
+		err = terror.Annotate(err, "time cost to get flavor info exceeds flavorGetTimeout")
+	}
```
csuzhangxc: Can these lines be removed? and only use the L287?

If you want to use this message, it's better to replace `flavorGetTimeout` with the accurate timeout value.

lichunzhu: These lines are added because when ctx is canceled because of timeout the returned error will just be like `context deadline exceeded` which lacks much information.

csuzhangxc: fine. so how about updating `flavorGetTimeout` to a meaningful value?

lichunzhu: Sure, I will revise that.

dm/worker/config.go
```diff
@@ -279,10 +279,11 @@ func (c *Config) adjustFlavor() error {
 	}
 	defer fromDB.Close()
 
-	ctx, _ := context.WithTimeout(context.Background(), flavorGetTimeout)
+	ctx, cancel := context.WithTimeout(context.Background(), flavorGetTimeout)
+	defer cancel()
 	c.Flavor, err = utils.GetFlavor(ctx, fromDB.DB)
 	if ctx.Err() != nil {
-		err = terror.Annotate(err, "time cost to get flavor info exceeds flavorGetTimeout")
+		err = terror.Annotate(err, "time cost to get flavor info exceeds "+flavorGetTimeout.String())
```
csuzhangxc: use `terror.Annotatef`?

dm/worker/config.go
```diff
@@ -239,6 +256,38 @@ func (c *Config) configFromFile(path string) error {
...
+	if err != nil {
+		return terror.WithScope(err, terror.ScopeUpstream)
+	}
+	defer fromDB.Close()
+
+	ctx, cancel := context.WithTimeout(context.Background(), flavorGetTimeout)
+	defer cancel()
+	c.Flavor, err = utils.GetFlavor(ctx, fromDB.DB)
+	if ctx.Err() != nil {
+		err = terror.Annotatef(err, "time cost to get flavor info exceeds %v in adjustFlavor", flavorGetTimeout.String())
```
csuzhangxc: Can this work?
```suggestion
		err = terror.Annotatef(err, "time cost to get flavor info exceeds %s", flavorGetTimeout)
```

PR #266, opened 2019-09-02
dm-worker: refine conn

3pointer:
### What problem does this PR solve? 
Optimize Database connections in dm-worker units, after unify DB operations

### What is changed and how it works?
1. one worker use one individual connection
2. same functional connection comes from one DB object
3. define some package for DB object and connections, all DB object generate in `conn.DBProvider.Apply`
4. split upstream DB connection and downstream(worker) DB connection, because they have totally different DB operations, it's better use `*sql.DB` in upstream, and use individual connection for each worker in downstream.
5. define two basic object, BaseDB and BaseConn, 
 BaseDB generates BaseConn, BaseConn wraps sql.Conn with own retry policy
 BaseDB -> BaseConn correspond to sql.DB -> sql.Conn
 In our scenario, there are two main reasons why we need BaseConn
   1. we often need one fixed DB connection to execute sql
   2. we need own retry policy during execute failed
 So we split a fixed sql.Conn out of sql.DB, and wraps it to BaseConn
 And Similar with sql.Conn, all BaseConn generated from one BaseDB shares this BaseDB to reset

### Check List 

Tests 

 - Unit test
 - Integration test
 - Manual test (add detailed scripts or steps below)
 - No code

Code changes

 - Has exported function/method change
 - Has exported variable/fields change
 - Has interface methods change
 - Has persistent data change

Side effects

 - Possible performance regression
 - Increased code complexity
 - Breaking backward compatibility

Related changes

 - Need to cherry-pick to the release branch
 - Need to update the documentation
 - Need to update the `dm/dm-ansible`
 - Need to be included in the release note

PR #325, opened 2019-10-18
dm-worker: refine conn (#266)

3pointer:
cherry-pick this [PR](https://github.com/pingcap/dm/pull/266/) to release-1.0

Optimize Database connections in dm-worker units, after unify DB operations

1. one worker use one individual connection
2. same functional connection comes from one DB object
3. define some package for DB object and connections, all DB object generate in `conn.DBProvider.Apply`
4. split upstream DB connection and downstream(worker) DB connection, because they have totally different DB operations, it's better use `*sql.DB` in upstream, and use individual connection for each worker in downstream.
5. define two basic object, BaseDB and BaseConn, 
 BaseDB generates BaseConn, BaseConn wraps sql.Conn with own retry policy
 BaseDB -> BaseConn correspond to sql.DB -> sql.Conn
 In our scenario, there are two main reasons why we need BaseConn
   1. we often need one fixed DB connection to execute sql
   2. we need own retry policy during execute failed
 So we split a fixed sql.Conn out of sql.DB, and wraps it to BaseConn
 And Similar with sql.Conn, all BaseConn generated from one BaseDB shares this BaseDB to reset

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes


csuzhangxc:
how about changing the title (and the commit message later) to `dm-worker: refine conn (#266)`

csuzhangxc:

PR #328, opened 2019-10-21
worker: generate index flavor automatically by querying version info from DB (#323)

lichunzhu:
cherry-pick #323 to release-1.0

---

[TOOL-1166](https://internal.pingcap.net/jira/browse/TOOL-1166)
Now, we need to set flavor for DM-worker manually, but it may be set incorrectly.

use `dbutil.ShowVersion` and `check.IsMariaDB` from [tidb-tools](https://github.com/pingcap/tidb-tools) to detect upstream MySQL/MariaDB server type and set flavor automatically.

### Check List 

Tests 

 - Unit test
 - Integration test

Related changes

### DIFF
--- a/dm/worker/config.go
+++ b/dm/worker/config.go
@@ -15,17 +15,21 @@ package worker
 
 import (
 	"bytes"
+	"context"
 	"encoding/base64"
 	"encoding/json"
 	"flag"
 	"fmt"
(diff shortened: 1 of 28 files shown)


csuzhangxc:
@WangXiangUSTC PTAL

PR #330, opened 2019-10-22
*: only retry for special message with 1105

csuzhangxc:

`1105` means an unknown error in MySQL/TiDB. some of these errors are retryable, but some are not (like `error="Error 1105: statement count 5001 exceeds the transaction limitation`).

only retry for retryable errors, now including:
- `Information schema is out of date`
- `Information schema is changed`

### Check List 

Tests 

 - Unit test

Related changes

### DIFF
--- a/syncer/db.go
+++ b/syncer/db.go
@@ -238,7 +238,7 @@ func (conn *DBConn) executeSQLWithIgnore(tctx *tcontext.Context, ignoreError fun
 				sqlRetriesTotal.WithLabelValues("stmt_exec", conn.cfg.Name).Add(1)
 				return true
 			}
-			if retry.IsRetryableErrorFastFailFilter(err) {
+			if retry.IsRetryableError(err) {
 				tctx.L().Warn("execute statements", zap.Int("retry", retryTime),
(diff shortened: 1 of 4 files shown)


csuzhangxc:

syncer/db.go
```diff
@@ -238,7 +238,7 @@ func (conn *DBConn) executeSQLWithIgnore(tctx *tcontext.Context, ignoreError fun
 				sqlRetriesTotal.WithLabelValues("stmt_exec", conn.cfg.Name).Add(1)
 				return true
 			}
-			if retry.IsRetryableErrorFastFailFilter(err) {
+			if retry.IsRetryableError(err) {
```
WangXiangUSTC: why change to this function? seems change retry logic for ddl

csuzhangxc: `IsRetryableErrorFastFailFilter` as a black-list for `1105`, but we have changed to use a white-list for `1105` now.

so, the _fast fail error_ in `IsRetryableErrorFastFailFilter` will not be retry in `IsRetryableError`.

WangXiangUSTC:
LGTM

PR #331, opened 2019-10-22
DM-worker: fix toml decode time.duration in config

amyangfei:

`time.Duration` is type alias for int64, the toml library we use doesn't support decode `time.Duration` directly

Create a new `duration` struct which embeds the `time.Duration`
- Hack the `UnmarshalText` method for toml decoding
- Hack the `MarshalJSON` for json format string

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

Related changes

   - Optimize time.Duration type parameter in the configuration of DM-worker


dm/worker/task_checker.go
```diff
@@ -90,21 +91,47 @@ func (bs ResumeStrategy) String() string {
...
+func (d duration) UnmarshalText(text []byte) error {
+	var err error
+	d.Duration, err = time.ParseDuration(string(text))
+	return err
+}
+
+// MarshalJSON hacks to satisfy the json.Marshaler interface
+func (d *duration) MarshalJSON() ([]byte, error) {
+	return json.Marshal(&struct {
+		Duration string `json:"duration"`
```
csuzhangxc: we have JSON flag in L96, is this still needed?

amyangfei: yes, the JSON tag here is must, which is really used during JSON marshal

csuzhangxc: can we remove this `MarshalJSON` method and only rely on the JSON flag in L96?

amyangfei: If we don't implement this method, the JSON output will be `int64` value of `Duratioin`  in nanosecond level, which is not human readable

csuzhangxc:
@WangXiangUSTC PTAL

dm/worker/task_checker.go
```diff
@@ -90,21 +91,47 @@ func (bs ResumeStrategy) String() string {
 	return fmt.Sprintf("unsupported resume strategy: %d", bs)
 }
 
+// duration is used to hold a time.Duration field
+type duration struct {
+	time.Duration `toml:"duration" json:"duration"`
```
WangXiangUSTC: why need `toml:"duration"`?

amyangfei: removed

amyangfei:
PTAL @WangXiangUSTC

PR #355, opened 2019-11-14
loader/syncer: filter context cancel error while executing sqls

lichunzhu:
When an error occurs and the context in loader/syncer is canceled, the normally running sql will be canceled and return a `context.Canceled` error, which will confuse the users.

Filter all the `context.Canceled` in loader/syncer. Don't return error and don't print log.

### Check List 

Tests 

 - Unit test
 - Integration test

Related changes


lichunzhu:

loader/db.go
```diff
@@ -95,10 +95,12 @@ func (conn *DBConn) querySQL(ctx *tcontext.Context, query string, args ...interf
 			return ret, err
 		})
 	if err != nil {
-		ctx.L().Error("query statement failed after retry",
-			zap.String("query", utils.TruncateString(query, -1)),
-			zap.String("argument", utils.TruncateInterface(args, -1)),
-			log.ShortError(err))
+		if !utils.IsContextCanceledError(err) {
```
csuzhangxc: 1. should we need to filter other logs in this file? like
```go
err = conn.resetConn(ctx)
if err != nil {
    ctx.L().Error("reset connection failed", zap.Int("retry", retryTime),
        zap.String("query", utils.TruncateInterface(query, -1)),
        zap.String("arguments", utils.TruncateInterface(args, -1)),
        log.ShortError(err))
    return false
}
```

can this `err` also become `ContextCanceled`.

2. how about adding a method in `pkg/log/Logger` to do this check (err != ***)?

loader/loader.go
```diff
@@ -174,7 +174,11 @@ func (w *Worker) run(ctx context.Context, fileJobQueue chan *fileJob, runFatalCh
 				if err != nil {
 					// expect pause rather than exit
 					err = terror.WithScope(terror.Annotatef(err, "file %s", job.file), terror.ScopeDownstream)
-					runFatalChan <- unit.NewProcessError(pb.ErrorType_ExecSQL, err)
+					if utils.IsContextCanceledError(err) {
+						runFatalChan <- nil
```
csuzhangxc: Is this needed?

lichunzhu: If we don't deal with `runFatalChan`, the user will still get cancel error from dmctl. Shall we revise `unit.NewProcessError` to filter that instead of this?

WangXiangUSTC: if err is context cancel error, don't need to send nil to `runFatalChan`

pkg/log/log.go
```diff
@@ -130,7 +132,7 @@ func SetLevel(level zapcore.Level) zapcore.Level {
 // just repeats known information. You should almost always use `ShortError`
 // instead of `zap.Error`, unless the error is no longer propagated upwards.
 func ShortError(err error) zap.Field {
-	if err == nil {
+	if err == nil || errors.Cause(err) == context.Canceled {
```
csuzhangxc: which one is better:
1. update `ShortError`
2. add a new function

@WangXiangUSTC what's your opinion?

WangXiangUSTC: I think 2 is better

lichunzhu: Do you mean put `err == nil || errors.Cause(err) == context.Canceled`... into a new function?
@WangXiangUSTC

WangXiangUSTC: how about adding a new function to judge  whether an error can be filter
```
func FilterError(err error) error {
    if errors.Cause(err) == context.Canceled {
         return nil
    }
    return err
}
```

then we can use `ShortError(FilterError(err))`

csuzhangxc: approve the method from @WangXiangUSTC, but maybe a meaningful name needed?

lichunzhu: How about `FilterCancelError`?

csuzhangxc: `FilterCancelError` LGTM

lichunzhu: which one is better:
1. return `error`. the usage would be `log.ShortError(log.FilterCancelError(err))`
2. return `zap.Field`. the usage would be `log.FilterCancel(err)`
@csuzhangxc

lichunzhu:
@csuzhangxc Revised, PTAL

pkg/log/log.go
```diff
@@ -136,6 +138,14 @@ func ShortError(err error) zap.Field {
 	return zap.String("error", err.Error())
 }
 
+// FilterCancelError will skip context.Canceled error
+func FilterCancelError(err error) error {
+	if errors.Cause(err) == context.Canceled {
```
WangXiangUSTC: how about use `IsContextCanceledError`?

lichunzhu:
@csuzhangxc Revised, PTAL

PR #359, opened 2019-11-15
mydumper/: change Equal to HasPrefix in func needToGenerateDoTables

lichunzhu:
Current `needToGenerateDoTables` can not identify args like `--regex=aaa` and will generate do table args to cover the args specified by users.

Change `=` to `strings.HasPrefix`. Add more terror infomation.

### Check List 

Tests 

 - Unit test
 - Integration test

Related changes


lichunzhu:
@csuzhangxc @WangXiangUSTC PTAL.

PR #365, opened 2019-11-20
dmctl/:Clear workers cache before we run a command

lichunzhu:
1. Every time we run command in dmctl with -w arg the argument worker will be accumulated which doesn't match our expectations.
image
2. After we set flag to rootCmd, the value will be kept. Although we don't specify this arg in the next time. it still has a value.
image

1. Clear workers array every time before we run a command.
2. New rootCmd every time before we run a command. I tried `rootCmd.ResetFlags()` but it will remove all flags and we can not specify any arg any more. Is there any better solution?

### Check List 

Tests 

 - Unit test
 - Integration test
1. image
2. image

Related changes

### DIFF
--- a/dm/ctl/ctl.go
+++ b/dm/ctl/ctl.go
@@ -28,10 +28,7 @@ import (
 
 var (
 	commandMasterFlags = CommandMasterFlags{}
-	rootCmd            = &cobra.Command{
-		Use:   "dmctl",
-		Short: "DM control",
-	}
+	rootCmd            *cobra.Command
 )
 
 // CommandMasterFlags are flags that used in all commands for dm-master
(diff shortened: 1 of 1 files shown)


dm/ctl/ctl.go
```diff
@@ -116,6 +116,7 @@ func PrintHelp(args []string) {
 
 // Start starts running a command
 func Start(args []string) {
+	commandMasterFlags.workers = commandMasterFlags.workers[:0]
```
csuzhangxc: add a method like `Reset` for `CommandMasterFlags` to do this? (we may add more fields into `CommandMasterFlags` later)

lichunzhu:
DNM now. I find that the other local args also have the same problem. I will try to fix them in this PR.

PR #367, opened 2019-11-20
dm-master: leader election

csuzhangxc:

add leader election supported for DM-master based on embed etcd

do leader election based on embed etcd

some code learn from https://github.com/pingcap/tidb/blob/v3.0.5/owner/manager.go.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

### DIFF
--- a/dm/master/election.go
+++ b/dm/master/election.go
@@ -35,7 +35,7 @@ func (s *Server) electionNotify(ctx context.Context) {
 				if err2 == nil {
 					log.L().Info("current member retire from the leader", zap.String("leader", leaderID), zap.String("current member", s.cfg.Name))
 				} else {
-					log.L().Error("get leader info", zap.Error(err2))
+					log.L().Warn("get leader info", zap.Error(err2))
 				}
 			}
 		case err := <-s.election.ErrorNotify():
--- a/dm/master/election.go
+++ b/dm/master/election.go
@@ -38,6 +38,12 @@ func (s *Server) electionNotify(ctx context.Context) {
 					log.L().Error("get leader info", zap.Error(err2))
 				}
 			}
+		case err := <-s.election.ErrorNotify():
+			// handle errors here, we do no meaningful things now.
+			// but maybe:
+			// 1. trigger an alert
+			// 2. shutdown the DM-master process
+			log.L().Error("receive error from election", zap.Error(err))
 		}
 	}
 }
--- a/dm/master/election.go
+++ b/dm/master/election.go
@@ -1,0 +1,43 @@
+// Copyright 2019 PingCAP, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package master
+
+import (
+	"context"
+
+	"go.uber.org/zap"
+
+	"github.com/pingcap/dm/pkg/log"
+)
+
+func (s *Server) electionNotify(ctx context.Context) {
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case leader := <-s.election.LeaderNotify():
+			// output the leader info.
+			if leader {
+				log.L().Info("current member become the leader", zap.String("current member", s.cfg.Name))
+			} else {
+				_, leaderID, err2 := s.election.LeaderInfo(ctx)
+				if err2 == nil {
+					log.L().Info("current member retire from the leader", zap.String("leader", leaderID), zap.String("current member", s.cfg.Name))
+				} else {
+					log.L().Error("get leader info", zap.Error(err2))
+				}
+			}
+		}
+	}
+}
--- a/dm/master/server.go
+++ b/dm/master/server.go
@@ -25,6 +25,7 @@ import (
 
 	"github.com/pingcap/errors"
 	"github.com/siddontang/go/sync2"
+	"go.etcd.io/etcd/clientv3"
 	"go.etcd.io/etcd/embed"
 	"go.uber.org/zap"
 	"google.golang.org/grpc"
@@ -34,10 +35,21 @@ import (
 	operator "github.com/pingcap/dm/dm/master/sql-operator"
 	"github.com/pingcap/dm/dm/master/workerrpc"
 	"github.com/pingcap/dm/dm/pb"
+	"github.com/pingcap/dm/pkg/election"
+	"github.com/pingcap/dm/pkg/etcdutil"
 	"github.com/pingcap/dm/pkg/log"
 	"github.com/pingcap/dm/pkg/terror"
 	"github.com/pingcap/dm/pkg/tracing"
 	"github.com/pingcap/dm/syncer"
+)
+
+const (
+	// the session's TTL in seconds for leader election.
+	// NOTE: select this value carefully when adding a mechanism relying on leader election.
+	electionTTL = 60
+	// the DM-master leader election key prefix
+	// DM-master cluster : etcd cluster = 1 : 1 now.
+	electionKey = "/dm-master/leader"
 )
 
 var (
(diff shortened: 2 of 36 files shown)


csuzhangxc:
@WangXiangUSTC , @GregoryIan or @amyangfei   PTAL

csuzhangxc:
I update the methods for `Election` after added `PTAL` label. It's ready for review again now.

tests/others_integration.txt
```diff
@@ -1,7 +1 @@
-all_mode
-sequence_sharding
-sequence_safe_mode
-relay_interrupt
 start_task
-initial_unit
-http_apis
```
csuzhangxc: I have updated the CI to run these cases separately

dm/master/election.go
```diff
@@ -0,0 +1,43 @@
...
+
+import (
+	"context"
+
+	"go.uber.org/zap"
+
+	"github.com/pingcap/dm/pkg/log"
+)
+
+func (s *Server) electionNotify(ctx context.Context) {
```
WangXiangUSTC: why need this function, seems `toBeLeader` and `retireLeader` already print the log

pkg/election/election.go
```diff
@@ -0,0 +1,251 @@
...
+func (e *Election) retireLeader() {
+	e.isLeader.Set(false)
+	select {
+	case e.leaderCh <- false:
+	default:
+	}
+	e.l.Info("retire from the leader", zap.String("member", e.id))
+}
+
+func (e *Election) watchLeader(ctx context.Context, session *concurrency.Session, key string) {
```
WangXiangUSTC: the key can only be e.key? if so don't need put it in arguments.

dm/master/server.go
```diff
@@ -151,14 +166,30 @@ func (s *Server) Start(ctx context.Context) (err error) {
...
+	// NOTE: we only use the local member's address now, but we can use all endpoints of the cluster if needed.
+	s.etcdClient, err = etcdutil.CreateClient([]string{s.cfg.MasterAddr})
+	if err != nil {
+		return
+	}
+
 	s.closed.Set(false) // the server started now.
 
+	// start leader election
+	s.election = election.NewElection(ctx, s.etcdClient, electionTTL, electionKey, s.cfg.Name)
```
WangXiangUSTC: do we need to handle `election`'s error from `election.ErrorNotify()`

pkg/election/election.go
```diff
@@ -0,0 +1,251 @@
...
+				return
+			}
+			if resp.Canceled {
+				e.l.Info("watch canceled")
+				return
+			}
+
+			for _, ev := range resp.Events {
+				if ev.Type == mvccpb.DELETE {
+					e.l.Info("fail to watch, the leader is deleted", zap.ByteString("key", ev.Kv.Key))
```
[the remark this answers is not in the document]
csuzhangxc: user may force to trigger a new campaign by deleting the key with _etcdctl_ (or _dmctl_ in the future); in the normal case, before electionTTL overdue, `Session` will `keep alive for the lifetime of a client`.

see:
- https://github.com/etcd-io/etcd/blob/v3.4.3/clientv3/concurrency/session.go#L26
- https://github.com/etcd-io/etcd/blob/v3.4.3/clientv3/concurrency/session.go#L63

dm/master/election.go
csuzhangxc: `toBeLeader` and `retireLeader` only log the leadership of the current member.
this function is used to show the usage of `Election` (and also log the leader name even if it's not the current member), do you think should we need to remove it?

pkg/election/election.go
csuzhangxc: `e.key` is the prefix of this `key`, like `e.key == /dm-master/leader`, but `key == /dm-master/leader/105a6e8c8aee0806` (with the `lease` as the suffix), and this key can change for every campaign loop.

dm/master/server.go
csuzhangxc: I update `Election` to let it more like which one in TiDB's owner pkg in https://github.com/pingcap/dm/pull/367/commits/80ae3d23a14157b78fdc4b7ac2a16e424ad18fef and https://github.com/pingcap/dm/pull/367/commits/5c4c8204aeb8eca110439574ea63596c6901752b.

1. Detect any potential error (for unrecoverable error) in `NewElection`, and do not start `Server` if an error occurred.
2. Retry for errors (for recoverable error) until the context is done after entering the campaign loop.
3. receive errors from `election.ErrorNotify()` in _server/election.go_, but do no meaningful things now.

csuzhangxc:
@WangXiangUSTC PTAL again

dm/master/election.go
csuzhangxc: one more thing, `pkg/election` is a pkg, but this function is in the application. Maybe we need to add some metrics for the leadership later, and do that in this function is better.

pkg/election/election.go
WangXiangUSTC: ok

pkg/election/election.go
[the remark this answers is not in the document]
WangXiangUSTC: how about adding a comment in which case will get delete event

dm/master/election.go
WangXiangUSTC: if master needs to do something(for example: meet error in the election, master exit) according to the election's result I think putting here is better. 
it is up to you

pkg/election/election.go
```diff
@@ -0,0 +1,284 @@
...
+		if i > 0 {
+			select {
+			case e.ech <- terror.ErrElectionCampaignFail.Delegate(err, "create a new session"):
+			default:
+			}
+
+			select {
+			case <-time.After(newSessionRetryInterval):
+			case <-ctx.Done():
+				break
```
WangXiangUSTC: do you means break the for loop, this break only break the `select`
https://stackoverflow.com/questions/11104085/in-go-does-a-break-statement-break-from-a-switch-select

pkg/election/election.go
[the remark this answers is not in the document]
csuzhangxc: added in https://github.com/pingcap/dm/pull/367/commits/b99ae4653c5039cfb5070e5b0d62d6bdad77599a.

dm/master/election.go
csuzhangxc: I chose to remove the log in `pkg/election` in https://github.com/pingcap/dm/pull/367/commits/b99ae4653c5039cfb5070e5b0d62d6bdad77599a.

pkg/election/election.go
csuzhangxc: good catch!, fixed in https://github.com/pingcap/dm/pull/367/commits/b99ae4653c5039cfb5070e5b0d62d6bdad77599a.

WangXiangUSTC:
LGTM

dm/master/election.go
```diff
@@ -0,0 +1,49 @@
...
+		case leader := <-s.election.LeaderNotify():
+			// output the leader info.
+			if leader {
+				log.L().Info("current member become the leader", zap.String("current member", s.cfg.Name))
+			} else {
+				_, leaderID, err2 := s.election.LeaderInfo(ctx)
+				if err2 == nil {
+					log.L().Info("current member retire from the leader", zap.String("leader", leaderID), zap.String("current member", s.cfg.Name))
+				} else {
+					log.L().Error("get leader info", zap.Error(err2))
```
amyangfei: Is `Warn` level better

pkg/election/election.go
```diff
@@ -0,0 +1,285 @@
...
+			case <-ctx.Done():
+				break forLoop
+			}
+		}
+
+		// add more options if needed.
+		// NOTE: I think use the client's context is better than something like `concurrency.WithContext(ctx)`,
+		// so we can close the session when the client is still valid.
+		session, err = concurrency.NewSession(e.cli, concurrency.WithTTL(e.sessionTTL))
+		if err == nil || errors.Cause(err) == ctx.Err() {
```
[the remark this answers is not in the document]
csuzhangxc: updated in https://github.com/pingcap/dm/pull/367/commits/708c2c1288d3bd3a379399d6e22f76572488b9fb.

dm/master/election.go
csuzhangxc: updated in https://github.com/pingcap/dm/pull/367/commits/708c2c1288d3bd3a379399d6e22f76572488b9fb.

PR #372, opened 2019-11-21
checker/:ignore unnecessary privilege for different task modes

lichunzhu:
Now we check all the privileges when we start dm which is unreasonable. For task mode **full** we don't need **replication** privilege while for task mode **increment** we don't need **dump** privilege.

Add unnecessary privilege check to ignore checking items for different task modes.

### Check List 

Tests 

 - Unit test
 - Integration test

Related changes

### DIFF
--- a/checker/cmd.go
+++ b/checker/cmd.go
@@ -39,8 +39,16 @@ func CheckSyncConfig(ctx context.Context, cfgs []*config.SubTaskConfig) error {
 		return nil
 	}
 
-	// all `IgnoreCheckingItems` of sub-task are same, so we take first one
-	checkingItems := config.FilterCheckingItems(cfgs[0].IgnoreCheckingItems)
+	// all `IgnoreCheckingItems` and `Mode` of sub-task are same, so we take first one
+	// for ModeFull we don't need replication privilege; for ModeIncrement we don't need dump privilege
+	ignoreCheckingItems := cfgs[0].IgnoreCheckingItems
+	switch cfgs[0].Mode {
+	case config.ModeFull:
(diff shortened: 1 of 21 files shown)


checker/cmd.go
```diff
@@ -39,8 +39,16 @@ func CheckSyncConfig(ctx context.Context, cfgs []*config.SubTaskConfig) error {
...
 	}
 
-	// all `IgnoreCheckingItems` of sub-task are same, so we take first one
-	checkingItems := config.FilterCheckingItems(cfgs[0].IgnoreCheckingItems)
+	// all `IgnoreCheckingItems` and `Mode` of sub-task are same, so we take first one
+	// for ModeFull we don't need replication privilege; for ModeIncrement we don't need dump privilege
+	ignoreCheckingItems := cfgs[0].IgnoreCheckingItems
+	switch cfgs[0].Mode {
+	case config.ModeFull:
+		ignoreCheckingItems = append(ignoreCheckingItems, config.ReplicationPrivilegeChecking)
```
csuzhangxc: 1. for `full`, more privileges can be ignored, including `BinlogEnableChecking`, `BinlogFormatChecking`, and `BinlogRowImageChecking`.
2. if the user already specified `ReplicationPrivilegeChecking` in `ignoreCheckingItems`, duplicate `append` will have no side effect, right?

lichunzhu: 1. I will add it.
2. I have thought that before. For this case the duplicate `append` will make map `checkingItems`  delete nonexistent key, which is supported for **golang**. It might be a bit little tricky. Shall we check it before adding?

csuzhangxc: for 2, add a comment to explain it?

lichunzhu:
@csuzhangxc All comments addressed. PTAL.

PR #377, opened 2019-11-27
*: fix context usage for SQL operation

csuzhangxc:

when retrying SQL operation to downstream, `stop-task` / `pause-task` may not take effect because of the misuse of _context_.

fix context usage for SQL operation to downstream, including:
- data migration
- checkpoint update
- online DDL meta update

**NOTE**:

- shard DDL meta operation not updated in this PR because we are refactoring it
- better improvement for these operations can be done when moving them to etcd.
- `retry_cancel` integration testing has much wait time, and I'll split it run separately in CI after received 2 LGTM.

### Check List 

Tests 

 - Unit test
 - Integration test

Side effects

 - Increased code complexity

Related changes


csuzhangxc:
@WangXiangUSTC @GregoryIan PTAL

loader/checkpoint.go
```diff
@@ -80,13 +80,14 @@ type RemoteCheckPoint struct {
 }
 
 func newRemoteCheckPoint(tctx *tcontext.Context, cfg *config.SubTaskConfig, id string) (CheckPoint, error) {
-	db, dbConns, err := createConns(tctx, cfg, 1)
+	tctx2, cancel := tctx.WithTimeout(defaultDBContextTimeout)
+	defer cancel()
+
+	db, dbConns, err := createConns(tctx2, cfg, 1)
```
GregoryIan: does context with timeout affect `resetBaseConnFn` in `createConns`?

loader/checkpoint.go
```diff
@@ -95,36 +96,36 @@ func newRemoteCheckPoint(tctx *tcontext.Context, cfg *config.SubTaskConfig, id s
 		finishedTables: make(map[string]struct{}),
 		schema:         cfg.MetaSchema,
 		table:          fmt.Sprintf("%s_loader_checkpoint", cfg.Name),
-		tctx:           newtctx,
+		tctx:           tctx.WithLogger(tctx.L().WithFields(zap.String("component", "remote checkpoint"))),
 	}
 
-	err = cp.prepare()
+	err = cp.prepare(tctx2)
```
GregoryIan: is `defaultDBContextTimeout(10 seconds)` enough?

GregoryIan: sometimes tidb is not fast, espicallly DDL

loader/checkpoint.go
csuzhangxc: No, `tctx` in `resetBaseConnFn` is passed in (it's a function argument) when reseting connetion like in `func (conn *DBConn) resetConn(tctx *tcontext.Context) error`.

this timeout context only used to `GetBaseConn` and log something.

loader/checkpoint.go
csuzhangxc: In fact, I'm not sure, but this timeout only used for some checkpoint operations and they may not expensive (and if timeout, they can be retried through restring the task). Are there any suggestions for me?

BTW, "better improvement for these operations can be done when moving them to etcd"  in the PR's description, and it's better to add `context` support for all I/O (network) operations later, then we can control the timeout in the upper caller.

pkg/streamer/reader_test.go
```diff
@@ -290,7 +290,7 @@ func (t *testReaderSuite) TestParseFileRelaySubDirUpdated(c *C) {
 		err2 := ioutil.WriteFile(nextPath, replication.BinLogFileHeader, 0600)
 		c.Assert(err2, IsNil)
 	}()
-	ctx3, cancel3 := context.WithTimeout(context.Background(), time.Second)
+	ctx3, cancel3 := context.WithTimeout(context.Background(), 2*time.Second)
```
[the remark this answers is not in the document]
csuzhangxc: YES, sometimes, the case may fail because of timeout...

syncer/syncer.go
```diff
@@ -856,7 +862,16 @@ func (s *Syncer) flushCheckPoints() error {
...
+
+	if utils.IsContextCanceledError(tctx.Ctx.Err()) {
+		// when canceling (stop-task/pause-task), we still need to flush the checkpoint.
+		var cancel context.CancelFunc
+		tctx, cancel = tcontext.Background().WithTimeout(defaultDBContextTimeout)
+		defer cancel()
+		s.tctx.L().Info("flushing checkpoint when canceling the task", zap.Stringer("checkpoint", s.checkpoint))
+	}
+
+	err := s.checkpoint.FlushPointsExcept(tctx, exceptTables, shardMetaSQLs, shardMetaArgs)
```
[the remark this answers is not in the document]
csuzhangxc: I'm not sure whether needing different timeout values (rather than the default one used when canceling) here.

Do you prefer to remove the `tctx` argument here?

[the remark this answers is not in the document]
WangXiangUSTC: yes, how about @GregoryIan

csuzhangxc:
@WangXiangUSTC @GregoryIan PTAL again.

NOTE: I do not rename `tctx` in syncer to `logCtx` now because there are many places using it, and I think we should update them one by one later.

GregoryIan:
LGTM

GregoryIan:
In addition to the db operation, are there any operations that will block dm-ctl operation?

GregoryIan:
I think we should add a chaos case to verify this pr, make sure it has enough full and incremental data and run long enough, while doing various operations(`stop-task`, `pause-task`, `query-task`) on it

syncer/syncer.go
```diff
@@ -856,7 +858,16 @@ func (s *Syncer) flushCheckPoints() error {
 		shardMetaSQLs, shardMetaArgs = s.sgk.PrepareFlushSQLs(exceptTableIDs)
 		s.tctx.L().Info("prepare flush sqls", zap.Strings("shard meta sqls", shardMetaSQLs), zap.Reflect("shard meta arguments", shardMetaArgs))
 	}
-	err := s.checkpoint.FlushPointsExcept(exceptTables, shardMetaSQLs, shardMetaArgs)
+
+	// when canceling (stop-task/pause-task), we still need to flush the checkpoint.
+	timeout, err := time.ParseDuration(maxDMLConnectionTimeout)
```
WangXiangUSTC: we can define a new variable with `time.Duration` type, and then don't need parse every time.

loader/checkpoint.go
```diff
@@ -349,7 +347,12 @@ func (cp *RemoteCheckPoint) Count() (int, error) {
 }
 
 func (cp *RemoteCheckPoint) String() string {
```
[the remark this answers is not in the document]
csuzhangxc: agree.

syncer/syncer.go
csuzhangxc: addressed in https://github.com/pingcap/dm/pull/377/commits/f59a6aace6dfb8ad49ea64d149674d56a1cb3ff9.

WangXiangUSTC:
LGTM

PR #380, opened 2019-11-28
config/task: add check for duplicate config items

lichunzhu:
The user may add duplicate route-rules or filter-rules by mistake which will cause failure while building router or event-filter.
e.g.: route-rules: ["route-rule-1", "route-rule-1"]

Add check in `TaskConfig.adjust` and returns error if route-rules/filter-rules are duplicate.

### Check List 

Tests 

 - Unit test
 - Integration test

Related changes

### DIFF
--- a/dm/config/task.go
+++ b/dm/config/task.go
@@ -443,6 +443,13 @@ func (c *TaskConfig) adjust() error {
 		if inst.SyncerThread != 0 {
 			inst.Syncer.WorkerCount = inst.SyncerThread
 		}
+
+		if dupeRule, hasDupe := checkDuplicateString(inst.RouteRules); hasDupe {
+			return terror.ErrConfigDuplicateCfgItem.Generate(i, "route-rules", dupeRule)
+		}
+		if dupeRule, hasDupe := checkDuplicateString(inst.FilterRules); hasDupe {
+			return terror.ErrConfigDuplicateCfgItem.Generate(i, "filter-rules", dupeRule)
+		}
 	}
 
 	if c.Timezone != "" {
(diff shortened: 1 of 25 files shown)


lichunzhu:
@csuzhangxc Revised, PTAL

dm/config/task_test.go
```diff
@@ -192,4 +202,67 @@ syncers:
...
+	err = ioutil.WriteFile(filepath, configContent, 0644)
+	c.Assert(err, IsNil)
+	taskConfig = NewTaskConfig()
+	err = taskConfig.DecodeFile(filepath)
+	c.Assert(err, IsNil)
+	taskConfig.MySQLInstances[0].RouteRules = []string{"route-rule-1", "route-rule-2", "route-rule-1", "route-rule-2"}
+	taskConfig.MySQLInstances[1].FilterRules = []string{"filter-rule-1", "filter-rule-2", "filter-rule-3", "filter-rule-2"}
+	err = taskConfig.adjust()
+	c.Assert(terror.ErrConfigDuplicateCfgItem.Equal(err), IsTrue)
+	c.Assert(err, ErrorMatches, "[\\s\\S]*mysql-instance\\(0\\)'s route-rules: route-rule-1, route-rule-2[\\s\\S]*")
```
csuzhangxc: you can use ``` ` ``` instead of ` " `to wrap a literal string, then it can be more readable.

csuzhangxc:
@WangXiangUSTC PTAL

dm/config/task.go
```diff
@@ -516,3 +529,21 @@ func (c *TaskConfig) SubTaskConfigs(sources map[string]DBConfig) ([]*SubTaskConf
...
 }
+
+// checkDuplicateString checks whether the given string array has duplicate string item
+// if there is duplicate, it will return **all** the duplicate strings
+func checkDuplicateString(ruleNames []string) []string {
+	mp := make(map[string]bool, len(ruleNames))
+	dupeArray := make([]string, 0)
+	for _, name := range ruleNames {
+		if added, ok := mp[name]; ok {
+			if !added {
```
WangXiangUSTC: seems that don't need judge `added`, if `ok` is true means this `ruleName` is duplicate

csuzhangxc: I think its used to prevent `append` more than once.

lichunzhu: added is used to avoid to print "duplicate" duplicates.
For example, ["a", "a", "a"], "a" will be appened for two times.

PR #381, opened 2019-11-28
loader/syncer: fix bug in #355

lichunzhu:
Son PR of https://github.com/pingcap/dm/pull/355.
`log.ShortError(err)` may contains extra msg for `context.Canceled`.

change `==` to `strings.Contains`

### Check List 

Tests 

 - Unit test
 - Integration test


pkg/log/log.go
```diff
@@ -76,7 +77,7 @@ func (l Logger) WithFields(fields ...zap.Field) Logger {
 // ErrorFilterContextCanceled wraps Logger.Error() and will filter error log when error is context.Canceled
 func (l Logger) ErrorFilterContextCanceled(msg string, fields ...zap.Field) {
 	for _, field := range fields {
-		if field.Key == "error" && field.String == context.Canceled.Error() {
+		if field.Key == "error" && strings.Contains(field.String, context.Canceled.Error()) {
```
csuzhangxc: Can this handle `zap.Error` (`filed.String` is empty) rather than `log.ShortError`?

lichunzhu: No, this can't handle. I have updated the dealing for such situations.

lichunzhu:
@csuzhangxc PTAL

pkg/log/log.go
```diff
@@ -76,9 +78,15 @@ func (l Logger) WithFields(fields ...zap.Field) Logger {
...
 			return
 		}
+		if field.Type == zapcore.ErrorType {
```
csuzhangxc: 1. How about using a `switch` to check `filed.Type` first?
2. Can you add a unit test case for this? ref <https://github.com/pingcap/tidb-lightning/blob/605760d1b2025d1e1a8b7d0c668c74863d7d1271/lightning/log/log_test.go#L43>.

pkg/log/testlogger.go
```diff
@@ -0,0 +1,36 @@
...
+
+package log
+
+import (
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+	"go.uber.org/zap/zaptest"
+)
+
+// MakeTestLogger creates a Logger instance which produces JSON logs.
```
WangXiangUSTC: how add a note `this function only used for test`?

WangXiangUSTC: or we can move it to test file

lichunzhu:
@WangXiangUSTC Revised. PTAL

WangXiangUSTC:
LGTM

PR #382, opened 2019-11-28
loader/syncer: filter context cancel error while executing sqls (#355)

lichunzhu:
cherry-pick #355 and #381 to release-1.0

---

When an error occurs and the context in loader/syncer is canceled, the normally running sql will be canceled and return a `context.Canceled` error, which will confuse the users.

Filter all the `context.Canceled` in loader/syncer. Don't return error and don't print log.

### Check List 

Tests 

 - Unit test
 - Integration test

Related changes


csuzhangxc:
@WangXiangUSTC PTAL

PR #394, opened 2019-12-03
*: register DM-worker instance to DM-master dynamically

csuzhangxc:

support register/unregister DM-worker instance to DM-master

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes


csuzhangxc:
better implementation in https://github.com/pingcap/dm/pull/431.

PR #399, opened 2019-12-04
dm-worker: support election with same source-id

WangXiangUSTC:

support election with the same source-id for dm-worker

1. add dm-masters' address config for dm-worker
2. use pkg/election to election leader on `source-id`

### Check List 

Tests 

 - Unit test


dm/master/config.go
```diff
@@ -242,7 +242,7 @@ func (c *Config) adjust() error {
 		if err != nil {
 			return terror.ErrMasterGetHostnameFail.Delegate(err)
 		}
-		c.Name = fmt.Sprintf("%s-%s", defaultNamePrefix, hostname)
+		c.Name = fmt.Sprintf("%s-%s-%s", defaultNamePrefix, hostname, port)
```
csuzhangxc: 👍

dm/worker/worker.go
```diff
@@ -68,6 +79,11 @@ type Worker struct {
 	tracer *tracing.Tracer
 
 	taskStatusChecker TaskStatusChecker
+
+	etcdClient *etcd.Client
+	election   *election.Election
+
+	name string
```
WangXiangUSTC: useless, I will remove this line next commit

PR #400, opened 2019-12-04
*: fix context usage for SQL operation (#377)

csuzhangxc:
cherry pick #377 manually before merged for testing.

---

when retrying SQL operation to downstream, `stop-task` / `pause-task` may not take effect because of the misuse of _context_.

fix context usage for SQL operation to downstream, including:
- data migration
- checkpoint update
- online DDL meta update

**NOTE**:

- shard DDL meta operation not updated in this PR because we are refactoring it
- better improvement for these operations can be done when moving them to etcd.
- `retry_cancel` integration testing has much wait time, and I'll split it run separately in CI after received 2 LGTM.

### Check List 

Tests 

 - Unit test
 - Integration test

Side effects

 - Increased code complexity

Related changes


csuzhangxc:
@WangXiangUSTC @GregoryIan PTAL

WangXiangUSTC:
LGTM

PR #403, opened 2019-12-06
*: fix DB connection close and all_mode integration test

csuzhangxc:

- `resume-task` may fail because `resetBaseConnFn` returning an invalid existing connection
- all_mode may fail with `/tmp/dm_test/all_mode/worker1/log/dm-worker.log dosen't contain dispatch auto resume task`.

- close the connection but do not put it back to the underlying connection pool
- wait for the task running before kill TiDB

### Check List 

Tests 

 - Integration test

Related changes


csuzhangxc:
@3pointer @amyangfei PTAL

PR #407, opened 2019-12-06
 Fix the missing var dm_worker_instances for rolling_update_monitor

SE-Bin:

Fix the missing var "dm_worker_instances "  for rolling_update_monitor

Add the var " dm_worker_instances " with the same logical as start.yml


csuzhangxc:
@amyangfei PTAL

Issue #410, opened 2019-12-07
DM Portal optimization


baurine:
# DM Portal optimization

## Abstract

According to issue #404 , the DM will optimize the configuration, so the DM Portal needs to be updated to support the new configuration, and we will also do some more optimizations to make it easier to use.

## Problem Statement
- It can't sync with the newly added tables and schemas
- The feature about editing task.yml is rarely to use

## Success Criteria
- Support the latest DM configuration
- Can sync with the newly added tables and schemas

## TODO list

- [ ] Move DM Portal to `pingcap/dm` repo
- [ ] Change the usage for `/generate_config` API to support sync with the newly added tables and schemas
- [ ] Remove the editing task.yml feature
- [ ] Support latest DM configuration

## Difficulty
Medium

## Score
1800

## Mentor(s)
@baurine 

## Recommended Skills
DM operation

## References
- [DM-portal source](https://github.com/pingcap/tidb-enterprise-tools/tree/dm-portal/dm-portal)
- [DM-portal user manual](https://docs.google.com/document/d/13ilFXJZErBXBRRREcskd5zhtocF-apLxFICp45OWbfs/edit)

GregoryIan:
a few months ago we implemented the first version of the portal, but unfortunately we haven't proven it useful, there is no conclusion whether the portal can solve user's problems. 

What do you think about this problem @baurine?
Maybe this special week is an opportunity to promote the portal,  you can discuss with @csuzhangxc ，and add a todo for it, like `promote and improve dm portal`

baurine:
hey @GregoryIan ,

@ericsyh as the PM of the first version of the portal, he has collected some feedback from clients - [DM Portal v0.1.0 feedback](https://docs.google.com/document/d/1tpMvyRHNYlSxH2IMFYq_eG_y9FjlUL0NAWoqnJCKvJY), so we decide to polish it according to their feedback in this special week and then to see the effect.

Of course, we will continue to listen to the clients and think about how to promote and improve it (added it as a TODO).

GregoryIan:
👍 @baurine

PR #417, opened 2019-12-10
Update DM Portal

baurine:

#410 

- Remove the editing task.yml feature
- Validate the task name, can't contain invalid characters
- Add option to support sync with the upstream newly added tables and schemas
- Differ the undo and reset actions more clearly

dm-portal-remove-edit

---

dm-portal-validate-task-name

---

dm-portal-auto-sync

---

dm-portal-undo
dm-portal-reset


baurine:
@ericsyh @WangXiangUSTC please help review again, thanks!

PR #416, opened 2019-12-10
*: split unit test according to upstream MySQL

csuzhangxc:
### What problem does this PR solve? 

- when multi test cases in different test suite write data in the upstream, the binlog may become out of order.
- run all unit test cases in one job in CI is a  little slow.

### What is changed and how it works?

- split unit tests according to pkg (in _Makefile_), and now split them into:
  - unit_test_relay
  - unit_test_syncer
  - unit_test_pkg_binlog
  - unit_test_others
- update jenkins pipeline

### Check List 

Tests 

 - Integration test

Related changes

 - Need to cherry-pick to the release branch

PR #418, opened 2019-12-11
*: cherry pick #378 and #416

csuzhangxc:

split unit test according to upstream MySQL

cherry pick #378 and #416 (both of them modify the same line in _Makefile_).

### Check List 

Tests 

 - Unit test
 - Integration test


csuzhangxc:
@WangXiangUSTC @kennytm PTAL

PR #419, opened 2019-12-11
dm-portal: don't generate table list for mydumper  if DoTables is empty

WangXiangUSTC:

fix a bug, when DoTables is empty will generate "-T " for Mydumper, which is invalid.

fix it

### Check List 

Tests 

 - Manual test
 - Unit test

Related changes


baurine:
LGTM

PR #423, opened 2019-12-12
pbmock: generate pb mock with reflect mode

amyangfei:
1. the pbmock generator took `dmmaster.pb.gw.go` as `dmmaster.pb.go` by mistaken
2. the `source mode` has the `imported package collision: "backoff" imported twice` problem

1. scan `*pb.go` files only
2. use `reflect mode` when generating mock files. ref: https://stackoverflow.com/a/52502366/1115857

### Check List 

Tests 

 - Unit test
 - Integration test

Related changes


csuzhangxc:
@WangXiangUSTC PTAL

PR #426, opened 2019-12-13
ha: support none-leader master server forward api request

glorv:
- support non-leader master forward api request to leader
- support dmctl config multi master address, if at least one address is currently valid, the request will be success 

- change dm-ctl to retry all the address to create connect
- change dm-master etcd leader key from `id` to json encoded sting `{"id": "id-value", address: "node-address"}`, so we can get the leader address from current leader key easily
- watch campaign leader events, after each leader change event, will update the gRPC connect to newest leader so we can redirect api request to the right address
- wrap master grpc API to add check and redirect to leader logic when current node is not master leader

### Check List 

Tests 

 - Unit test
 - Integration test 
   - add `forward_master` integration test, to test dm-ctl config multi-address along with non-master forward request

Code changes

Side effects
 - Increased code complexity for dm-master election

Related changes


dm/master/server.go
```diff
@@ -2027,3 +2037,128 @@ func (s *Server) workerArgsExtractor(args ...interface{}) (workerrpc.Client, str
...
+// redirect gRPC request to Leader if needed
+type redirectLeaderServer struct {
+	server pb.MasterServer
+	// return true if we should redirect to master leader
+	checkFunc func() bool
+}
+
+func (r *redirectLeaderServer) StartTask(ctx context.Context, req *pb.StartTaskRequest) (*pb.StartTaskResponse, error) {
+	if r.checkFunc() {
+		return common.MasterClient().StartTask(ctx, req)
```
csuzhangxc: We may know who is the leader before creating the client.

BTW, is it possible when _A_ is the leader and _B_, _C_ is the follower, then:
_dmctl --> B --> C --> B --> C --> ***_ and never forward the request to _A_.

glorv: the MasterClient() is connected to  current leader node, so it's true there maybe a change requests will be forwarded to outdated leader if several node's master connect is a circle (eg A->B, B->C, C->A), but this won't last long as within one lease, nodes should all get leadership change event thus will reset connect to right leader.

In Contrast, PD add a `redirect` header to http request if request should be redirected, and if the header already exist, it just return an error. To implement this need to change the gRPC request struct, so it's not easy to do so.

csuzhangxc: How do we ensure `the MasterClient() is connected to current leader node`?

tests/forward_master/data/db1.prepare.sql
```diff
@@ -0,0 +1,2 @@
+drop database if exists `forward_master`;
+
```
csuzhangxc: Do not create `forward_master` again?

tests/forward_master/run.sh
```diff
@@ -0,0 +1,60 @@
...
+    run_sql_file $cur/data/db2.prepare.sql $MYSQL_HOST2 $MYSQL_PORT2
+    check_contains 'Query OK, 3 rows affected'
+
+    run_dm_worker $WORK_DIR/worker1 $WORKER1_PORT $cur/conf/dm-worker1.toml
+    check_rpc_alive $cur/../bin/check_worker_online 127.0.0.1:$WORKER1_PORT
+    run_dm_worker $WORK_DIR/worker2 $WORKER2_PORT $cur/conf/dm-worker2.toml
+    check_rpc_alive $cur/../bin/check_worker_online 127.0.0.1:$WORKER2_PORT
+    run_dm_master $WORK_DIR/master1 $MASTER_PORT $cur/conf/dm-master1.toml
+
+    #check_rpc_alive $cur/../bin/check_master_online 127.0.0.1:$MASTER_PORT
```
csuzhangxc: Why do you comment on this line?

tests/forward_master/run.sh
```diff
@@ -0,0 +1,60 @@
+#!/bin/bash
```
csuzhangxc: In order to run the newly added case in CI, you should add it to _tests/other_integration.txt_.

tests/forward_master/run.sh
glorv: fixed, previous version is just for local test

dm/master/server.go
glorv: in election.campaign, after each leader change event, we will reconnect to new leader, so we can ensure we always connected to current leader

glorv:
@csuzhangxc  I push another reversion to fix the issue of recreate gRPC connection when mater leader changes and  add an integration test for forward master test. 
This is one problem remain: It seems currently, when master leader is down, one follower will win the leader campaign and become new leader after one session lease, but some running information is store in memory, so these information would lost when leader is down. So before master HA is implemented, the test will fill at [this line](https://github.com/pingcap/dm/blob/bb338c631f572ab37dc3bfc943b91e116258584c/tests/forward_master/run.sh#L44), should we wait for it, or just skip it currently

glorv:
@nolouch PTAL

csuzhangxc:
> @csuzhangxc I push another reversion to fix the issue of recreate gRPC connection when mater leader changes and add an integration test for forward master test.
> This is one problem remain: It seems currently, when master leader is down, one follower will win the leader campaign and become new leader after one session lease, but some running information is store in memory, so these information would lost when leader is down. So before master HA is implemented, the test will fill at [this line](https://github.com/pingcap/dm/blob/bb338c631f572ab37dc3bfc943b91e116258584c/tests/forward_master/run.sh#L44), should we wait for it, or just skip it currently

I this PR, maybe you can use `refresh-worker-tasks` to recovery the _worker-subtask_ mapping in the new leader before `stop-task`. But I think you can wait for HA ready.

dm/common/util.go
```diff
@@ -0,0 +1,44 @@
...
+		masterClient = nil
+		return errors.Trace(err)
+	}
+	masterClient = pb.NewMasterClient(conn)
+	return nil
+}
+
+// ResetMasterClient reset masterClient when no master is available
+func ResetMasterClient() {
+	masterClient = nil
```
csuzhangxc: Do we need to use any mutex to protect the access of `masterClient`?

dm/common/util.go
```diff
@@ -0,0 +1,44 @@
+package common
+
+import (
+	"github.com/pingcap/dm/dm/pb"
+	"github.com/pingcap/dm/pkg/log"
+	"github.com/pingcap/errors"
+	"go.uber.org/zap"
+	"google.golang.org/grpc"
+	"time"
```
csuzhangxc: Maybe you can:
1. add a license header for every newly added file
2. reorder the `import`s to 3 sections: builtin, 3th-party, and current project (the same for other updated files).

pkg/election/election.go
```diff
@@ -58,14 +64,29 @@ type Election struct {
...
 }
 
+// NodeMeta represent node name/address that stored in etcd master key
+type NodeMeta struct {
+	ID      string `json:"id"`
+	Address string `json:"address"`
+}
+
+func (m *NodeMeta) String() string {
+	return fmt.Sprintf("NodeMeat[id: %s, address: %s]", m.ID, m.Address)
```
csuzhangxc:
```suggestion
	return fmt.Sprintf("NodeMeta[id: %s, address: %s]", m.ID, m.Address)
```

pkg/election/election.go
```diff
@@ -80,6 +101,32 @@ func NewElection(ctx context.Context, cli *clientv3.Client, sessionTTL int, key,
...
 
+	for {
+		_, meta, err := e.LeaderInfo(ctx)
+		if err != nil {
+			return nil, errors.WithStack(err)
+		}
+		if meta == nil {
+			// fake set leader here
+			log.L().Warn("seems there is no leader now, try campaign leader.")
+			err := e.tryCampaignLeader(ctx, session.Lease())
```
csuzhangxc: `make check` got `declaration of "err" shadows declaration at line 105`.

pkg/election/election.go
```diff
@@ -80,6 +101,32 @@ func NewElection(ctx context.Context, cli *clientv3.Client, sessionTTL int, key,
 		return nil, terror.ErrElectionCampaignFail.Delegate(err, "create the initial session")
 	}
 
+	for {
+		_, meta, err := e.LeaderInfo(ctx)
+		if err != nil {
+			return nil, errors.WithStack(err)
```
csuzhangxc: We often use `errors.Trace(err)` and `errors.WithStack` is deprecated. or it's event better to add some terror in _pkg/terror/error_list.go_.

pkg/election/election.go
```diff
@@ -171,46 +223,94 @@ func (e *Election) campaignLoop(ctx context.Context, session *concurrency.Sessio
...
-
-		e.toBeLeader() // become the leader now
-		e.watchLeader(ctx, session, leaderKey)
-		e.retireLeader() // need to re-campaign
 	}
 }
 
+func (e *Election) tryCampaignLeader(ctx context.Context, leaseID clientv3.LeaseID) error {
+	value, _ := json.Marshal(&e.meta)
+	_, err := kv.NewSlowLogTxn(e.cli).
```
csuzhangxc: Is `SlowLog` needed here?

pkg/election/election.go
```diff
@@ -80,6 +101,32 @@ func NewElection(ctx context.Context, cli *clientv3.Client, sessionTTL int, key,
 		return nil, terror.ErrElectionCampaignFail.Delegate(err, "create the initial session")
 	}
 
+	for {
```
csuzhangxc: Block launching until one leader elected, right?

pkg/election/election.go
```diff
@@ -171,46 +223,94 @@ func (e *Election) campaignLoop(ctx context.Context, session *concurrency.Sessio
...
+	_, err := kv.NewSlowLogTxn(e.cli).
+		If(clientv3.Compare(clientv3.CreateRevision(e.key), "=", 0)).
+		Then(clientv3.OpPut(e.key, string(value), clientv3.WithLease(leaseID))).
+		Commit()
+	return errors.WithStack(err)
+}
+
 func (e *Election) toBeLeader() {
 	e.isLeader.Set(true)
+	e.leaderMeta.Store(e.meta)
```
csuzhangxc: Is it needed to combine `isLeader` into `leaderMeta`?

pkg/election/election.go
glorv: yes, because some other action in the init phase depend on the gRPC api such as `ShowDDLLocks`, so to make things easy, we try to determine a leader first. As it's embeded etcd, this would be quite fast

pkg/election/election.go
glorv: yes, I remove field isLeader and use leaderMeta to judge whether current node is leader so we can gain high consistency with only one atomic operation.

csuzhangxc:
the job have been done at https://github.com/pingcap/dm/pull/477.

PR #441, opened 2019-12-24
*: refine coordinator startup

nolouch:
Signed-off-by: nolouch <nolouch@gmail.com>

 coordinator startup:

- only leader startup 
- recover the registered workers and their binding relationship,

### Check List 

Tests 

 - Unit test

### DIFF
--- a/dm/master/coordinator/coordinator.go
+++ b/dm/master/coordinator/coordinator.go
@@ -1,3 +1,5 @@
+// Copyright 2019 PingCAP, Inc.
+//
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
--- a/dm/master/coordinator/coordinator.go
+++ b/dm/master/coordinator/coordinator.go
@@ -29,6 +29,7 @@ import (
 var (
 	etcdTimeouit = 3 * time.Second
 
+	// ErrNotStarted coordinator does not start.
 	ErrNotStarted = errors.New("coordinator does not start")
 )
 
@@ -80,7 +81,6 @@ func (c *Coordinator) Start(ctx context.Context, etcdClient *clientv3.Client) er
 	}
 
 	for _, kv := range resp.Kvs {
-		//addr := strings.TrimPrefix(common.WorkerRegisterKeyAdapter.Path(), string(kv.Key))
 		addr := common.WorkerRegisterKeyAdapter.Decode(string(kv.Key))[0]
 		name := string(kv.Value)
 		c.workers[addr] = NewWorker(name, addr)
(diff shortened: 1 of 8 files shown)


dm/master/election.go
```diff
@@ -30,13 +30,19 @@ func (s *Server) electionNotify(ctx context.Context) {
 			// output the leader info.
 			if leader {
 				log.L().Info("current member become the leader", zap.String("current member", s.cfg.Name))
+				err := s.coordinator.Start(ctx, s.etcdClient)
```
csuzhangxc: If before the old leader called `s.coordinator.Stop()`, the new leader called `s.coordinator.Start`, is it OK?

dm/worker/server.go
```diff
@@ -215,7 +228,17 @@ func (s *Server) StartSubTask(ctx context.Context, req *pb.StartSubTaskRequest)
...
+		ctx, cancel := context.WithTimeout(s.etcdClient.Ctx(), 3*time.Second)
+		defer cancel()
+		_, err = s.etcdClient.Put(ctx, common.UpstreamSubTaskKeyAdapter.Encode(s.cfg.WorkerAddr), cfg.String())
+		if err != nil {
+			resp.Result = false
+			resp.Msg = err.Error()
+		}
 	}
+	// FIXME: handle error
+	_ = w.OperateSubTask(cfg.Name, pb.TaskOp_Stop)
```
csuzhangxc: `TaskOp_Stop` 🤔

dm/master/election.go
nolouch: I think we can add a validator before RPC to handle that check if the dm-master is a leader. that may be related to the gRPC proxy.

csuzhangxc: OK

dm/master/coordinator/coordinator.go
```diff
@@ -55,9 +63,87 @@ func NewCoordinator() *Coordinator {
...
+
+	for _, kv := range resp.Kvs {
+		addr := common.UpstreamBoundWorkerKeyAdapter.Decode(string(kv.Key))[0]
+		sourceID := string(kv.Value)
+		w, ok := c.workers[addr]
+		if !ok {
+			log.L().Error("worker not exist but binding relationship exist", zap.String("addr", addr), zap.String("source", sourceID))
+			continue
+		}
+		resp, err = etcdClient.Get(ctx, common.UpstreamConfigKeyAdapter.Encode(sourceID))
```
csuzhangxc: Does this `resp` overwrite the value of it in L96?

nolouch: actually `for ... range` will clone the values, not affect the origin value.

PR #449, opened 2020-01-09
syncer: don't clear online ddl if execute ddl failed

WangXiangUSTC:

when execute ddl failed, ignore clean online ddl information

- check `execErrorDetected` after add ddl job, if `execErrorDetected` is true, return error for `handleQueryEvent`
- update sync unit's `IsFreshJob` by the way, if have shard table's checkpoint information, the unit should not be fresh job

### Check List 

Tests 

 - Integration test(will add it later)

Related changes


syncer/db.go
```diff
@@ -192,6 +193,15 @@ func (conn *DBConn) querySQL(tctx *tcontext.Context, query string, args ...inter
 }
 
 func (conn *DBConn) executeSQLWithIgnore(tctx *tcontext.Context, ignoreError func(error) bool, queries []string, args ...[]interface{}) (int, error) {
+
+	failpoint.Inject("ExecuteSQLWithIgnoreFailed", func(val failpoint.Value) {
```
WangXiangUSTC: will use this failpoint later

syncer/syncer.go
```diff
@@ -1054,9 +1055,9 @@ func (s *Syncer) Run(ctx context.Context) (err error) {
 			s.tctx.L().Error("panic log", zap.Reflect("error message", err1), zap.Stack("statck"))
 			err = terror.ErrSyncerUnitPanic.Generate(err1)
 		}
-		// flush the jobs channels, but if error occurred, we should not flush the checkpoints
-		if err1 := s.flushJobs(); err1 != nil {
```
WangXiangUSTC: seems that don't need to execute theses sqls, if sync unit exit because of error or user pause the task, will cancel the context, and all sqls will execute failed.

csuzhangxc:
@csuzhangxc @lichunzhu PTAL

syncer/checkpoint.go
```diff
@@ -473,6 +476,18 @@ func (cp *RemoteCheckPoint) GlobalPoint() mysql.Position {
 	return cp.globalPoint.MySQLPos()
 }
 
+// ShardTablePoint implements CheckPoint.ShardTablePoint
+func (cp *RemoteCheckPoint) ShardTablePoint() map[string]map[string]mysql.Position {
+	tablePoints := make(map[string]map[string]mysql.Position)
```
csuzhangxc: 1. Do we need to `RLock` here? Is `cp.points` possible to be changed in any other calls?
2. Is this `func` return all tables' checkpoint rather than just _shard_ tables?

syncer/db.go
```diff
@@ -192,6 +193,15 @@ func (conn *DBConn) querySQL(tctx *tcontext.Context, query string, args ...inter
 }
 
 func (conn *DBConn) executeSQLWithIgnore(tctx *tcontext.Context, ignoreError func(error) bool, queries []string, args ...[]interface{}) (int, error) {
+
```
csuzhangxc: How about removing this blank line?

syncer/syncer.go
```diff
@@ -1675,11 +1676,19 @@ func (s *Syncer) handleQueryEvent(ev *replication.QueryEvent, ec eventContext) e
...
 		job := newDDLJob(nil, needHandleDDLs, *ec.lastPos, *ec.currentPos, nil, nil, *ec.traceID)
 		err = s.addJobFunc(job)
 		if err != nil {
 			return err
 		}
+
+		// when add ddl job, will execute ddl and then flush checkpoint.
+		// if execute ddl failed, the execErrorDetected will be true.
+		if s.execErrorDetected.Get() {
+			return terror.ErrSyncerUnitHandleDDLFailed.Generate()
```
csuzhangxc: 1. How about including more information about this DDL job (at least the DDL statement)?
2. One more question is there any problem when resuming from an older checkpoint before any online DDL operation, like:
```
1. some online DDL operation
2. save it into TiDB (https://github.com/pingcap/dm/blob/8182696221688b61571bca4febd7d5df17194cb0/syncer/online_ddl.go#L205-L207)
3. error occurred
4. resume from any position older than step-1.
```
Is it possible `append`ed multiple times in step-2? 😂

syncer/syncer.go
csuzhangxc: Do we need to `s.jobWg.Wait()` all operation jobs to return?

syncer/syncer.go
WangXiangUSTC: 1. add ddl query in error in 92ede5b
2. judge ddl before save it in 3c602b0

syncer/db.go
WangXiangUSTC: done

syncer/checkpoint.go
WangXiangUSTC: done in 6252cf5

syncer/syncer.go
WangXiangUSTC: add in 20cfd53

csuzhangxc:
@WangXiangUSTC should we need to add the label of `should-cherry-pick-1.0`?

@lichunzhu PTAL again.

WangXiangUSTC:
> @WangXiangUSTC should we need to add the label of `should-cherry-pick-1.0`?
> 
> @lichunzhu PTAL again.

I add the tag

csuzhangxc:
@WangXiangUSTC Please cherry-pick this PR manually.

PR #450, opened 2020-01-11
HA: add `advertise-addr` for DM-worker

csuzhangxc:

DM-worker may listen on some special address (like `0.0.0.0:8262`) in a k8s pod, when registering it into the DM-master cluster, the DM-master may not access it with this special address.

- add `advertise-addr` config item for DM-worker to handle client traffic.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Related changes


csuzhangxc:

tests/_utils/run_dm_worker
```diff
@@ -25,7 +25,7 @@ PWD=$(pwd)
 echo "[$(date)] <<<<<< START DM-WORKER on port $port, config: $conf >>>>>>"
 cd $workdir
 $binary -test.coverprofile="$TEST_DIR/cov.$TEST_NAME.worker.$port.$(date +"%s").out" DEVEL \
-    --worker-addr=:$port \
+    --worker-addr=0.0.0.0:$port \
```
WangXiangUSTC: do we need to check the host? `0.0.0.0:$port` will become the `advertise-addr`, but it can not be used by dm-master

WangXiangUSTC:
LGTM

tests/_utils/run_dm_worker
csuzhangxc: 👍 addressed in https://github.com/pingcap/dm/pull/450/commits/2024c7c100401830597c2b2f847290fe4b3e73ac.
1. add the check for `0.0.0.0` as the `advertise-addr`
2. add `advertise-addr` when running dm-worker in tests.

lichunzhu:
There same to be some problem with `make check`.

csuzhangxc:
> There same to be some problem with `make check`.

Yes, the problem was introduced when merging changes from for other PR, and I've updated in https://github.com/pingcap/dm/pull/450/commits/d6f808436576aacc9d1b92c966d98e46a1735900.

lichunzhu:
LGTM

PR #451, opened 2020-01-11
ha: fix bug for get subtask from etcd && add integration test 

WangXiangUSTC:

there is a bug when get subtask

fix it, and add a simple integration test for it

### Check List 

Tests 

 - Integration test


WangXiangUSTC:
@lichunzhu @csuzhangxc PTAL

dm/worker/server.go
```diff
@@ -640,21 +640,23 @@ func (s *Server) startWorker(cfg *config.MysqlConfig) error {
...
+	resp, err := s.etcdClient.KV.Get(ectx, key, clientv3.WithPrefix())
+	if err != nil {
+		return err
+	}
+	for _, kv := range resp.Kvs {
+		infos := common.UpstreamSubTaskKeyAdapter.Decode(string(kv.Key))
+		taskName := infos[1]
+		task := string(kv.Value)
+		cfg := config.NewSubTaskConfig()
+		if err := cfg.Decode(task); err != nil {
```
csuzhangxc:
```suggestion
		if err = cfg.Decode(task); err != nil {
```

the same with L656.

dm/worker/server.go
```diff
@@ -640,21 +640,23 @@ func (s *Server) startWorker(cfg *config.MysqlConfig) error {
...
+			return nil
```
csuzhangxc: Do we need to return `err` here? the same with L657.

dm/worker/server.go
```diff
@@ -640,21 +640,23 @@ func (s *Server) startWorker(cfg *config.MysqlConfig) error {
...
+		}
+		if err := w.StartSubTask(cfg); err != nil {
+			return nil
 		}
+		log.L().Info("load subtask successful", zap.String("sourceID", cfg.SourceID), zap.String("name", taskName))
```
csuzhangxc:
```suggestion
		log.L().Info("load subtask successful", zap.String("sourceID", cfg.SourceID), zap.String("task", taskName))
```

Is `task` more meaningful here? just as we did in _syncer.go_. `tcontext.Background().WithLogger(log.With(zap.String("task", cfg.Name), zap.String("unit", "binlog replication")))`.

tests/ha/conf/dm-master.toml
```diff
@@ -0,0 +1,9 @@
+# Master Configuration.
+
+[[deploy]]
+source-id = "mysql-replica-01"
+dm-worker = "127.0.0.1:8262"
+
+[[deploy]]
+source-id = "mysql-replica-02"
+dm-worker = "127.0.0.1:8263"
```
csuzhangxc: Can we remove these now? (and maybe we need to remove them for other cases.)

tests/ha/run.sh
```diff
@@ -0,0 +1,67 @@
+#!/bin/bash
```
csuzhangxc: Do you forget to add this case in _tests/others_integration.txt_?

dm/worker/server.go
csuzhangxc: the same with L656

WangXiangUSTC: done

dm/worker/server.go
WangXiangUSTC: yes, need to return error, fixed

tests/ha/conf/dm-master.toml
WangXiangUSTC: all removed

tests/ha/run.sh
WangXiangUSTC: yes, I add it in 34cd486

tests/ha/run.sh
```diff
@@ -0,0 +1,67 @@
...
+    echo "start DM task"
+    dmctl_start_task
+
+    echo "use sync_diff_inspector to check full dump loader"
+    check_sync_diff $WORK_DIR $cur/conf/diff_config.toml
+
+    echo "start dm-worker3 and kill dm-worker2"
+    run_dm_worker $WORK_DIR/worker3 $WORKER3_PORT $cur/conf/dm-worker3.toml
+    check_rpc_alive $cur/../bin/check_worker_online 127.0.0.1:$WORKER3_PORT
+    ps aux | grep dm-worker2 |awk '{print $2}'|xargs kill || true
```
csuzhangxc: Do we need to wait for dm-worker2 to become offline?
I find this in the log of CI: `kill: sending signal to 466 failed: No such process`.

WangXiangUSTC: 1. I add it in 3721168
2.it is the process of grep

lichunzhu:
Can this commit pass HA integration_test?
You can add
```go
	common.PrettyPrintResponse(resp)
```
in [dm/ctl/master/operate_mysql_worker.go](https://github.com/pingcap/dm/blob/ha-dev/dm/ctl/master/operate_mysql_worker.go#L73) to try it

PR #457, opened 2020-01-15
relay: start with the min binlog position in checkpoint

WangXiangUSTC:

relay use the latest binlog position as default, if increment task migrate from `dm-worker1` to `dm-worker2`, sync unit will get binlog from checkpoint's binlog position, and this position  <= the latest position, so sync unit can't get the binlog.

get the min binlog position from all subtasks, and relay start with this min position.

### Check List 

Tests 

 - Integration test


dm/master/coordinator/coordinator.go
```diff
@@ -418,7 +425,7 @@ func (c *Coordinator) restartMysqlTask(w *Worker, cfg *config.MysqlConfig) bool
 		Op:     pb.WorkerOp_StartWorker,
 		Config: task,
 	}
-	resp, err := w.OperateMysqlWorker(context.Background(), req, time.Second*3)
+	resp, err := w.OperateMysqlWorker(context.Background(), req, time.Second*5)
```
WangXiangUSTC: just add duration to let test pass, maybe @lichunzhu will fix it

pkg/binlog/position.go
```diff
@@ -132,3 +132,15 @@ func verifyUUIDSuffix(suffix string) bool {
 	}
 	return true
 }
+
+// AdjustPosition adjusts the filename with uuid suffix in mysql position
+// for example: mysql-bin|000001.000002 -> mysql-bin.000002
+func AdjustPosition(pos gmysql.Position) gmysql.Position {
```
WangXiangUSTC: copy from master branch

WangXiangUSTC:
this pr contain https://github.com/pingcap/dm/pull/451, we can merge 451 at the first, or review this pr directly and close 451 @lichunzhu @csuzhangxc

lichunzhu:
> this pr contain #451, we can merge 451 at the first, or review this pr directly and close 451 @lichunzhu @csuzhangxc

@WangXiangUSTC I have already approved #451. If this PR runs fine in your machine, I think it's OK to merge it directly.

dm/worker/server.go
```diff
@@ -749,3 +780,85 @@ func makeCommonWorkerResponse(reqErr error) *pb.CommonWorkerResponse {
...
+			if minPos.Compare(*pos) >= 1 {
+				minPos = pos
+			}
+		}
+	}
+
+	return minPos, nil
+}
+
+func getMinPosForSubTask(subTaskCfg *config.SubTaskConfig) (minPos *mysql.Position, err error) {
```
csuzhangxc: This function depends on the internal implementation of Checkpoint. How about adding such a method in Checkpoint?

dm/worker/server.go
```diff
@@ -749,3 +780,85 @@ func makeCommonWorkerResponse(reqErr error) *pb.CommonWorkerResponse {
...
+	syncCheckpointTable := fmt.Sprintf("%s_syncer_checkpoint", subTaskCfg.Name)
+	exist, err := utils.TableExists(db.DB, subTaskCfg.MetaSchema, syncCheckpointTable)
+	if err != nil {
+		return nil, err
+	}
+	if !exist {
+		return nil, nil
+	}
+
+	query := fmt.Sprintf("SELECT `binlog_name`, `binlog_pos` FROM %s.%s where id = ?", subTaskCfg.MetaSchema, syncCheckpointTable)
```
csuzhangxc: Is it enough to just get the _global_ checkpoint?

dm/worker/server.go
```diff
@@ -674,21 +676,50 @@ func (s *Server) startWorker(cfg *config.MysqlConfig) error {
...
+		if err = subTaskcfg.Decode(task); err != nil {
+			return err
+		}
+		subTaskcfg.LogLevel = s.cfg.LogLevel
+		subTaskcfg.LogFile = s.cfg.LogFile
+
+		subTaskCfgs = append(subTaskCfgs, subTaskcfg)
+	}
+
+	if len(cfg.RelayBinLogName) == 0 && len(cfg.RelayBinlogGTID) == 0 {
```
csuzhangxc: If the user specify `relay-binlog-name` explicitly in the config, then run this relay task in DM-worker-A for a long time and scheduled to DM-worker-B, this `relay-binlog-name` may not be exist in the upstream MySQL now. If we use it to start the DM-worker-B, it will fail to fetch binlog, right? Should we need to handle this case?

dm/worker/server.go
```diff
@@ -674,21 +676,50 @@ func (s *Server) startWorker(cfg *config.MysqlConfig) error {
...
+		minPos, err := getMinPosInAllSubTasks(subTaskCfgs)
+		if err != nil {
 			return err
 		}
-		cfg.LogLevel = s.cfg.LogLevel
-		cfg.LogFile = s.cfg.LogFile
+		if minPos != nil {
```
csuzhangxc: We do not support GTID now, add some TODO here?

dm/master/coordinator/coordinator.go
```diff
@@ -392,18 +392,26 @@ func (c *Coordinator) tryRestartMysqlTask() {
 	for hasTaskToSchedule {
 		select {
 		case source := <-c.waitingTask:
+			log.L().Info("try restart mysql task", zap.String("source", source))
```
lichunzhu: Actually the log **try restart mysql task** info is not accurate here. The tasks started for the first time are also scheduled in this function. Check `c.waitingTask` in goland you will find the only read access is here.

WangXiangUSTC: I update the log 70d6e8e

dm/worker/server.go
WangXiangUSTC: add in 736aa48

dm/worker/server.go
WangXiangUSTC: update in 79442de

dm/worker/server.go
WangXiangUSTC: yes, I use checkpoint.GlobalPoint in 69363d7

dm/worker/server.go
WangXiangUSTC: update in 69363d7

dm/worker/server.go
```diff
@@ -668,7 +667,7 @@ func (s *Server) startWorker(cfg *config.MysqlConfig) error {
 
 	subTaskCfgs := make([]*config.SubTaskConfig, 0, 3)
 
-	ectx, cancel := context.WithTimeout(s.etcdClient.Ctx(), time.Second*3)
+	ectx, cancel := context.WithTimeout(s.etcdClient.Ctx(), time.Second*5)
```
csuzhangxc: How about giving it another name because it's also used for database operation now (no only for **e**tcd)?

dm/worker/server.go
```diff
@@ -808,59 +807,24 @@ func getMinPosInAllSubTasks(subTaskCfgs []*config.SubTaskConfig) (minPos *mysql.
...
+func getMinPosForSubTask(ctx context.Context, subTaskCfg *config.SubTaskConfig) (minPos *mysql.Position, err error) {
+	if subTaskCfg.Mode == config.ModeFull {
+		return nil, nil
 	}
-	defer db.Close()
 
-	// only read the sync unit's checkpoint information
-	syncCheckpointTable := fmt.Sprintf("%s_syncer_checkpoint", subTaskCfg.Name)
-	exist, err := utils.TableExists(db.DB, subTaskCfg.MetaSchema, syncCheckpointTable)
+	tctx := tcontext.Background().WithLogger(log.With())
```
csuzhangxc: The passed in `ctx` is not used.

dm/worker/server.go
WangXiangUSTC: I create a new ctx for getting min position

dm/worker/server.go
WangXiangUSTC: update in ce3635a

csuzhangxc:
@lichunzhu PTAL again.

lichunzhu:
The unit test didn't pass. Could you pls check it?

WangXiangUSTC:
@lichunzhu I fix unit test in 7802a15 and 1f82f5a, PTAL again

PR #462, opened 2020-01-19
ha: handle invalid operate type on worker

WangXiangUSTC:

if operate type is not `create` or `update` will use `stop` as default

only use `stop` operate when op is `stop`, and handle invalid operate type 

### Check List 

Tests 

 - Integration test


dm/master/server.go
```diff
@@ -1747,7 +1747,7 @@ func (s *Server) OperateMysqlWorker(ctx context.Context, req *pb.MysqlWorkerRequ
 		if resp, err = w.OperateMysqlWorker(ctx, req, s.cfg.RPCTimeout); err != nil {
 			return makeMysqlWorkerResponse(err)
 		}
-	} else {
+	} else if req.Op == pb.WorkerOp_StopWorker {
```
csuzhangxc: How about unifying to use `switch .. case`?

dm/master/server.go
```diff
@@ -1764,7 +1764,13 @@ func (s *Server) OperateMysqlWorker(ctx context.Context, req *pb.MysqlWorkerRequ
...
 			s.coordinator.HandleStoppedWorker(w, cfg)
 		}
+	} else {
+		return &pb.MysqlWorkerResponse{
+			Result: false,
+			Msg:    "invalid operate on worker",
+		}, nil
 	}
+
 	return &pb.MysqlWorkerResponse{
```
lichunzhu: Are these lines same to `return resp`?

dm/master/server.go
WangXiangUSTC: update in 0b80cf4

dm/master/server.go
WangXiangUSTC: yes, update in 0b80cf4

PR #464, opened 2020-02-03
dm-ha/: refine return resp instead of error, refine dm-ha logic

lichunzhu:
1. Some requests are not returning error through resp.
2. For some workers' resp, write etcd fail will not affect `resp.Result`.
3. When restartMysqlWorkerFail coordinator will refer to `WorkerRegisterKeyAdapter` which should be `WorkerKeepAliveKeyAdapter`.
4. `coordinator.start` is set to true before `ObserveWorkers`.

1. Return non-communicational error through resp.
2. Change `resp.Result` according to write etcd result.
3. Change to `WorkerKeepAliveKeyAdapter`.
4. Set `coordinator.start` to true before `ObserveWorkers`.
### Check List 

Tests 

 - Unit test
 - Integration test


dm/master/server.go
```diff
@@ -290,11 +290,17 @@ func (s *Server) RegisterWorker(ctx context.Context, req *pb.RegisterWorkerReque
...
+			Result: false,
+			Msg:    errors.ErrorStack(err),
+		}, nil
 	}
 	if !resp.Succeeded {
 		if len(resp.Responses) == 0 {
-			return nil, errors.Errorf("the response kv is invalid length, request key: %s", k)
+			return &pb.RegisterWorkerResponse{
+				Result: false,
+				Msg:    fmt.Sprintf("the response kv is invalid length, request key: %s\n", k),
```
csuzhangxc: Is `\n` needed?

dm/master/server.go
```diff
@@ -353,7 +359,11 @@ func (s *Server) OfflineWorker(ctx context.Context, req *pb.OfflineWorkerRequest
 		Then(clientv3.OpDelete(k)).
 		Commit()
 	if err != nil {
-		return nil, err
+		err = errors.Annotatef(err, "Delete key %s from etcd failed", req.Address)
```
csuzhangxc:
```suggestion
		err = errors.Annotatef(err, "delete key %s from etcd failed", req.Address)
```

dm/worker/server.go
```diff
@@ -706,6 +708,12 @@ func (s *Server) startWorker(cfg *config.MysqlConfig) error {
 		return err
 	}
 	s.setWorker(w, false)
+	defer func() {
+		// close w when some error occurs
+		if err1 != nil {
+			w.Close()
```
csuzhangxc: `Close` worker when starting any subtask failed?

lichunzhu: If we start subtask failed, shouldn't we free this worker for other sources?

dm/worker/server.go
```diff
@@ -289,6 +289,7 @@ func (s *Server) StartSubTask(ctx context.Context, req *pb.StartSubTaskRequest)
 	if resp.Result {
 		op1 := clientv3.OpPut(common.UpstreamSubTaskKeyAdapter.Encode(cfg.SourceID, cfg.Name), req.Task)
 		resp.Msg = s.retryWriteEctd(op1)
```
WangXiangUSTC: this function return failed msg when write etcd failed? seems strange, can we refine it

csuzhangxc: How about returning `error` directly?

WangXiangUSTC: I think it's ok

lichunzhu: Revised in https://github.com/pingcap/dm/pull/464/commits/63859013fce5644898bab2991f9d7980020c785b.
For operations related to tasks there is no side effect. But for `OperateMysqlWorker`, when worker has been started yet but write etcd failed, `OperateMysqlWorker` will set this worker to free but actually this worker can't handle other source any more. Same problem like communication error.
This problem should be resolved by refining the design of etcd keys.

dm/worker/server.go
```diff
@@ -321,8 +323,10 @@ func (s *Server) OperateSubTask(ctx context.Context, req *pb.OperateSubTaskReque
 		// clean subtask config when we stop the subtask
 		if req.Op == pb.TaskOp_Stop {
 			op1 := clientv3.OpDelete(common.UpstreamSubTaskKeyAdapter.Encode(w.cfg.SourceID, req.Name))
-			resp.Msg = s.retryWriteEctd(op1)
-			resp.Result = len(resp.Msg) == 0
+			err = s.retryWriteEctd(op1)
+			if err != nil {
+				return nil, err
+			}
```
WangXiangUSTC:
```
err = s.retryWriteEctd(op1)
if err != nil {
    resp.Msg = err.Error()
    resp.Result = false
} else {
    resp.Result = true
}
```
how about update to these code?

WangXiangUSTC:
LGTM

dm/worker/server.go
```diff
@@ -288,7 +288,10 @@ func (s *Server) StartSubTask(ctx context.Context, req *pb.StartSubTaskRequest)
 
 	if resp.Result {
 		op1 := clientv3.OpPut(common.UpstreamSubTaskKeyAdapter.Encode(cfg.SourceID, cfg.Name), req.Task)
-		resp.Msg = s.retryWriteEctd(op1)
+		err = s.retryWriteEctd(op1)
+		if err != nil {
+			return nil, err
```
csuzhangxc: Do not return a `resp` with `err` here? The same with L328 and L778.

dm/master/coordinator/coordinator.go
```diff
@@ -492,9 +498,9 @@ func (c *Coordinator) restartMysqlTask(w *Worker, cfg *config.MysqlConfig) bool
 	if w.State() == WorkerClosed {
 		ectx, cancel := context.WithTimeout(c.etcdCli.Ctx(), etcdTimeout)
 		defer cancel()
-		resp, err := c.etcdCli.Get(ectx, common.WorkerRegisterKeyAdapter.Encode(w.Address(), w.Name()))
+		resp, err := c.etcdCli.Get(ectx, common.WorkerKeepAliveKeyAdapter.Encode(w.Address(), w.Name()))
```
csuzhangxc: You can continue to think about the more reasonable design of these keys (do not forget the `TODO` in L374).

dm/master/coordinator/coordinator.go
```diff
@@ -492,9 +498,9 @@ func (c *Coordinator) restartMysqlTask(w *Worker, cfg *config.MysqlConfig) bool
...
 		if err != nil {
-			if resp.Count > 0 {
+			if resp != nil && resp.Count > 0 {
```
csuzhangxc: `if err != nil`, can `resp != nil` be true?

dm/worker/server.go
lichunzhu: If gRPC service returns an `err` here, dm-master will not check about the resp.

csuzhangxc: So why don't we unify the return values ​​here?

dm/master/coordinator/coordinator.go
lichunzhu: emmmm. I think it should be
`if err != nil ` -> `if err == nil`

dm/worker/server.go
lichunzhu: `return resp, err`?

csuzhangxc: Can not `return resp, nil`?

dm/master/coordinator/coordinator.go
```diff
@@ -499,7 +499,7 @@ func (c *Coordinator) restartMysqlTask(w *Worker, cfg *config.MysqlConfig) bool
 		ectx, cancel := context.WithTimeout(c.etcdCli.Ctx(), etcdTimeout)
 		defer cancel()
 		resp, err := c.etcdCli.Get(ectx, common.WorkerKeepAliveKeyAdapter.Encode(w.Address(), w.Name()))
-		if err != nil {
+		if err == nil {
```
csuzhangxc: How do we handle `err != nil`?

lichunzhu: `err != nil` means because of some reasons we can't get the keepAlive result for this worker. I think add retry may refine this problem, but it can't solve this problem. It's hard to solve this problem in current mechanism.

csuzhangxc: So, how about just logging it for debugging now?

lichunzhu: OK, I will add a log.

lichunzhu:
New scheduler is used in DM-HA. This PR is useless now.

PR #469, opened 2020-02-07
ha: save HandleSQLs request to etcd

WangXiangUSTC:
HandleSQLs request is not saved to etcd now, if dm-worker is offline and task is scheduled to another dm-worker, this request will be inoperative.

save HandleSQLs to etcd, and load request from etcd before subtask run. 

### Check List 

Tests 

 - Integration test


WangXiangUSTC:
@csuzhangxc @lichunzhu PTAL

PR #471, opened 2020-02-10
.*: support HA

WangXiangUSTC:

support ha

merge code from ha-dev, and refine some import's format

### Check List 

Side effects

Related changes


WangXiangUSTC:
@lichunzhu @csuzhangxc PTAL

dm/master/config.go
```diff
@@ -50,13 +50,15 @@ var SampleConfigFile string
 // NewConfig creates a config for dm-master
 func NewConfig() *Config {
 	cfg := &Config{}
+	cfg.Debug = false
```
lichunzhu: It sames that `cfg.Debug` is useless now. We can delete it in next PR.

csuzhangxc: some more useless code needs to be deleted later.

PR #470, opened 2020-02-08
syncer: fix lock in streamer controller

WangXiangUSTC:
### What problem does this PR solve? 
1. `GetEvent` may cost too many time, and will lock the `StreamerController`, need refine it
2. both `Start` and `Close` have lock, but when `Start` failed will call the `Close` function, need fix the dead lock

### What is changed and how it works?

fix the lock

### Check List 

Tests 
 - Manual test 

Related changes

 - Need to cherry-pick to the release branch

PR #472, opened 2020-02-10
syncer: fix lock in streamer controller(#470)

WangXiangUSTC:

1. GetEvent may cost too many time, and will lock the StreamerController, need refine it
2. both Start and Close have lock, but when Start failed will call the Close function, need fix the dead lock

cherry-pick from #470 and resolve conflicts

### Check List 

Tests 

 - Manual test


WangXiangUSTC:
@lichunzhu @csuzhangxc PTAL

PR #473, opened 2020-02-10
ha: refactor the schedule model

csuzhangxc:

the HA model before this PR can't tolerate well, and can't support pause-relay/pause-task correctly.

rewrite the HA mode, see _pkg/ha/doc.go_.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes


pkg/ha/doc.go
```diff
@@ -0,0 +1,146 @@
...
+// - the running stage:
+//   - NOTE: persist the current stage of the relay and subtask if needed later.
+//   - the bound relationship between the DM-worker instance and the upstream MySQL source (including relevant relay and subtasks):
+//     - PUT when scheduling the source to a DM-worker instance by DM-master.
+//       - PUT with `the expectant stage of the relay` in one txn.
+//     - GET (through WATCH) by DM-worker to know relevant relay/subtasks have to do.
+//     - DELETE when removing an upstream by DM-master.
+//       - DELETE with `the upstream MySQL config` in one txn.
+//       - DELETE with `the expectant stage of the relay` in one txn.
+//     - UPDATE when scheduling the source to another DM-worker instance by DM-master.
```
lichunzhu: How to `UPDATE`?

lichunzhu: Besides, what's the form of the key of the running stage? `/sourceID/workerAddr` or `/workerAddr/sourceID`?

csuzhangxc: PUT the new value to overwrite the old one.
May be `/sourceID` (key) -> `worker-name` (value), I'll add that later.

pkg/ha/doc.go
```diff
@@ -0,0 +1,146 @@
...
+// 19. the relay of the DM-worker continues to run.
+// 20. the user remove the upstream config.
+// 21. DM-master DELETE the upstream MySQL config, the bound relationship and the expectant stage of the relay.
+// 22. DM-worker shutdown.
+// 23. the user marks the DM-worker as offline.
+//   - DELETE DM-worker instance information in etcd.
+//
+// when the DM-worker (with relay and subtasks) is down:
+// 0. the status of the old DM-worker is un-health (keep-alive failed).
+// 1. DM-master choose another DM-worker instance for failover.
```
WangXiangUSTC: maybe DM-master need to wait a while(two lease is enough? the old dm-worker need stop all subtask) and then schedule

csuzhangxc: 👍 you are right, I added a note in https://github.com/pingcap/dm/pull/473/commits/ca3191dd5c64461bce0c4c7c97de4869a151dc89.

We need to think about it before we know how long it is enough.

pkg/ha/bound.go
```diff
@@ -0,0 +1,176 @@
...
+	if err != nil {
+		return bound, 0, err
+	}
+
+	return bound, resp.Header.Revision, nil
+}
+
+// WatchSourceBound watches PUT & DELETE operations for the bound relationship of the specified DM-worker.
+// For the DELETE operations, it returns an empty bound relationship.
+func WatchSourceBound(ctx context.Context, cli *clientv3.Client,
```
lichunzhu: Shall we return error for retry?

pkg/ha/source.go
```diff
@@ -0,0 +1,94 @@
...
+	if err != nil {
+		return nil, 0, err
+	}
+
+	scm := make(map[string]config.MysqlConfig)
+	for _, kv := range resp.Kvs {
+		var cfg config.MysqlConfig
+		err = cfg.Parse(string(kv.Value))
+		if err != nil {
+			return nil, 0, err
```
WangXiangUSTC: how about add the key's information in error?

pkg/ha/stage.go
```diff
@@ -0,0 +1,364 @@
...
+		return stage, 0, nil
+	} else if resp.Count > 1 {
+		// TODO(csuzhangxc): add terror.
+		// this should not happen.
+		return stage, 0, fmt.Errorf("too many relay stage (%d) exist for source %s", resp.Count, source)
+	}
+
+	stage, err = stageFromJSON(string(resp.Kvs[0].Value))
+	if err != nil {
+		return stage, 0, err
```
WangXiangUSTC: ditto

pkg/ha/stage.go
```diff
@@ -0,0 +1,364 @@
...
+}
+
+// GetAllSubTaskStage gets all subtask stages.
+// k/v: source ID -> task name -> subtask stage.
+func GetAllSubTaskStage(cli *clientv3.Client) (map[string]map[string]Stage, int64, error) {
+	ctx, cancel := context.WithTimeout(cli.Ctx(), etcdutil.DefaultRequestTimeout)
+	defer cancel()
+
+	resp, err := cli.Get(ctx, common.StageSubTaskKeyAdapter.Path(), clientv3.WithPrefix())
+
```
WangXiangUSTC: can delete this line

pkg/ha/stage.go
```diff
@@ -0,0 +1,364 @@
...
+	return ops, nil
+}
+
+// deleteRelayStageOp returns a DELETE etcd operation for the relay stage.
+func deleteRelayStageOp(source string) clientv3.Op {
+	return clientv3.OpDelete(common.StageRelayKeyAdapter.Encode(source))
+}
+
+// deleteSubTaskStageOp returns a list of DELETE etcd operation for the subtask stage.
+func deleteSubTaskStageOp(stages ...Stage) []clientv3.Op {
```
WangXiangUSTC: these subtasks have same task name or same source?

pkg/ha/source.go
```diff
@@ -0,0 +1,94 @@
...
+		return 0, err
+	}
+	key := common.UpstreamConfigKeyAdapter.Encode(cfg.SourceID)
+
+	return etcdutil.DoOpsInOneTxn(cli, clientv3.OpPut(key, value))
+}
+
+// GetSourceCfg gets the config of the specified source.
+// if the config for the source not exist, return with `err == nil` and `revision=0`.
+func GetSourceCfg(cli *clientv3.Client, source string, rev int64) (config.MysqlConfig, int64, error) {
```
WangXiangUSTC: why need revision in the argument?

lichunzhu: After worker is started, worker will get and watch source bound stage which will returns a revision. We use revision to get relative source config for a source bound event.
For example:
the time sequence is:
1. Put sourceCfg
2. Put sourceBound
3. Delete sourceBound
4. Delete sourceCfg
5. Put sourceCfg2(but has same sourceID with sourceCfg)

If when worker get newest sourceCfg at step.2 etcd has done step1~5, worker will start source with the config of sourceCfg2 which is not correct.

pkg/ha/stage.go
csuzhangxc: may have the same task name, but often have different source IDs (like stoping a task with multi subtasks with multi sources).

pkg/ha/source.go
csuzhangxc: YES, we need it. I add a TODO in https://github.com/pingcap/dm/pull/473/commits/73cee97e06a098dd821d97b742e3b0556bac8f19, and we need to design the terror code for these etcd operations later.

pkg/ha/stage.go
csuzhangxc: YES, but will do that later.

pkg/ha/stage.go
csuzhangxc: deleted in https://github.com/pingcap/dm/pull/473/commits/73cee97e06a098dd821d97b742e3b0556bac8f19.

dm/master/scheduler/scheduler.go
```diff
@@ -0,0 +1,1139 @@
...
+		return
+	}
+
+	s.logger.Info("the scheduler is closing")
+	if s.cancel != nil {
+		s.cancel()
+		s.cancel = nil
+	}
+
+	s.wg.Wait()
```
WangXiangUSTC: some function needs lock in `handleWorkerEv`, this may cause dead lock.

dm/master/scheduler/scheduler.go
```diff
@@ -0,0 +1,1139 @@
...
+		workerName string // empty should be fine below.
+		worker     *Worker
+	)
+	if w, ok := s.bounds[source]; ok {
+		worker = w
+		workerName = w.BaseInfo().Name
+	}
+
+	// 4. delete the info in etcd.
+	_, err := ha.DeleteSourceCfgRelayStageSourceBound(s.etcdCli, source, workerName)
```
lichunzhu: what if source is unbounded? which means workerName=""

dm/master/scheduler/scheduler.go
csuzhangxc: this `s.wg.Wait()` should wait `handleWorkerEv` return.
```go
go func() {
	defer s.wg.Done()
	s.handleWorkerEv(ctx, workerEvCh, workerErrCh)
}()
```

dm/master/scheduler/scheduler.go
csuzhangxc: `clientv3.OpDelete(common.UpstreamBoundWorkerKeyAdapter.Encode(worker))` works fine for `workerName=""` because no `clientv3.WithPrefix()` used.

And, I have already added a comment above
```go
workerName string // empty should be fine below.
```

lichunzhu: sorry, I haven't noticed that before

dm/master/scheduler/scheduler.go
WangXiangUSTC: for example:
1. Close  -> lock
2. execute `handleWorkerOffline in` `handleWorkerEv` -> lock, this will wait lock
3. s.wg.wait, wait `handleWorkerEv` finish, this will cause dead lock

csuzhangxc: Good catchup, fixed in https://github.com/pingcap/dm/pull/473/commits/bdbe56f3e6640aaa9d066ab48cd666a81c8399a1.

dm/worker/worker.go
```diff
@@ -306,6 +304,110 @@ func (w *Worker) QueryError(name string) []*pb.SubTaskError {
...
+// purgeRelayDir will clear all contents under w.cfg.RelayDir
+func (w *Worker) purgeRelayDir() error {
+	if !w.cfg.EnableRelay {
+		return nil
+	}
+	dir := w.cfg.RelayDir
+	d, err := os.Open(dir)
+	// fail to open dir, return directly
+	if err != nil {
+		return nil
```
WangXiangUSTC: only when err is `no such file or directory` can return nil?

dm/master/scheduler/scheduler.go
```diff
@@ -0,0 +1,1168 @@
...
+		if stage, ok := stagesM[source]; !ok {
+			notExistSourcesM[source] = struct{}{}
+		} else {
+			stages = append(stages, stage)
+		}
+		if cfg, ok := cfgsM[source]; ok {
+			cfgs = append(cfgs, cfg)
+		}
+	}
+	notExistSources := strMapToSlice(notExistSourcesM)
```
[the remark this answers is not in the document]
csuzhangxc: used in `terror.ErrSchedulerSubTaskOpSourceNotExist.Generate(notExistSources)` two.

dm/worker/config.go
```diff
@@ -54,6 +55,7 @@ func NewConfig() *Config {
 	// NOTE: add `advertise-addr` for dm-master if needed.
 	fs.StringVar(&cfg.Join, "join", "", `join to an existing cluster (usage: dm-master cluster's "${master-addr}")`)
 	fs.StringVar(&cfg.Name, "name", "", "human-readable name for DM-worker member")
+	fs.Int64Var(&cfg.KeepAliveTTL, "keepalive-ttl", defaultKeepAliveTTL, "dm-worker's TTL for keepalive with etcd (in seconds)")
```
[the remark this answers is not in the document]
csuzhangxc: good suggestion, maybe we can use a default at currently.

dm/worker/worker.go
csuzhangxc: @lichunzhu PTAL

lichunzhu: Revised in https://github.com/pingcap/dm/pull/473/commits/2783ba63bdef5122af8ba5ff5d8d46afc66e0a46

dm/worker/config.go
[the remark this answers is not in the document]
WangXiangUSTC: I think it's ok

dm/master/server.go
```diff
@@ -1348,92 +1104,62 @@ func (s *Server) CheckTask(ctx context.Context, req *pb.CheckTaskRequest) (*pb.C
...
+		return err
 	}
-	if err = cfg.Adjust(fromDB.DB); err != nil {
-		return makeMysqlWorkerResponse(err)
+	return nil
+}
+
+// OperateSource will create or update an upstream source.
+func (s *Server) OperateSource(ctx context.Context, req *pb.OperateSourceRequest) (*pb.OperateSourceResponse, error) {
+	log.L().Info("", zap.Stringer("payload", req), zap.String("request", "OperateSource"))
```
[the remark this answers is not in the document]
csuzhangxc: YES

PR #474, opened 2020-02-11
dm-syncer: minor fix on password && print parse error

WangXiangUSTC:

1. don't decrypt password, will connect to database failed
2.  log is not initial, need to use `fmt` to print parse error

fix these two problems

### Check List 

Tests 

 - Manual test

### DIFF
--- a/cmd/dm-syncer/main.go
+++ b/cmd/dm-syncer/main.go
@@ -36,13 +36,15 @@ func main() {
 	// 1. init conf
 	commonConfig := newCommonConfig()
 	conf, err := commonConfig.parse(os.Args[1:])
-	switch errors.Cause(err) {
-	case nil:
-	case flag.ErrHelp:
-		os.Exit(0)
-	default:
-		log.L().Error("parse cmd flags err " + err.Error())
-		os.Exit(2)
+	if err != nil {
+		switch errors.Cause(err) {
+		case nil:
+		case flag.ErrHelp:
+			os.Exit(0)
+		default:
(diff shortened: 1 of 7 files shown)


cmd/dm-syncer/main.go
```diff
@@ -36,13 +36,15 @@ func main() {
...
-	switch errors.Cause(err) {
-	case nil:
-	case flag.ErrHelp:
-		os.Exit(0)
-	default:
-		log.L().Error("parse cmd flags err " + err.Error())
-		os.Exit(2)
+	if err != nil {
+		switch errors.Cause(err) {
+		case nil:
```
kennytm: No need to keep the `case nil` here since `err != nil`... Why add the `if err != nil` anyway? `errors.Cause(nil) == nil`.

WangXiangUSTC: you are right, update in 6800d8b

cmd/dm-syncer/main.go
```diff
@@ -63,7 +63,13 @@ func main() {
 		log.L().Info("", zap.Stringer("dm-syncer conf", conf))
 	})
 
-	sync := syncer.NewSyncer(conf)
+	conf2, err := conf.DecryptPassword()
```
csuzhangxc: Can we let the user use an unencrypted password just as old _Syncer_ did?

WangXiangUSTC: update in 820012b

Issue #475, opened 2020-02-11
English UI for DM portal


george24601:
## Feature Request

**Is your feature request related to a problem? Please describe:**
DM portal is difficult to use for our non-Chinese engineers, since the portal is all in Chinese. 

**Describe the feature you'd like:**
English UI for DM portal

**Describe alternatives you've considered:**
N/A

**Teachability, Documentation, Adoption, Migration Strategy:**

Part of i18n

csuzhangxc:
@baurine PTAL

baurine:
Got, will add i18n support asap.

WangXiangUSTC:
finished in https://github.com/pingcap/dm/pull/476

PR #476, opened 2020-02-12
dm-portal: add i18n support

baurine:

Add i18n support for DM Portal, satisfy #475 feature request.

You can choose which language want to use.


WangXiangUSTC:
LGTM

PR #482, opened 2020-02-19
dm-syncer: add config for task name

WangXiangUSTC:
fix issue https://github.com/pingcap/dm/issues/478
if all dm-syncer use one task name, they will update same checkpoint table.

add a config for task name, and it can't be empty

### Check List 

Tests 

 - Integration test


GregoryIan:
@WangXiangUSTC Please fix CI

tests/dm_syncer/run.sh
```diff
@@ -50,7 +50,7 @@ function run() {
     cat $cur/conf/dm-syncer-2.toml > $WORK_DIR/dm-syncer-2.toml
     cat $cur/conf/old_meta_file > $WORK_DIR/old_meta_file
     # $worker1_run_source_1 > 0 means source1 is operated to worker1
-    worker1_run_source_1=$(sed "s/$SOURCE_ID1/$SOURCE_ID1\n/g" $WORK_DIR/worker1/log/dm-worker.log | grep -c "$SOURCE_ID1")
+    worker1_run_source_1=$(sed "s/$SOURCE_ID1/$SOURCE_ID1\n/g" $WORK_DIR/worker1/log/dm-worker.log | grep -c "$SOURCE_ID1") || true
```
WangXiangUSTC: if can't find `SOURCE_ID1` in  log, will exit with 1, so just add `|| true`

cmd/dm-syncer/config.go
```diff
@@ -99,7 +102,7 @@ func (c *commonConfig) newConfigFromOldConfig(args []string) (*config.SubTaskCon
 	fs.IntVar(&cfg.MaxRetry, "max-retry", 100, "maxinum retry when network interruption")
 	fs.BoolVar(&cfg.EnableANSIQuotes, "enable-ansi-quotes", false, "enable ANSI_QUOTES sql_mode")
 	fs.StringVar(&cfg.TimezoneStr, "timezone", "", "target database timezone location string")
-	fs.BoolVar(&OldConfigFormat, "old-config-format", false, "read old config format")
+	fs.BoolVar(&SyncerConfigFormat, "syncer-config-format", false, "read dyncer config format")
```
csuzhangxc:
```suggestion
	fs.BoolVar(&SyncerConfigFormat, "syncer-config-format", false, "read syncer config format")
```

cmd/dm-syncer/config.go
```diff
@@ -302,14 +308,14 @@ func loadMetaFile(metaFile string) (*config.Meta, error) {
...
+func (oc *syncerConfig) convertToNewFormat() (*config.SubTaskConfig, error) {
 	meta, err := loadMetaFile(oc.Meta)
 	if err != nil {
 		return nil, errors.Trace(err)
 	}
 	newTask := &config.SubTaskConfig{
-		Name:     "dm-syncer",
-		SourceID: "dm-syncer-from-old-config",
+		Name:     oc.Name,
+		SourceID: fmt.Sprintf("%s:%d", oc.From.Host, oc.From.Port),
```
csuzhangxc: If the user config `From` to another instance in the same replica group and enabled GTID mode (supported later), then the checkpoint can't be used anymore?

cmd/dm-syncer/config.go
```diff
@@ -34,6 +34,8 @@ import (
 type commonConfig struct {
 	*flag.FlagSet `json:"-"`
 
+	// task name
+	Name         string
```
csuzhangxc: Do we need to ensure it's not empty?

WangXiangUSTC: this will be checked by `subtask config`

cmd/dm-syncer/config.go
WangXiangUSTC: update in 41f8c7a

PR #484, opened 2020-02-19
*: export `advertise-addr` as the command arg; use `master-addr`

csuzhangxc:

export `advertise-addr` to make DM easier to use in cloud.

export `advertise-addr` as the command arg; use `master-addr`

### Check List 

Tests 

 - Unit test
 - Integration test

### DIFF
--- a/dm/master/config.go
+++ b/dm/master/config.go
@@ -58,6 +58,7 @@ func NewConfig() *Config {
 	fs.BoolVar(&cfg.printSampleConfig, "print-sample-config", false, "print sample config file of dm-worker")
 	fs.StringVar(&cfg.ConfigFile, "config", "", "path to config file")
 	fs.StringVar(&cfg.MasterAddr, "master-addr", "", "master API server and status addr")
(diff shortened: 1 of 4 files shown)


csuzhangxc:

dm/master/config.go
```diff
@@ -202,18 +203,25 @@ func (c *Config) configFromFile(path string) error {
...
 
-	// AdvertiseAddr's format must be "host:port"
-	host, port, err := net.SplitHostPort(c.AdvertiseAddr)
-	if err != nil {
-		return terror.ErrMasterAdvertiseAddrNotValid.Delegate(err, c.AdvertiseAddr)
-	}
-	if len(host) == 0 || len(port) == 0 {
-		return terror.ErrMasterAdvertiseAddrNotValid.Generate(c.AdvertiseAddr)
+	if c.AdvertiseAddr == "" {
+		if host == "" || host == "0.0.0.0" {
```
lichunzhu: Should we check port here?

csuzhangxc: addressed in https://github.com/pingcap/dm/pull/484/commits/e7b1f86e81423420748b3241c0cdcba3193b42f4.

PR #486, opened 2020-02-20
ha: minor fix on election's unit test

WangXiangUSTC:

may miss `IsNotLeader` from `e2.leaderCh` if cost more than 100ms, and will cause to fail in L152

fix it

### Check List 

Tests 

 - Unit test


WangXiangUSTC:
@lichunzhu PTAL

PR #489, opened 2020-02-20
.*: use password in integration test

WangXiangUSTC:

1. fix the bug that forget to decrypt password
2. add password in integration test

### Check List 

Tests 

 - Integration test


dm/worker/subtask.go
```diff
@@ -110,11 +110,6 @@ func NewSubTaskWithStage(cfg *config.SubTaskConfig, stage pb.Stage, etcdClient *
 
 // Init initializes the sub task processing units
 func (st *SubTask) Init() error {
-	cfgDecrypted, err := st.cfg.DecryptPassword()
```
WangXiangUSTC: when `Resume` will call `Init` again, so will DecryptPassword twice

WangXiangUSTC: https://github.com/pingcap/dm/blob/187935150417195fa738bf6a73a0107fa0b91fa1/dm/worker/subtask.go#L423 
https://github.com/pingcap/dm/blob/187935150417195fa738bf6a73a0107fa0b91fa1/dm/worker/subtask.go#L183
`Resume` -> `Run` -> `Init`

csuzhangxc: is `if !st.initialized.Get()` will be `false` when resuming? https://github.com/pingcap/dm/blob/187935150417195fa738bf6a73a0107fa0b91fa1/dm/worker/subtask.go#L422

go.mod
```diff
@@ -52,7 +52,7 @@ require (
 	github.com/spaolacci/murmur3 v1.1.0 // indirect
 	github.com/spf13/cobra v0.0.5
 	github.com/spf13/pflag v1.0.5 // indirect
-	github.com/syndtr/goleveldb v1.0.1-0.20190318030020-c3a204f8e965 // indirect
+	github.com/syndtr/goleveldb v1.0.1-0.20190318030020-c3a204f8e965
```
csuzhangxc: this is not correct? you can run `go mod tidy` again.

Is there some un-committed code in your local repo still using it?

WangXiangUSTC: yes， already revert this change

dm/worker/subtask.go
WangXiangUSTC: but may set false here when initial failed https://github.com/pingcap/dm/blob/e31e5de0c0ad0c5a399b41afb09e6d4768f37802/dm/worker/subtask.go#L121

dm/worker/server.go
```diff
@@ -147,7 +149,10 @@ func (s *Server) Start() error {
 			log.L().Error("fail to start gRPC server", log.ShortError(err2))
 		}
 	}()
-	go InitStatus(httpL) // serve status
+
+	go initStatusOnce.Do(func() {
```
WangXiangUSTC: only need do once, may cause data race in unit test

WangXiangUSTC:
@csuzhangxc @lichunzhu PTAL

csuzhangxc:
Why do we use different users for upstream MySQL (`root`) and downstream TiDB (`test`)?

WangXiangUSTC:
> Why do we use different users for upstream MySQL (`root`) and downstream TiDB (`test`)?

MySQL can begin with password, for example, use docker run MySQL:
```
docker run --rm --name mysql-3306 -p 3306:3306 -e MYSQL_ROOT_PASSWORD=123456 mysql:5.7.22 --log-bin=mysql-bin --port=3306 --bind-address=0.0.0.0 --binlog-format=ROW --server-id=1 --gtid_mode=ON --enforce-gtid-consistency=true
```

but for TiDB, the default password is empty for root, we must set password by SQL. If we use `root` for TiDB,  We must know whether to use the password or not

dm/worker/server.go
csuzhangxc: Does the unit test calls `Start` multiple times for the same `Server` instance?

dm/worker/server.go
```diff
@@ -648,18 +653,20 @@ func (s *Server) startWorker(cfg *config.SourceConfig) error {
...
-	dctx, dcancel := context.WithTimeout(s.etcdClient.Ctx(), time.Duration(len(subTaskCfgs))*3*time.Second)
-	defer dcancel()
-	minPos, err := getMinPosInAllSubTasks(dctx, subTaskCfgs)
-	if err != nil {
-		return err
-	}
+	if cfg.EnableRelay {
+		dctx, dcancel := context.WithTimeout(s.etcdClient.Ctx(), time.Duration(len(subTaskCfgs))*3*time.Second)
+		defer dcancel()
+		minPos, err := getMinPosInAllSubTasks(dctx, subTaskCfgs)
```
csuzhangxc: `declaration of "err" shadows declaration at line 637` reported by `make check`.

dm/worker/worker.go
```diff
@@ -209,6 +210,12 @@ func (w *Worker) StartSubTask(cfg *config.SubTaskConfig) {
 	// directly put cfg into subTaskHolder
 	// the unique of subtask should be assured by etcd
 	st := NewSubTask(cfg, w.etcdClient)
+	cfg2, err := cfg.DecryptPassword()
+	if err != nil {
+		st.fail(errors.Annotate(err, "start sub task"))
```
csuzhangxc: Do we need to `return` if `DecryptPassword` failed?

If needed, how about putting `DecryptPassword` after `w.subTaskHolder.recordSubTask(st)`, so we can query this error with `query-status` (because the subtask is still existing).

dm/config/subtask.go
```diff
@@ -223,27 +223,27 @@ func (c *SubTaskConfig) Toml() (string, error) {
 }
 
 // DecodeFile loads and decodes config from file
-func (c *SubTaskConfig) DecodeFile(fpath string) error {
+func (c *SubTaskConfig) DecodeFile(fpath string, decryptPassword bool) error {
```
csuzhangxc: How about choosing a better variable name rather than `decryptPassword`, this may be misleading it will update this `SubTaskConfig` with the decrypted password?

dm/worker/server.go
WangXiangUSTC: no, this update dosen't fix the data race. In fact it is conflict with etcd, you can look https://internal.pingcap.net/idc-jenkins/blue/organizations/jenkins/dm_ghpr_test/detail/dm_ghpr_test/3961/pipeline/66/
I revert this commit and fix the real data race in 091fbe5bf819bad6fa2e8eaa0c5b42a88b7a6613

tests/_utils/test_prepare
```diff
@@ -2,6 +2,10 @@ MYSQL_HOST1=${MYSQL_HOST1:-127.0.0.1}
 MYSQL_HOST2=${MYSQL_HOST2:-127.0.0.1}
 MYSQL_PORT1=${MYSQL_PORT1:-3306}
 MYSQL_PORT2=${MYSQL_PORT2:-3307}
+MYSQL_PASSWORD1=${MYSQL_PASSWORD1:-123456}
+MYSQL_PASSWORD2=${MYSQL_PASSWORD2:-123456}
+TIDB_PASSWORD=${TIDB_PASSWORD:-123456}
```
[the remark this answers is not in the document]
WangXiangUSTC: yes, I will do this later

tests/run.sh
```diff
@@ -29,15 +30,18 @@ start_services() {
     mkdir -p "$TEST_DIR"
     rm -rf "$TEST_DIR/*.log"
 
-    $CUR/_utils/run_tidb_server 4000
+    $CUR/_utils/run_tidb_server 4000 123456
```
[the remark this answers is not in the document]
WangXiangUSTC: update in 78ad70b

dm/worker/server.go
WangXiangUSTC: fix in 812b25f

dm/config/subtask.go
WangXiangUSTC: update name to `verifyDecryptPassword` in edca6b4

dm/worker/worker.go
WangXiangUSTC: yes，I fix it in edca6b4

PR #490, opened 2020-02-24
.*: add more dm-ctl info in ctl operations

lichunzhu:
Current dm-ctl operation are async. Users need to query-status and get results after they do the operations which is not easy.

dm-master will run `query-status` and check whether the result is correct and return the simplified result to users.

### Check List 

Tests 

 - Unit test
 - Integration test

Related changes


lichunzhu:

dm/master/server.go
```diff
@@ -1334,37 +1341,232 @@ func (s *Server) generateSubTask(ctx context.Context, task string) (*config.Task
...
 
-// taskConfigArgsExtractor extracts SubTaskConfig from args and returns its relevant
-// grpc client, worker id (host:port), subtask config in toml, task name and error
-func (s *Server) taskConfigArgsExtractor(cfg *config.SubTaskConfig) (*scheduler.Worker, string, string, error) {
-	handleErr := func(err error) error {
-		log.L().Error("response", zap.Error(err))
-		return err
+func extractWorkerError(result *pb.ProcessResult) error {
+	if result != nil && len(result.Errors) > 0 {
+		return terror.ErrMasterOperRespNotSuccess.Generate(result.Errors[0].String())
```
WangXiangUSTC: how about adding all error's information in `result.Errors`?

dm/master/server.go
```diff
@@ -1334,37 +1341,232 @@ func (s *Server) generateSubTask(ctx context.Context, task string) (*config.Task
...
+		},
 	}
 
-	worker := s.scheduler.GetWorkerBySource(cfg.SourceID)
-	if worker == nil {
-		return nil, "", "", handleErr(terror.ErrMasterTaskConfigExtractor.Generatef("%s relevant worker-client not found", cfg.SourceID))
+	for num := 0; num < maxRetryNum; num++ {
+		// check whether source relative worker has been removed by scheduler
+		switch masterReq.(type) {
+		case *pb.OperateSourceRequest:
```
WangXiangUSTC: seems that no need to use `switch`

dm/master/server.go
lichunzhu: revised in https://github.com/pingcap/dm/pull/490/commits/6d43a1ddce1f583c1a734b38e787d65cc9ad6261

dm/master/server.go
lichunzhu: revised in https://github.com/pingcap/dm/pull/490/commits/6d43a1ddce1f583c1a734b38e787d65cc9ad6261

dm/worker/server.go
```diff
@@ -210,20 +212,30 @@ func (s *Server) setWorker(worker *Worker, needLock bool) {
...
 }
 
-func (s *Server) setWorkerErr(workerErr error, needLock bool) {
+func (s *Server) setSourceStatus(source string, err error, needLock bool) {
 	if needLock {
 		s.Lock()
 		defer s.Unlock()
 	}
-	s.workerErr = workerErr
+	s.sourceStatus = pb.SourceStatus{
```
csuzhangxc: Do we never set `RelayStatus` for `s.sourceStatus`, right?

dm/worker/server.go
```diff
@@ -435,8 +447,7 @@ func (s *Server) QueryStatus(ctx context.Context, req *pb.QueryStatusRequest) (*
 	}
 
 	resp.SubTaskStatus = w.QueryStatus(req.Name)
-	resp.RelayStatus = relayStatus
-	resp.Source = w.cfg.SourceID
+	sourceStatus.RelayStatus = relayStatus
```
csuzhangxc: How about setting `RelayStatus` to `nil` to make the response of `query-status` clearer when `enable-relay: false`? cc @WangXiangUSTC .

dm/worker/server.go
```diff
@@ -446,24 +457,30 @@ func (s *Server) QueryStatus(ctx context.Context, req *pb.QueryStatusRequest) (*
 // QueryError implements WorkerServer.QueryError
 func (s *Server) QueryError(ctx context.Context, req *pb.QueryErrorRequest) (*pb.QueryErrorResponse, error) {
 	log.L().Info("", zap.String("request", "QueryError"), zap.Stringer("payload", req))
+	sourceStatus := s.getSourceStatus(true)
+	sourceError := pb.SourceError{
+		Source: sourceStatus.Source,
+		Worker: s.cfg.Name,
+	}
+	if sourceStatus.Result != nil && len(sourceStatus.Result.Errors) > 0 {
+		sourceError.SourceError = sourceStatus.Result.Errors[0].String()
```
csuzhangxc: Can we `join` all `Result.Errors` here? or it will have at most one error?

dm/worker/server.go
lichunzhu: Yes

csuzhangxc: Add a comment to clarify it?

dm/master/server.go
```diff
@@ -305,6 +307,7 @@ func subtaskCfgPointersToInstances(stCfgPointers ...*config.SubTaskConfig) []con
 func (s *Server) StartTask(ctx context.Context, req *pb.StartTaskRequest) (*pb.StartTaskResponse, error) {
 	log.L().Info("", zap.Stringer("payload", req), zap.String("request", "StartTask"))
 
+	resp := &pb.StartTaskResponse{}
```
csuzhangxc: You can declare this variable after `isLeaderAndNeedForward` check.

dm/worker/server.go
lichunzhu: addressed in https://github.com/pingcap/dm/pull/490/commits/b9f4793fa1b4d18afa02c9fd9987ce42991dad44

dm/worker/server.go
lichunzhu: addressed in https://github.com/pingcap/dm/pull/490/commits/b9f4793fa1b4d18afa02c9fd9987ce42991dad44

dm/worker/server.go
WangXiangUSTC: I think it's ok

PR #498, opened 2020-02-28
election: output the leader's information

WangXiangUSTC:

`pkg/election` notify master that it's leader, retire from leader or not leader at `time1`, but the master still need to query etcd to get the master's information at `time2`. But the leader information may be changed between `time1` and `time2`, master may do a wrong action.

`pkg/election` notify master the leader's information, so the master don't need to query etcd again.

### Check List 

Tests 

 - Integration test


WangXiangUSTC:
@csuzhangxc @lichunzhu PTAL

WangXiangUSTC:
@csuzhangxc @lichunzhu PTAL

pkg/election/election.go
```diff
@@ -255,67 +263,60 @@ func (e *Election) campaignLoop(ctx context.Context, session *concurrency.Sessio
...
 			cancel2()
 			compaignWg.Wait()
 			continue
 		}
 
-		if leaderID != e.info.ID {
-			e.l.Info("current member is not the leader", zap.String("current member", e.info.ID), zap.String("leader", leaderID))
-			e.isNotLeader()
+		if leaderInfo.ID != e.info.ID {
+			e.l.Info("current member is not the leader", zap.String("current member", e.info.ID), zap.String("leader", leaderInfo.ID))
```
csuzhangxc: How about logging `e.info` and `leaderInfor` directly (including `ID` and `Addr`)? If so, ditto for other places.

pkg/election/election.go
```diff
@@ -255,67 +263,60 @@ func (e *Election) campaignLoop(ctx context.Context, session *concurrency.Sessio
...
 	}
-}
 
-func (e *Election) isNotLeader() {
-	e.isLeader.Set(false)
 	select {
-	case e.leaderCh <- IsNotLeader:
-	default:
+	case e.leaderCh <- leaderInfo:
+	case <-time.After(NotifyBlockTime):
```
csuzhangxc: Why can't we use default here? Is to avoid lost notify when restarting the `Election` instance?

pkg/election/election_test.go
```diff
@@ -73,10 +75,15 @@ func (t *testElectionSuite) SetUpTest(c *C) {
 	case <-time.After(10 * time.Second):
 		c.Fatal("start embed etcd timeout")
 	}
+
+	// some notify leader information is not handled, just reduce the block time and ignore them
+	t.originNotifyBlockTime = NotifyBlockTime
+	NotifyBlockTime = 100 * time.Millisecond
```
csuzhangxc: How about passing the block time to `NewElection` as an argument?

dm/master/election.go
```diff
@@ -47,29 +59,14 @@ func (s *Server) electionNotify(ctx context.Context) {
...
-				}
+			} else {
+				// this member is not leader
+				log.L().Info("get new leader", zap.String("leader", leaderInfo.ID), zap.String("current member", s.cfg.Name))
 
 				s.Lock()
-				s.leader = leader
-				s.createLeaderClient(leaderAddr)
+				s.leader = leaderInfo.ID
+				s.createLeaderClient(leaderInfo.Addr)
```
csuzhangxc: `s.Lock` is used to protect `s.createLeaderClient(leaderInfo.Addr)`, and no need to protect `s.leader`, right?

pkg/election/election.go
```diff
@@ -255,67 +263,60 @@ func (e *Election) campaignLoop(ctx context.Context, session *concurrency.Sessio
...
+		e.isLeader.Set(false)
 	}
-}
 
-func (e *Election) isNotLeader() {
-	e.isLeader.Set(false)
 	select {
-	case e.leaderCh <- IsNotLeader:
-	default:
+	case e.leaderCh <- leaderInfo:
```
lichunzhu: `ctx` is not used in `notifyLeader`

pkg/election/election.go
```diff
@@ -255,67 +263,60 @@ func (e *Election) campaignLoop(ctx context.Context, session *concurrency.Sessio
...
+		if leaderInfo.ID != e.info.ID {
+			e.l.Info("current member is not the leader", zap.String("current member", e.info.ID), zap.String("leader", leaderInfo.ID))
+			e.notifyLeader(ctx, leaderInfo)
 			cancel2()
 			compaignWg.Wait()
 			continue
 		}
 
-		e.toBeLeader() // become the leader now
+		e.notifyLeader(ctx, leaderInfo) // become the leader now
```
lichunzhu: Should we add a log before this line?
```go
		e.l.Info("become leader", zap.String("current member", e.info.ID))
```

WangXiangUSTC: add in 2c624cc

pkg/election/election.go
WangXiangUSTC: remove in 2c624cc

pkg/election/election.go
WangXiangUSTC: update in 2c624cc

pkg/election/election.go
WangXiangUSTC: if use default, the master may lose some notify if the master is busy with old notify, for example, start the `scheduer` when it becomes leader, it will cost some time.

dm/master/election.go
WangXiangUSTC: also need protect `s.leader`, because need to get `s.leader` when call function `isLeaderAndNeedForward`

pkg/election/election_test.go
WangXiangUSTC: update in 106b035

dm/master/election.go
csuzhangxc: Do we need to protect `s.leader = ""` in L41?

pkg/election/election.go
WangXiangUSTC: add in e3cce1e, add `<-ctx.Done()` to avoid block when ctx is canceled

dm/master/election.go
WangXiangUSTC: yes, I add lock in a33eecb

dm/master/election.go
```diff
@@ -38,11 +38,15 @@ func (s *Server) electionNotify(ctx context.Context) {
...
 					s.scheduler.Close()
+					s.Lock()
 					s.leader = ""
+					s.Unlock()
 
 					log.L().Info("current member retire from the leader", zap.String("current member", s.cfg.Name))
 				}
+				s.Lock()
 				s.closeLeaderClient()
+				s.Unlock()
```
csuzhangxc: How about putting these two `Lock` into only one to ensure the state is consistent?

WangXiangUSTC: update in 0a352a3

dm/master/election.go
```diff
@@ -83,10 +86,7 @@ func (s *Server) electionNotify(ctx context.Context) {
 }
 
 func (s *Server) createLeaderClient(leaderAddr string) {
-	if s.leaderGrpcConn != nil {
-		s.leaderGrpcConn.Close()
-		s.leaderGrpcConn = nil
-	}
+	s.closeLeaderClient()
```
lichunzhu: Should we lock here?

Issue #509, opened 2020-03-03
DM-worker panic when query-status or query-error


csuzhangxc:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do? If possible, provide a recipe for reproducing the error.

`query-error`.

2. What did you expect to see?

return anything.

3. What did you see instead?

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x58 pc=0x1867bff]

goroutine 17712 [running]:
github.com/pingcap/dm/dm/worker.(*Worker).Error(0xc00012a000, 0xc002c95250, 0x9, 0x114, 0x0, 0x0)
	/home/jenkins/agent/workspace/build_dm_master/go/src/github.com/pingcap/dm/dm/worker/status.go:161 +0x29f
github.com/pingcap/dm/dm/worker.(*Worker).QueryError(0xc00012a000, 0xc002c95250, 0x9, 0x0, 0x0, 0x0)
	/home/jenkins/agent/workspace/build_dm_master/go/src/github.com/pingcap/dm/dm/worker/worker.go:348 +0x110
github.com/pingcap/dm/dm/worker.(*Server).QueryError(0xc000500080, 0x206eb60, 0xc0008ea150, 0xc0049ff470, 0xc000500080, 0xc0008ea150, 0xc000285a80)
	/home/jenkins/agent/workspace/build_dm_master/go/src/github.com/pingcap/dm/dm/worker/server.go:284 +0x2b9
github.com/pingcap/dm/dm/pb._Worker_QueryError_Handler(0x1c75ca0, 0xc000500080, 0x206eb60, 0xc0008ea150, 0xc00365a540, 0x0, 0x206eb60, 0xc0008ea150, 0xc002c95240, 0xb)
	/home/jenkins/agent/workspace/build_dm_master/go/src/github.com/pingcap/dm/dm/pb/dmworker.pb.go:3811 +0x217
google.golang.org/grpc.(*Server).processUnaryRPC(0xc00048f800, 0x208c180, 0xc0001b0600, 0xc005fa4e00, 0xc000442f60, 0x33b1fc0, 0x0, 0x0, 0x0)
	/go/pkg/mod/google.golang.org/grpc@v1.25.1/server.go:1007 +0x460
google.golang.org/grpc.(*Server).handleStream(0xc00048f800, 0x208c180, 0xc0001b0600, 0xc005fa4e00, 0x0)
[...]

PR #512, opened 2020-03-03
dm-worker/: refine query error

lichunzhu:
https://github.com/pingcap/dm/issues/509
Current DM-worker may have no `currUnit` when some error occurs during `Init`, which will cause **nil-pointer error** while using `worker.QueryError`.

Avoid query `currUnit.Error` when `currUnit` is nil. Use `result.Errors` as Error when `currUnit` is nil.

### Check List 

Tests 

 - Unit test
 - Integration test

Related changes


dm/worker/status.go
```diff
@@ -158,26 +165,33 @@ func (w *Worker) Error(stName string) []*pb.SubTaskError {
...
-			case pb.UnitType_Check:
-				stError.Error = &pb.SubTaskError_Check{Check: us.(*pb.CheckError)}
-			case pb.UnitType_Dump:
-				stError.Error = &pb.SubTaskError_Dump{Dump: us.(*pb.DumpError)}
-			case pb.UnitType_Load:
-				stError.Error = &pb.SubTaskError_Load{Load: us.(*pb.LoadError)}
-			case pb.UnitType_Sync:
-				stError.Error = &pb.SubTaskError_Sync{Sync: us.(*pb.SyncError)}
+			if cu != nil {
+				// oneof error
```
WangXiangUSTC:
```suggestion
				// one of error
```

WangXiangUSTC:
LGTM

Issue #404, opened 2019-12-06
task configuration optimization


amyangfei:
# DM task configuration optimization

# Abstract

In this project we will focus on the optimization of DM task configuration and try to achieve some improvement in DM usability. The optimization includs the following items:

- Extend the wildcard matching ability used in `table-routing` and `binlog event filter`.
- Unify the matching strategy used in `table-routing`, `binlog event filter` and `black and white list`.
- Improve existing configuration related tools to make them more useful.

### Problem Statement

- The wildcard matching strategy used in `table-routing` and `binlog event filter` supports `*` and `?`, we want to extend them to support more matching rules, including the `[...]` matching rule.
  - Wildcard varies in different domains, we should first determine which wildcard regulation will be used. In this project we suggest to use [the most common syntax](https://en.wikipedia.org/wiki/Glob_(programming)#Syntax), which contains `*`, `?` and `[...]`.
  - As we use a trie tree to store matching rules, it is easy to implement positive matching but it is difficult to implement the negative matching, which often in the `[!...]` format, this will be a bonus of this problem.
- There exists three table/binlog matching features in DM, `table routing`, `black and white list` and `binlog event filter`. The matching rules used in these three features are not unified, which often makes user confused and difficult to use.
- Many data replication scenarios have large scale of tables and complicated replication requirements, in which cases user have to write a long, complicated task config.
- We have already developed some config related tools, including [DM-portal](https://docs.google.com/document/d/13ilFXJZErBXBRRREcskd5zhtocF-apLxFICp45OWbfs/edit) (already released) and [DM-simulator](https://github.com/pingcap/dm/pull/371) (PR in review). However the DM-portal is not up to date according to the recent config refactor. We need to find out whether DM-portal is a proper way to solve the configuration automatic generation problem, if it is the correct way, we should make the config strategy generated by it up to date, otherwise we could try to come up with a better solution.

# Success Criteria

- Extend the wildcard matching strategy, support `[...]` type matching in everywhere we use wildcard matching in DM.
- Support wildcard in black and white list feature, but still keep the original regular expression way.
- ~~Make `DM-portal` to keep up with current DM configuration regulation.~~
- Make `DM-simulator` come true.

# TODO list

- [ ]  Extend wildcard matching, support `[...]` type matching.
- [ ]  Extend wildcard matching, support negative `[...]` type matching.
- [ ]  Support wildcard matching in black and white list feature.
[...]

PR #515, opened 2020-03-05
go.mod: update tidb-tools dependency to v3.0.11

lichunzhu:
In #404, we have updated tidb-tools to support wildcard matching better. We should update tidb-tools dependency to use these new features.

Update tidb-tools dependency in go.mod.

### Check List 

Tests 

 - Unit test
 - Integration test

Related changes


WangXiangUSTC:
LGTM

Issue #517, opened 2020-03-05
HA: Handle error when using etcd client's `Get with revision`


lichunzhu:
## Feature Request

**Is your feature request related to a problem? Please describe:**

Current uses of etcd client's `clientv3.Get(key, clientv3.WithRev())` contain two situations:
1. Get relative etcd info while receive info from `clientv3.Watcher`.
2. Get relative etcd and revision info before we start to watch.

**Describe the feature you'd like:**

For 1, reset watcher like we did in https://github.com/pingcap/dm/pull/499.
For 2, get relative etcd info in one etcd transaction to avoid compaction.

PR #499, opened 2020-02-28
dm/master, dm/worker: handle etcd retryable errors for watcher

lichunzhu:
### What problem does this PR solve? 
https://github.com/pingcap/dm/issues/501
When receiving etcd compaction error from wather, the handle goroutine will lose event if we don't do anything.

### What is changed and how it works?
1. Rebuild info and restart handle etcd event routine when receiving etcd retryable error (`ErrCompacted, ErrNoLeader, ErrNoSpace`).
2. Return `resp.Header.Revision` even if we don't get value from etcd server.

### Check List 

Tests 

 - Unit test
 - Integration test

Side effects

 - Increased code complexity

Related changes

 - Need to cherry-pick to the release branch
 - Need to update the `dm/dm-ansible`

PR #518, opened 2020-03-05
.*/: handle etcd retryable error when using etcd client's `Get with revision`

lichunzhu:
complete https://github.com/pingcap/dm/issues/517

1. Reset get error in watcher like we did in https://github.com/pingcap/dm/pull/499.
2. Get relative etcd info in one etcd transaction to avoid using get with rev.

### Check List 

Tests 

 - Unit test
 - Integration test


dm/worker/server.go
```diff
@@ -361,11 +365,12 @@ func (s *Server) operateSourceBound(bound ha.SourceBound) error {
 	if bound.IsDeleted {
 		return s.stopWorker(bound.Source)
 	}
-	sourceCfg, _, err := ha.GetSourceCfg(s.etcdClient, bound.Source, bound.Revision)
+	scm, _, err := ha.GetSourceCfg(s.etcdClient, bound.Source, bound.Revision)
 	if err != nil {
 		// TODO: need retry
 		return err
 	}
+	sourceCfg := scm[bound.Source]
```
csuzhangxc: Do we need to check empty sourceCfg here?

pkg/ha/stage.go
```diff
@@ -263,6 +289,32 @@ func subTaskStageFromKey(key string) (Stage, error) {
...
+func subTaskStageFromResp(source, task string, resp *clientv3.GetResponse) (map[string]map[string]Stage, error) {
+	stages := make(map[string]map[string]Stage)
+	if source != "" {
+		stages[source] = make(map[string]Stage) // avoid stages[source] is nil
+	}
+
+	if resp.Count == 0 {
+		return stages, nil
+	} else if source != "" && task != "" && resp.Count > 1 {
+		return stages, fmt.Errorf("too many stage (%d) exist for subtask {sourceID: %s, task name: %s}", resp.Count, source, task)
```
WangXiangUSTC: how about add a TODO to use terror?

PR #528, opened 2020-03-10
pkg/: add retry for etcd put/delete operation

lichunzhu:
When etcd PUT/DELETE is failed because of some network problems or leader not exist problems, it's OK to retry them. Putting twice into the etcd is still fine for currently used PUT/DELETE operations.

Add retry for some errors of etcd PUT/DELETE errors.

### Check List 

Tests 

 - Unit test
 - Integration test

Side effects

 - Increased code complexity


pkg/etcdutil/etcdutil.go
```diff
@@ -33,6 +37,27 @@ const (
...
+	RetryCount:         5,
+	FirstRetryDuration: time.Second,
+	BackoffStrategy:    retry.Stable,
+	IsRetryableFn: func(retryTime int, err error) bool {
+		switch err {
+		// Etcd ResourceExhausted errors, may recover after some time
+		case v3rpc.ErrNoSpace, v3rpc.ErrTooManyRequests:
+			return true
+		// Etcd Unavailable errors, may be available after some time
+		case v3rpc.ErrNoLeader, v3rpc.ErrLeaderChanged, v3rpc.ErrNotCapable, v3rpc.ErrStopped, v3rpc.ErrTimeout,
```
csuzhangxc: Can it be successful after retry for `ErrNotCapable` and `ErrStopped`?

lichunzhu: I think it can be successful.
https://github.com/etcd-io/etcd/pull/9934/files#diff-6d8785d0c9eaf96bc3e2b29c36493c04R162-R167
ErrStopped:
// one of the etcd nodes stopped from failure injection
ErrNotCapable:
// capability check has not been done (in the beginning)

WangXiangUSTC: how  about add in comment

lichunzhu:
@WangXiangUSTC @csuzhangxc PTAL again

PR #529, opened 2020-03-10
election: reset oldLeaderID after retire from leader

WangXiangUSTC:

when master retire from leader, not set `oldLeaderInfo = ""`, and if this master become leader again, will contine at L273 and not notify the master 

set `oldLeaderInfo = ""` after retire from leader


csuzhangxc:
@lichunzhu PTAL

Issue #526, opened 2020-03-09
Replace mydumper in DM to dumpling


lichunzhu:
## Feature Request

**Is your feature request related to a problem? Please describe:**

Now we dump sql using mydumper. It's not easy to use for users because they need to download a mydumper binary file.
**Describe the feature you'd like:**

Integration dumpling into DM and replace part mydumper.

csuzhangxc:
already done in #540.

PR #540, opened 2020-03-13
dump/: integrate dumpling

lichunzhu:
https://github.com/pingcap/dm/issues/526

Dump sql using dumpling. It's not easy to use for users because they need to download a mydumper binary file.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

 - Has persistent data change

Side effects

Related changes


lichunzhu:

tests/all_mode/data/db1.increment.sql
```diff
@@ -25,7 +25,7 @@ update t1 set name = 'gentestxxxxxx' where gen_id = 124;
 delete from t1 where gen_id > 124;
 
 -- test alter database
-alter database all_mode CHARACTER SET = utf8mb4;
+-- alter database all_mode CHARACTER SET = utf8mb4;
```
csuzhangxc: Is it still not support this?

tests/print_status/conf/dm-task.yaml
```diff
@@ -30,7 +30,8 @@ mydumpers:
     threads: 4
     chunk-filesize: 0
     skip-tz-utc: true
-    extra-args: "--statement-size=5000"
+    statement-size: 4000
```
csuzhangxc: Why change `5000` to `4000` here?

tests/sequence_sharding/run.sh
```diff
@@ -33,7 +33,7 @@ function run() {
     run_sql_file $cur/data/db1.increment.sql $MYSQL_HOST1 $MYSQL_PORT1 $MYSQL_PASSWORD1
     run_sql_file $cur/data/db2.increment.sql $MYSQL_HOST2 $MYSQL_PORT2 $MYSQL_PASSWORD2
 
-    sleep 3
+    sleep 10
```
csuzhangxc: why do we need to change from `3` to `10`?

dm/worker/metrics.go
```diff
@@ -62,6 +63,7 @@ func RegistryMetrics() {
 
 	relay.RegisterMetrics(registry)
 	mydumper.RegisterMetrics(registry)
+	dumpling.RegisterMetrics(registry)
```
csuzhangxc: Do we need to remove `mydumper.RegisterMetrics(registry)` now?

dumpling/dumpling.go
```diff
@@ -0,0 +1,246 @@
...
+	})
+
+	// NOTE: remove output dir before start dumping
+	// every time re-dump, loader should re-prepare
+	err := os.RemoveAll(m.cfg.Dir)
+	if err != nil {
+		m.logger.Error("fail to remove output directory", zap.String("directory", m.cfg.Dir), log.ShortError(err))
+	}
+
+	// Cmd cannot be reused, so we create a new cmd when begin processing
```
csuzhangxc: We can remove this log now.

dumpling/dumpling.go
```diff
@@ -0,0 +1,246 @@
...
+		ret = append(ret, "--skip-tz-utc")
+	}
+
+	extraArgs := strings.Fields(cfg.ExtraArgs)
+	if len(extraArgs) > 0 {
+		ret = append(ret, ParseArgLikeBash(extraArgs)...)
+	}
+
+	// TODO: support String for export.Config
+	// m.logger.Info("create dumpling", zap.Stringer("config", dumpConfig))
```
csuzhangxc: uncomment this line or just remove it.

dumpling/dumpling.go
```diff
@@ -0,0 +1,246 @@
...
+func (m *Dumpling) IsFreshTask(ctx context.Context) (bool, error) {
+	return true, nil
+}
+
+// constructArgs constructs arguments for exec.Command
+func (m *Dumpling) constructArgs() (*export.Config, error) {
+	cfg := m.cfg
+	db := cfg.From
+
+	dumpConfig := export.DefaultConfig()
```
csuzhangxc: 1. Can we disable `StatusAddr` now? because we already have it in DM, and if we enable it, may we need more ports for the process.
2. Do we need to set `ServerInfo` based on `flavor`?

tests/all_mode/data/db1.increment.sql
lichunzhu: This can't be passed because the default CHARACTER is set to latin1. If we revise this line to
```sql
-- alter database all_mode CHARACTER SET = latin1;
```
I think we can pass it.

tests/print_status/conf/dm-task.yaml
lichunzhu: Because the difference of the meaning of statement-size between mydumper and dumpling. dumpling only counts the data's size, not counting the joiners. So I shorten this to split it to more files.

dumpling/dumpling.go
lichunzhu: 1. I will block it.
2. dumpling will adjust `ServerInfo` during the `Dump` function.

lichunzhu:
https://github.com/pingcap/dumpling/pull/51
dumpling has proposed a PR to combine logger with DM. This PR should include this feature and be merged after this PR is merged.

dumpling/dumpling.go
```diff
@@ -0,0 +1,228 @@
...
+		m.logger.Info("dump unit runs forever", zap.String("failpoint", "dumpUnitProcessForever"))
+		<-ctx.Done()
+		failpoint.Return()
+	})
+
+	// NOTE: remove output dir before start dumping
+	// every time re-dump, loader should re-prepare
+	err := os.RemoveAll(m.cfg.Dir)
+	if err != nil {
+		m.logger.Error("fail to remove output directory", zap.String("directory", m.cfg.Dir), log.ShortError(err))
```
WangXiangUSTC: do we need to append err to `errs` and then return? If the old dir is not removed, may load wrong data to upstream

dumpling/dumpling.go
```diff
@@ -0,0 +1,228 @@
...
+
+	err = export.Dump(m.dumpConfig)
+
+	if err != nil {
+		dumplingExitWithErrorCounter.WithLabelValues(m.cfg.Name).Inc()
+		errs = append(errs, unit.NewProcessError(err))
+	} else {
+		select {
+		case <-ctx.Done():
+			isCanceled = true
```
WangXiangUSTC: what is the `isCanceled` used for?

dumpling/dumpling.go
```diff
@@ -0,0 +1,228 @@
...
+		dumpConfig.StatementSize = cfg.StatementSize
+	}
+	if cfg.Rows > 0 {
+		dumpConfig.Rows = cfg.Rows
+	}
+	if len(cfg.Where) > 0 {
+		dumpConfig.Where = cfg.Where
+	}
+
+	var ret []string
```
WangXiangUSTC: `ret` is used to save some unsupported arguments? how about adding a comment

dumpling/util.go
```diff
@@ -0,0 +1,39 @@
+// Copyright 2019 PingCAP, Inc.
```
WangXiangUSTC:
```suggestion
// Copyright 2020 PingCAP, Inc.
```

dumpling/dumpling.go
```diff
@@ -0,0 +1,228 @@
+// Copyright 2019 PingCAP, Inc.
```
WangXiangUSTC:
```suggestion
// Copyright 2020 PingCAP, Inc.
```

dumpling/dumpling.go
```diff
@@ -0,0 +1,228 @@
...
+
+	var ret []string
+	if cfg.SkipTzUTC {
+		// TODO: support skip-tz-utc
+		ret = append(ret, "--skip-tz-utc")
+	}
+
+	extraArgs := strings.Fields(cfg.ExtraArgs)
+	if len(extraArgs) > 0 {
+		ret = append(ret, ParseArgLikeBash(extraArgs)...)
```
WangXiangUSTC: do we need to parse the `extra-args`, like some config in the integration test, the user may set `extra-args: "--statement-size=100"`,  it may cause incompatible problem

Issue #525, opened 2020-03-09
Revoke keepalive lease when dm-worker is closed normally


lichunzhu:
## Feature Request

**Is your feature request related to a problem? Please describe:**

1. After the master goes offline, it may wait for `ttl` before the cluster will be re-elected.
2. After the worker goes offline, It may wait for `ttl`(10s) before the source is released to another worker.

**Describe the feature you'd like:**

Revoke lease when we are trying to offline master/worker normally, like what we did in https://github.com/pingcap/dm/blob/c2da436a9ebd55a896fa55cb4a63878f414f06a6/pkg/ha/keepalive.go#L108.

PR #548, opened 2020-03-16
pkg/keepalive: revoke lease if we quit normally

lichunzhu:
https://github.com/pingcap/dm/issues/525

Revoke lease before we quit normally once we put a key with lease successfully.

### Check List 

Tests 

 - Unit test
 - Integration test


pkg/ha/keepalive.go
```diff
@@ -104,14 +106,17 @@ func KeepAlive(ctx context.Context, cli *clientv3.Client, workerName string, kee
...
 			log.L().Info("ctx is canceled, keepalive will exit now")
-			ctx, cancel := context.WithTimeout(cli.Ctx(), etcdutil.DefaultRequestTimeout)
-			cli.Revoke(ctx, lease.ID)
-			cancel()
 			return nil
 		}
 	}
 }
 
+func revokeLease(ctx context.Context, cli *clientv3.Client, id clientv3.LeaseID) (*clientv3.LeaseRevokeResponse, error) {
```
csuzhangxc: this `ctx` is not used.

pkg/ha/keepalive.go
```diff
@@ -104,14 +106,17 @@ func KeepAlive(ctx context.Context, cli *clientv3.Client, workerName string, kee
...
+	ctx, cancel := context.WithTimeout(cli.Ctx(), etcdutil.DefaultRevokeLeaseTimeout)
```
csuzhangxc: how about adding a comment to highlight _we must ensure `cli.Ctx()` not done when exiting_ (do not set `Context` when creating `cli` or do not cancel this `Context` or it's parent context).

lichunzhu: Addressed in https://github.com/pingcap/dm/pull/548/commits/6e69519b2d12e077049442d350f92c75ff44fc43

PR #559, opened 2020-03-20
worker/: don't quit worker immediately when dialed master is offline

lichunzhu:
https://github.com/pingcap/dm/issues/560
When worker dials an offline master, it will returns error directly and stop the worker process.

1. add `WithBlock` option for grpc dial to avoid dialing an offline master.
2. don't return error immediately when dialed master is offline. Try to dial other endpoint in `cfg.Join`.
3. do the register request in the loop to avoid dialing an online master which is blocked from other dm-masters.

### Check List 

Tests 

 - Unit test
 - Integration test

### DIFF
--- a/dm/worker/join.go
+++ b/dm/worker/join.go
@@ -36,32 +36,36 @@ func GetJoinURLs(addrs string) []string {
 // JoinMaster let dm-worker join the cluster with the specified master endpoints.
 func (s *Server) JoinMaster(endpoints []string) error {
 	// TODO: grpc proxy
-	var client pb.MasterClient
-	for _, endpoint := range endpoints {
-		conn, err := grpc.Dial(endpoint, grpc.WithInsecure(), grpc.WithBackoffMaxDelay(3*time.Second))
-		if err != nil {
-			return err
-		}
-		client = pb.NewMasterClient(conn)
-		break
-	}
-	if client == nil {
-		return errors.Errorf("cannot connect with master endpoints: %v", endpoints)
-	}
-	ctx, cancel := context.WithTimeout(context.Background(), 3*time.Second)
+	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
 	defer cancel()
+
 	req := &pb.RegisterWorkerRequest{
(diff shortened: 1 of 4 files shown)


dm/worker/join.go
```diff
@@ -36,32 +36,38 @@ func GetJoinURLs(addrs string) []string {
...
+		if err != nil {
+			log.L().Error("fail to dial dm-master", zap.Error(err))
+			continue
+		}
+		client = pb.NewMasterClient(conn)
+		ctx1, cancel1 = context.WithTimeout(ctx, 3*time.Second)
+		resp, err := client.RegisterWorker(ctx1, req)
+		cancel1()
+		if err != nil {
+			log.L().Error("fail to register worker: %s", zap.Error(err))
```
csuzhangxc:
```suggestion
			log.L().Error("fail to register worker", zap.Error(err))
```

ditto for L65.

please also do this in `cmd/dm-worker/main.go` for `join the cluster meet error %v`.

dm/worker/join.go
```diff
@@ -36,32 +36,38 @@ func GetJoinURLs(addrs string) []string {
...
+			continue
+		}
+		if !resp.GetResult() {
+			log.L().Error("fail to register worker", zap.String("error", resp.Msg))
+			continue
+		}
+		return nil
 	}
-	return nil
+	return errors.Errorf("cannot connect with master endpoints: %v", endpoints)
```
WangXiangUSTC: how about add a TODO: use terror

dm/worker/join.go
```diff
@@ -36,32 +36,38 @@ func GetJoinURLs(addrs string) []string {
...
 		Address: s.cfg.AdvertiseAddr,
 	}
-	resp, err := client.RegisterWorker(ctx, req)
-	if err != nil {
-		return err
-	}
-	if !resp.GetResult() {
-		return errors.Errorf("fail to register worker: %s", resp.GetMsg())
+
+	var client pb.MasterClient
```
WangXiangUSTC: seems that we don't need to define here

Issue #565, opened 2020-03-20
HA: some rows is not replicated in shard mode with sysbench


csuzhangxc:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do? If possible, provide a recipe for reproducing the error.

   - use sysbench to generate for one of tables in the upstream
   - use DM to replicate these data in shard mode

2. What did you expect to see?

    replication works as normal.

3. What did you see instead?

   some data lost.

4. Versions of the cluster

    - DM version (run `dmctl -V` or `dm-worker -V` or `dm-master -V`):

        ```
        DM nightly before March 20.
        ```

WangXiangUSTC:
fix in https://github.com/pingcap/dm/pull/572

PR #572, opened 2020-03-30
syncer: minor fix

WangXiangUSTC:

1. fix an issue that may cause data loss in shard situation #565 
when `re-replicate shard group` was completed, will call `closeShardingResync` and then return, so this event will be lost.
2. fix an issue that may cause `inconsistent DDL sequence` error.

1. only when the streamer is redirected then return (ignore the binlog event)
2. update `lastLocation` to the location which used for redirect streamer.
3. add test for it (simulate the customer's scenario)

### Check List 

Tests 

 - Integration test


csuzhangxc:
@lichunzhu PTAL

syncer/syncer.go
```diff
@@ -1075,27 +1075,33 @@ func (s *Syncer) Run(ctx context.Context) (err error) {
...
 			}
 
 			currentLocation = nextLocation.Clone()
 			err2 = s.streamerController.RedirectStreamer(s.tctx, nextLocation.Clone())
 			if err2 != nil {
-				return err2
+				return false, err2
 			}
+			redirect = true
+			lastLocation = nextLocation.Clone()
```
WangXiangUSTC: after redirect streamer, the `lastLocation` should be the begin of the streamer

tests/sharding2/data/db2.increment.sql
```diff
@@ -0,0 +1,6 @@
+create database sharding22;
+create table sharding22.t1 (id int primary key, name varchar(20));
+alter table sharding22.t1 add index abc(name);
+/*insert into sharding22.t1 values(7, "a"), (8, "b");*/
```
lichunzhu: Why do we comment these dmls?

tests/sharding2/data/db1.increment.sql
```diff
@@ -0,0 +1,6 @@
+create database sharding22;
+create table sharding22.t1 (id int primary key, name varchar(20));
+alter table sharding22.t1 add index abc(name);
+insert into sharding22.t1 values(1, "a"), (2, "b");
+insert into sharding22.t1 values(3, "c"), (4, "d");
+insert into sharding22.t1 values(5, "e"), (6, "f");
```
lichunzhu: How about adding more continuous ddls to test? IMO this bug is caused because of continuous ddls.
For example,
```sql
create table t3 (id bigint auto_increment,uid int,name varchar(20),primary key (`id`),unique key(`uid`)) DEFAULT CHARSET=utf8mb4;
alter table t3 add column c int;
alter table t3 add index c(c);
alter table t3 add column d int;
alter table t3 add index d(d);
alter table t3 add column e int, add index e(e);
```

WangXiangUSTC: add in 30e8adb

tests/sharding2/data/db2.increment.sql
WangXiangUSTC: update in 30e8adb

csuzhangxc:
@lichunzhu will you take a look again?

WangXiangUSTC:
@lichunzhu PTAL again, I refine the code in 736fe17

lichunzhu:
@WangXiangUSTC I'm taking a look.

Issue #564, opened 2020-03-20
HA: "cannot track DDL * Unknown column 'c' in 'sbtest'"


csuzhangxc:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do? If possible, provide a recipe for reproducing the error.

    `alter table sbtest modify c char(250)` in the upstream.

2. What did you expect to see?

    replication works as normal.

3. What did you see instead?

    `[2020/03/19 17:37:33.610 +08:00] [ERROR] [syncer.go:1917] ["cannot track DDL"] [task=test_dm] [unit="binlog replication"] [schema=my_ha] [statement="ALTER TABLE `my_ha`.`sbtest` MODIFY COLUMN `c` CHAR(250)"] [location="position: (binlog.000084, 24259928), gtid-set: 0-2704460194-263588123"] [error="[schema:1054]Unknown column 'c' in 'sbtest'"]`

4. Versions of the cluster

    - DM version (run `dmctl -V` or `dm-worker -V` or `dm-master -V`):

        ```
        DM nightly before March 20.
        ```

csuzhangxc:
fixed in https://github.com/pingcap/dm/pull/574

PR #574, opened 2020-04-01
syncer/: fix track ddl

lichunzhu:
1. https://github.com/pingcap/dm/issues/564
When `needShardingHandle == false` we also should to track ddl info in tracker. If not, checkpoint will flush empty track ddl info into mysql. If syncer is restarted again, it will load empty track info from checkpoint but won't seek ddl info from downstream any more. And then track ddl info will be invalid. (supposed track ddl info: the created table info, but is empty)
2. dm-master will meet data race when join a cluster because of the misuse of grpc logger.

1. track ddl info when `needShardingHandle == false`.
2. Refine integration_test `sharding2` to test this bug. Master branch can't pass new `sharding2` test but this PR is OK.
3. Set grpc logger before start etcd to avoid data race.
4. Add integration tests in `ha` to check whether joined master can work correctly.

### Check List 

Tests 

 - Unit test
 - Integration test

Related changes


WangXiangUSTC:
> LGTM
ignore it, this pr is in wip

PR #578, opened 2020-04-03
.*/: fix dm unit tests and integration tests

lichunzhu:
This PR aims to fix wrong and unstable tests.
1. Interationt tests `relay_interrupt`, `start_task` don't clear database `dm_meta` and relative database before each subtest. So the test result except the first one is meaningless,

1. clear database `dm_meta` and relative database before each subtest. Use different table name for each subtest.

### Check List 

Tests 

 - Unit test
 - Integration test

Related changes


loader/loader.go
```diff
@@ -358,6 +358,8 @@ type Loader struct {
 
 	// for every worker goroutine, not for every data file
 	workerWg *sync.WaitGroup
+	// for other goroutines
+	wg *sync.WaitGroup
```
csuzhangxc, line 362: I prefer to use `sync.WaitGroup` rather than `*sync.WaitGroup` for single `Loader` instance (then we can remove L392, do not need to keep uniform for old ugly code)

loader/loader.go
```diff
@@ -620,6 +627,7 @@ func (l *Loader) Close() {
 
 	l.stopLoad()
 
+	l.wg.Wait()
```
csuzhangxc, line 630: Do we need to put this `l.wg.Wait()` into `l.stopLoad()` just as we did for `l.workerWg.Wait()`?

loader/loader.go
lichunzhu: addressed in https://github.com/pingcap/dm/pull/578/commits/02afc7ebfad754598a6266bdcc1acd1056f6c3a3

loader/loader.go
lichunzhu: addressed in https://github.com/pingcap/dm/pull/578/commits/02afc7ebfad754598a6266bdcc1acd1056f6c3a3

Issue #514, opened 2020-03-04
binlog replication should fluch checkpoint before executing DDL


csuzhangxc:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do? If possible, provide a recipe for reproducing the error.

    1. execute DML
    2. execute DDL & interrupt the process. (like restart the process or schedule to another process)

2. What did you expect to see?

   The task should work correctly after restarted.

3. What did you see instead?

   may trigger error of `prune DML columns and data mismatch in length`.

4. Versions of the cluster

    - DM version (run `dmctl -V` or `dm-worker -V` or `dm-master -V`):

        ```
        v1.0.3
        ```

csuzhangxc:
already fixed in #586.

PR #581, opened 2020-04-05
tests/: add ha integration test cases

lichunzhu:
### What problem does this PR solve? 
Add more dm integration_test cases.

### What is changed and how it works?
Add 9 integration tests in ha_cases.
1. For `test_kill_master_in_sync` and `test_kill_worker_in_sync` we load dmls every 0.01s to keep worker syncing.
2. For `test_isolate_master`, we use failpoint to mock master is blocked and can't do etcd election.

### Check List 

Tests 

 - Unit test
 - Integration test

PR #587, opened 2020-04-09
syncer/: fix flush checkpoint for ddls for situations that IsSharding is false

lichunzhu:
https://github.com/pingcap/dm/issues/514
* When `IsSharding` is false, table checkpoint which contains schema tracker info was not updated in `dm_meta` instantly. But global checkpoint is updated. If worker fails now, it will cause a mismatch between tracker and global checkpoint.
* Fix a dead lock bug.

Update table point when we update global checkpoint.

### Check List 

Tests 

 - Unit test
 - Integration test

Related changes


lichunzhu:
Should be merged after https://github.com/pingcap/dm/pull/581/files, add uncomment the blocked integration tests

syncer/syncer.go
```diff
@@ -742,15 +742,19 @@ func (s *Syncer) addJob(job *job) error {
...
 		}
 		// reset sharding group after checkpoint saved
 		s.resetShardingGroup(job.targetSchema, job.targetTable)
 	case insert, update, del:
 		// save job's current pos for DML events
-		if len(job.sourceSchema) > 0 {
-			s.saveTablePoint(job.sourceSchema, job.sourceTable, job.currentLocation)
+		for i := range job.sourceSchemas {
+			if len(job.sourceSchemas[i]) > 0 {
+				s.saveTablePoint(job.sourceSchemas[i], job.sourceTables[i], job.location)
```
csuzhangxc, line 756: Should we use `job. location` or `job.currentLocation` here?

csuzhangxc:
in #514, I think we should also **flush** (not only **save**) checkpoint **before** executing DDL to solve a case:

1. `INSET TABLE tbl (c1) VALUES (1)`
2. `ALTER TABLE tbl ADD COLUMN (c2 INT)`
3. flush checkpoint into downstream TiDB

interrupted after step.2 and before step.3, and then restarted from step.1, this will trigger an error with `columns and data mismatch in length: 2 (columns) 1 (data)`

syncer/syncer.go
lichunzhu: updated in https://github.com/pingcap/dm/pull/587/commits/f1837f8d8dd5d9aba2c9442fc5f5f9326745a019

lichunzhu:
> in #514, I think we should also **flush** (not only **save**) checkpoint **before** executing (and track) DDL to solve a case:
> 
> 1. `INSET TABLE tbl (c1) VALUES (1)`
> 2. `ALTER TABLE tbl ADD COLUMN (c2 INT)`
> 3. flush checkpoint into downstream TiDB
> 
> interrupted after step.2 and before step.3, and then restarted from step.1, this will trigger an error with `columns and data mismatch in length: 2 (columns) 1 (data)`

updated in https://github.com/pingcap/dm/pull/587/commits/f1837f8d8dd5d9aba2c9442fc5f5f9326745a019. A test is also added in this commit.

syncer/optimist.go
```diff
@@ -78,6 +78,10 @@ func (s *Syncer) handleQueryEventOptimistic(
 		}
 	}
 
+	if err := s.flushCheckPoints(); err != nil {
```
csuzhangxc, line 81: Can we put it near `s.jobWg.Wait()` to make it more meaningful?

syncer/syncer.go
```diff
@@ -753,11 +753,16 @@ func (s *Syncer) addJob(job *job) error {
 		// save job's current pos for DML events
 		for i := range job.sourceSchemas {
 			if len(job.sourceSchemas[i]) > 0 {
-				s.saveTablePoint(job.sourceSchemas[i], job.sourceTables[i], job.location)
+				s.saveTablePoint(job.sourceSchemas[i], job.sourceTables[i], job.currentLocation)
 			}
 		}
 	}
 
+	failpoint.Inject("FlushCheckpointExit", func() {
```
csuzhangxc, line 761: in many places, we call `s.flushCheckPoints` directly, but they will not trigger this `FlushCheckpointExit`, how about giving it another name or move it into `s.flushCheckPoints`?

syncer/syncer.go
```diff
@@ -1706,6 +1711,10 @@ func (s *Syncer) handleQueryEvent(ev *replication.QueryEvent, ec eventContext) e
 			needHandleDDLs = appliedSQLs // maybe nil
 		}
 
+		if err := s.flushCheckPoints(); err != nil {
```
csuzhangxc, line 1714: If we do not dispatch a `flush` job, then we must call `s.jobWg.Wait()` before `s.flushCheckPoints()`.

(Is dispatch a `flush` job better?)

tests/sharding2/run.sh
```diff
@@ -46,6 +46,40 @@ function run() {
     # use sync_diff_inspector to check data now!
     echo "check sync diff for the increment replication"
     check_sync_diff $WORK_DIR $cur/conf/diff_config.toml
+
+    echo "kill dm-worker1"
+    ps aux | grep dm-worker1 |awk '{print $2}'|xargs kill || true
+    echo "kill dm-worker2"
+    ps aux | grep dm-worker2 |awk '{print $2}'|xargs kill || true
```
csuzhangxc, line 53: Do we need to `check_port_offline` here?

syncer/syncer.go
```diff
@@ -1799,7 +1798,7 @@ func (s *Syncer) handleQueryEvent(ev *replication.QueryEvent, ec eventContext) e
 
 	s.tctx.L().Info(annotate, zap.String("event", "query"), zap.String("source", source), zap.Strings("ddls", needHandleDDLs), zap.ByteString("raw statement", ev.Query), zap.Bool("in-sharding", needShardingHandle), zap.Stringer("start location", startLocation), zap.Bool("is-synced", synced), zap.Int("unsynced", remain))
 
-	if err := s.flushCheckPoints(); err != nil {
+	if err := s.flushJobs(); err != nil {
```
csuzhangxc, line 1801: It seems we can remove `s.jobWg.Wait()` in L1856 now? (or even we can remove `s.jobWg.Wait()` in L723? but we may still just keep it, the shard mechanism make this not uniform, 😢 )

syncer/job.go
```diff
@@ -58,8 +58,8 @@ func (t opType) String() string {
 
 type job struct {
 	tp              opType
-	sourceSchema    string
-	sourceTable     string
+	sourceSchemas   []string
+	sourceTables    []string
```
WangXiangUSTC, line 62: why job need to use multiple schemas and tables? can you add a comment

syncer/job.go
```diff
@@ -111,10 +112,21 @@ func newDDLJob(ddlInfo *shardingDDLInfo, ddls []string, location, cmdLocation bi
 	}
 
 	if ddlInfo != nil {
-		j.sourceSchema = ddlInfo.tableNames[0][0].Schema
-		j.sourceTable = ddlInfo.tableNames[0][0].Name
+		j.sourceSchemas = []string{ddlInfo.tableNames[0][0].Schema}
+		j.sourceTables = []string{ddlInfo.tableNames[0][0].Name}
 		j.targetSchema = ddlInfo.tableNames[1][0].Schema
 		j.targetTable = ddlInfo.tableNames[1][0].Name
+	} else if sourceTbls != nil {
```
WangXiangUSTC, line 119: in which case `ddlInfo` is nil and `sourceTbls` is not nil? can you add a example

syncer/job.go
WangXiangUSTC: and I think map is better

syncer/job.go
lichunzhu: updated in https://github.com/pingcap/dm/pull/587/commits/2733185c2787894f9104e20aab935ac154092e4b

syncer/job.go
lichunzhu: all updated in https://github.com/pingcap/dm/pull/587/commits/2733185c2787894f9104e20aab935ac154092e4b

Issue #576, opened 2020-04-02
stop-task may be blocked when sync is wating for relay to catchup


csuzhangxc:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do? If possible, provide a recipe for reproducing the error.

run `stop-task`

2. What did you expect to see?

the task should be stopped.

3. What did you see instead?

the task is still blocked in `unitTransWaitCondition`.

4. Versions of the cluster

    - DM version (run `dmctl -V` or `dm-worker -V` or `dm-master -V`):

        ```
        DM v1.0.4
        ```

csuzhangxc:
should already be fixed in #589.

PR #589, opened 2020-04-09
worker/: fix worker stuck in unitTransWaitCondition

lichunzhu:
https://github.com/pingcap/dm/issues/576
When subtask is started, it will use `unitTransWaitCondition` to wait for relay log to catch up. But if this task is paused or stopped, we should quit as fast as we can instead of keeping waiting.

1. Use `st.currCtx` to control wait of `unitTransWaitCondition`.
2. Use parent context `st.ctx` and son context `st.currCtx` to make sure task can be shut down.

### Check List 

Tests 

 - Unit test
 - Integration test

Related changes


dm/worker/subtask.go
```diff
@@ -73,9 +73,13 @@ type SubTask struct {
...
 
 	sync.RWMutex
-	wg     sync.WaitGroup
+	wg sync.WaitGroup
+	// ctx is used for the whole subtask. It will be created only when we new a subtask.
 	ctx    context.Context
 	cancel context.CancelFunc
+	// currCtx is used for one loop. It will be created each time we use st.run/st.Resume
+	currCtx    context.Context
+	currCancel context.CancelFunc
```
csuzhangxc, line 82: Can we do not keep these two fields in the `struct` (to make it clear)?

lichunzhu: Do you mean use one context?

csuzhangxc: pass the local variable between methods (it seems not enough?)

lichunzhu: How to pass local variable between methods?

dm/worker/subtask.go
```diff
@@ -186,36 +193,59 @@ func (st *SubTask) Run() {
...
+	st.setStage(pb.Stage_Running)
+	ctx, cancel := context.WithCancel(st.ctx)
+	st.setCurrCtx(ctx, cancel)
+	err := st.unitTransWaitCondition(ctx)
 	if err != nil {
 		st.l.Error("wait condition", log.ShortError(err))
 		st.fail(err)
 		return
+	} else if ctx.Err() != nil {
+		return
```
csuzhangxc, line 205: Will the stage keep as `Stage_Running`?

dm/worker/subtask.go
```diff
@@ -73,9 +73,13 @@ type SubTask struct {
 	l log.Logger
 
 	sync.RWMutex
-	wg     sync.WaitGroup
+	wg sync.WaitGroup
+	// ctx is used for the whole subtask. It will be created only when we new a subtask.
 	ctx    context.Context
 	cancel context.CancelFunc
+	// currCtx is used for one loop. It will be created each time we use st.run/st.Resume
+	currCtx    context.Context
```
csuzhangxc, line 81: Is it enough to just let `ctx` as `used for one loop`

dm/worker/subtask_test.go
```diff
@@ -479,3 +485,62 @@ func (t *testSubTask) TestSubtaskWithStage(c *C) {
 	c.Assert(st.CurrUnit(), Equals, nil)
 	c.Assert(st.Result(), IsNil)
 }
+
+func (t *testSubTask) TestSubtaskFastQuit(c *C) {
+	// case: test subtask stuck into unitTransWaitCondition
+	cfg := &config.SubTaskConfig{
+		Name: "testSubtaskFastQuit",
+		Mode: config.ModeFull,
```
csuzhangxc, line 493: Why not `All`?

dm/worker/subtask_test.go
```diff
@@ -479,3 +485,62 @@ func (t *testSubTask) TestSubtaskWithStage(c *C) {
...
+	mockLoader := NewMockUnit(pb.UnitType_Load)
+	mockSyncer := NewMockUnit(pb.UnitType_Sync)
+
+	st := NewSubTaskWithStage(cfg, pb.Stage_Paused, nil)
+	st.prevUnit = mockLoader
+	st.currUnit = mockSyncer
+
+	finished := make(chan struct{})
+	go func() {
+		close(finished)
```
csuzhangxc, line 515: Close directly or do you mean `defer`?

dm/worker/subtask.go
lichunzhu: I don't think it's enough. Suggest a situation:
1. we close a subtask and remove subtask from subtask holder.
2. After this subtask's cancel was runned, a goroutine build a new ctx and cancel and start `subtask.Run`.
If this worker is not stopped, this subtask will keep running without monitor. It's dangerous.

dm/worker/subtask.go
lichunzhu: `ctx.Err() != nil` means this context is canceled in other go routine, that go routine will change the stage.

csuzhangxc: So, how about checking the stage in the unit test cases?

csuzhangxc:
@lichunzhu please do not forget to cherry-pick it manually.

lichunzhu:
> @lichunzhu please do not forget to cherry-pick it manually.

@csuzhangxc OK

PR #590, opened 2020-04-09
syncer: add more metrics; refine batch handle

csuzhangxc:

add more metrics for binlog replication unit.

- add more metrics:
    - `binlogReadDurationHistogram`
    - `binlogEventSizeHistogram`
    - `conflictDetectDurationHistogram`
    - `queueSizeGauge`
    - `stmtHistogram`
- refine batch handle (use `time.After` to replcate `time.Sleep`)
- update Grafana dashboard

image

image

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

### DIFF
--- a/syncer/syncer.go
+++ b/syncer/syncer.go
@@ -998,6 +998,7 @@ func (s *Syncer) sync(tctx *tcontext.Context, queueBucket string, db *DBConn, jo
 	for {
 		select {
 		case sqlJob, ok := <-jobChan:
+			queueSizeGauge.WithLabelValues(s.cfg.Name, queueBucket).Set(float64(len(jobChan)))
 			if !ok {
 				return
 			}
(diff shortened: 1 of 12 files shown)


csuzhangxc:
@GregoryIan PTAL again

syncer/syncer.go
```diff
@@ -998,6 +998,7 @@ func (s *Syncer) sync(tctx *tcontext.Context, queueBucket string, db *DBConn, jo
 	for {
 		select {
 		case sqlJob, ok := <-jobChan:
+			queueSizeGauge.WithLabelValues(s.cfg.Name, queueBucket).Set(float64(len(jobChan)))
```
GregoryIan, line 1001: multiple queues share same bucket, it would be problem i under some cases. 
It is recommended that both slow query
 threshold and bucket count be configurable, and these can be optimized later

PR #594, opened 2020-04-10
syncer: improve performance

csuzhangxc:

- add metrics to diagnose perf issue
- improve perf

- add some metrics 
- increase job queue length
- use `strings.Builder` to replace some `fmt.Sprintf`
    - `strings.Builder`/`bytes.Buffer` with `sync.Pool` doesn't achieve better perf
- remove some unnecessary slice create
    - we can do more here, but it will make the code ugly and may not improve a lot, so I revert it

a sysbench workload in my laptop with the black hole downstream

`transform binlog event duration` (.90):

**before this PR: 143us --> after this PR: 72us**

`binlog event QPS`:

**before this PR: ~4k --> after this PR: ~6k**

**before this PR:**
image

**after this PR:**
image

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

 - Increased code complexity

Related changes

### DIFF
--- a/dm/config/task.go
+++ b/dm/config/task.go
@@ -55,10 +55,9 @@ var (
 	defaultPoolSize = 16
 	defaultDir      = "./dumped_data"
 	// SyncerConfig
-	defaultWorkerCount             = 16
-	defaultBatch                   = 100
-	defaultQueueSize               = 5120
-	defaultCheckpointFlushInterval = 30
+	defaultWorkerCount = 16
+	defaultBatch       = 100
+	defaultQueueSize   = 5120
 )
 
 // Meta represents binlog's meta pos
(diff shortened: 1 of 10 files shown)


csuzhangxc:
@GregoryIan 
we have a bug before https://github.com/pingcap/dm/pull/594/commits/107ef78d3d8a73308ec22635d73e492d227ad12e for  `transform binlog event duration` with `prometheus.ExponentialBuckets(0.0005, 2, 18)`, it will have a duration always > 450us.

In this PR, for a workload in my laptop the real `transform binlog event duration` decreases from ~310us to ~145us.

dm/config/task.go
```diff
@@ -57,6 +57,7 @@ var (
 	// SyncerConfig
 	defaultWorkerCount = 16
 	defaultBatch       = 100
+	defaultQueueSize   = 5120
```
WangXiangUSTC, line 60: 5120 seems too big, maybe cause OOM？

PR #597, opened 2020-04-13
syncer: improve performance - part 2

csuzhangxc:

improve binlog replication performance in some cases.

- do not use `null` value as the concurrence dispatch key
- decrease  default value of `queue-size` from `5120` to `1024`
- add `checkpoint-flush-interval` config item
- update `binlogSkippedEventsTotal` to `skipBinlogDurationHistogram` to support observe duration and update the Grafana dashboard

image

image

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

### DIFF
--- a/syncer/dml.go
+++ b/syncer/dml.go
@@ -130,7 +130,7 @@ func genInsertSQLs(param *genDMLParam) ([]string, [][]string, [][]interface{}, e
 			originalValue = extractValueFromData(originalDataSeq[dataIdx], originalColumns)
 		}
 
-		ks := genMultipleKeys(originalValue, originalIndexColumns, fullname)
+		ks := genMultipleKeys(originalValue, originalIndexColumns, originalColumns, fullname)
 		sqls = append(sqls, sql)
 		values = append(values, value)
 		keys = append(keys, ks)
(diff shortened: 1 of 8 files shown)


csuzhangxc:

syncer/dml.go
```diff
@@ -443,20 +444,42 @@ func findColumns(columns []*column, indexColumns map[string][]string) map[string
...
+		}
 		// for key, I think no need to add the `,` separator.
 		buf.WriteString(columnValue(data, columns[i].unsigned, columns[i].tp))
 	}
+	if buf.Len() == 0 {
+		log.L().Debug("all value are nil, no key generated", zap.String("table", table))
+		return "" // all values are `null`.
+	}
+
+	buf.WriteString(table)
```
WangXiangUSTC, line 460: why write table at the end?

syncer/syncer.go
```diff
@@ -1454,7 +1455,7 @@ func (s *Syncer) handleRowsEvent(ev *replication.RowsEvent, ec eventContext) err
 		return err
 	}
 	if ignore {
-		binlogSkippedEventsTotal.WithLabelValues("rows", s.cfg.Name).Inc()
+		skipBinlogDurationHistogram.WithLabelValues("rows", s.cfg.Name).Observe(time.Since(ec.startTime).Seconds())
```
WangXiangUSTC, line 1458: what is this metirc used for?  it may cost a lot of time in this function?

csuzhangxc: some users may need to ignore most of the data, you can take a look at our internal ONCALL-1243 issue.

syncer/dml.go
csuzhangxc: if all columns are `null`, the return `""` (without `table`).

csuzhangxc:
> seems that we have too many debug log, if we are debug other issue and need to enable debug log, these logs may disturb us.

em... but these logs are still needed for _debug_ (or just let them as the placeholder). maybe when debugging, we can choose to?

1. change some of them to _info_ or higher
2. do not run with _debug_ too long

WangXiangUSTC:
> > seems that we have too many debug log, if we are debug other issue and need to enable debug log, these logs may disturb us.
> 
> em... but these logs are still needed for _debug_ (or just let them as the placeholder). maybe when debugging, we can choose to?
> 
> 1. change some of them to _info_ or higher
> 2. do not run with _debug_ too long

OK，LGTM

PR #609, opened 2020-04-15
*: fix metrics

csuzhangxc:

fix two issues for metrics:
- add `task` filter for `binlog event size`.
- register `binlogTransformDurationHistogram` for relay.

as above.

### Check List 

Tests 

 - Unit test
 - Integration test


csuzhangxc:

Issue #584, opened 2020-04-08
binlog replication: wrong location compare if not enable GTID


csuzhangxc:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do? If possible, provide a recipe for reproducing the error.

   two MySQL instances in upstream with optimistic mode, add a column with different default value in upstream.

    for MySQL-1.
    ```sql
    alter table tbl add column c2 varchar(100) default '1234';
    ```

    for MySQL-2.
    ```sql
    alter table tbl add column c2 varchar(100) default '5678';
    ```

2. What did you expect to see?

    block DDL for the table in MySQL-2 even auto resume trigger.

3. What did you see instead?

    DDL for the table in MySQL-2 ignored after auto resume.

4. Versions of the cluster

    - DM version (run `dmctl -V` or `dm-worker -V` or `dm-master -V`):

        ```
        nightly before April 8.
        ```

```sql
| Log_name         | Pos  | Event_type     | Server_id | End_log_pos | Info                                                                |
+------------------+------+----------------+-----------+-------------+---------------------------------------------------------------------+
| mysql-bin.000003 | 4    | Format_desc    | 12        | 123         | Server ver: 5.7.22-log, Binlog ver: 4                               |
| mysql-bin.000003 | 123  | Previous_gtids | 12        | 194         | 11ee15b1-796d-11ea-85ec-0242ac130003:1-5                            |
| mysql-bin.000003 | 194  | Gtid           | 12        | 259         | SET @@SESSION.GTID_NEXT= '11ee15b1-796d-11ea-85ec-0242ac130003:6'   |
| mysql-bin.000003 | 259  | Query          | 12        | 347         | create database db                                                  |
| mysql-bin.000003 | 347  | Gtid           | 12        | 412         | SET @@SESSION.GTID_NEXT= '11ee15b1-796d-11ea-85ec-0242ac130003:7'   |
| mysql-bin.000003 | 412  | Query          | 12        | 519         | use `db`; create table tbl (c1 int primary key)                     |
| mysql-bin.000003 | 519  | Gtid           | 12        | 584         | SET @@SESSION.GTID_NEXT= '11ee15b1-796d-11ea-85ec-0242ac130003:8'   |
| mysql-bin.000003 | 584  | Query          | 12        | 654         | BEGIN                                                               |
| mysql-bin.000003 | 654  | Table_map      | 12        | 698         | table_id: 110 (db.tbl)                                              |
[...]

PR #614, opened 2020-04-17
syncer: minor fix on location

WangXiangUSTC:

1. fix issue https://github.com/pingcap/dm/issues/584
2. when the upstream switch the master and slave, the gtid set is same, but the binlog position will changed. The `CompareLocation` will compare by gtid set at the first.  If gtid set is equal, will then compare by position, it may cause some problems.

1.  the https://github.com/pingcap/dm/issues/584 is caused by seting `b.flushedLocation = b.location`  in flush, should use b.location.Clone().

2. compare location by gtid or position according to the config's `EnableGTID`.

### Check List 

Tests 

 - Unit test
 - Manual test


pkg/binlog/position.go
```diff
@@ -226,25 +226,43 @@ func (l Location) CloneWithFlavor(flavor string) Location {
...
+}
+
+// CompareGTID returns:
+//   1, true if gSet1 is bigger than gSet2
+//   0, true if gSet1 is equal to gSet2
+//   -1, true if gSet1 is less than gSet2
+// but if can't compare gSet1 and gSet2, will returns 0, false
+func CompareGTID(gSet1, gSet2 gtid.Set) (int, bool) {
+	if gSet1 != nil && len(gSet1.String()) != 0 &&
+		gSet2 != nil && len(gSet2.String()) != 0 {
```
csuzhangxc, line 250: Do we need to support compare a non-empty GTID Set with an empty GTID set? (may non-empty > empty? as we did in `gtid.Set.Contain`)

syncer/checkpoint.go
```diff
@@ -375,8 +378,12 @@ func (cp *RemoteCheckPoint) DeleteSchemaPoint(tctx *tcontext.Context, sourceSche
 	return nil
 }
 
-// IsNewerTablePoint implements CheckPoint.IsNewerTablePoint
-func (cp *RemoteCheckPoint) IsNewerTablePoint(sourceSchema, sourceTable string, location binlog.Location) bool {
+// IsNewerTablePoint implements CheckPoint.IsNewerTablePoint.
+// gte means greater than or equal, gte should judge by EnableGTID and the event type
+// - when enable GTID and binlog is DML, different location may have same GTID, so when GTID is equal, can treat it as new table point
+// - when enable GTID and binlog is DDL, different DDL have different GTID, so if GTID is euqal, it is a old table point
+// - when not enable GTID, just compare the position, and only when grater than the old point is newer table point
```
lichunzhu, line 385: Because DML doesn't update table checkpoint?

pkg/binlog/position.go
WangXiangUSTC: I support it in f06dcd4

syncer/checkpoint.go
WangXiangUSTC: no, because go-mysql update GTID set of rows event in XID event after this rows event. I update the comment and add a example.

pkg/binlog/position_test.go
```diff
@@ -373,22 +374,9 @@ func (t *testPositionSuite) TestCompareCompareLocation(c *C) {
...
-			gmysql.Position{
-				Name: "binlog.00001",
-				Pos:  123,
-			},
-			"",
-			gmysql.Position{
-				Name: "binlog.00001",
-				Pos:  123,
-			},
-			"",
```
lichunzhu, line 388: Why do we delete this?

WangXiangUSTC: I think it is duplicate with the test case above, but in fact they have different flavor. I add it back in 45c9f69

Issue #620, opened 2020-04-20
Can optimistic shard DDL replicate `DROP COLUMN` to downstream until DDLs received from all shards


csuzhangxc:
## Question

Now, in optimistic shard DDL mode, DM will replicate `DROP COLUMN` to downstream once received DDLs from any shard table.

Can it only replicate `DROP COLUMN` to downstream until DDLs received from all shard tables?

csuzhangxc:
No, we need to replicate `DROP INDEX` to downstream directly to make the schema more compatible.

Issue #613, opened 2020-04-16
optimistic shard DDL do not support `ADD COLUMN` and `DROP COLUMN` in one statement well


csuzhangxc:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do? If possible, provide a recipe for reproducing the error.

    execute one statement with both `ADD COLUMN` and `DROP COLUMN` in optimistic shard DDL mode

2. What did you expect to see?

    replication works normally.

3. What did you see instead?

    `ADD COLUMN` and `DROP COLUMN` replicated to the downstream directly, the DML from other tables can not replicated because miss column.

4. Versions of the cluster

    - DM version (run `dmctl -V` or `dm-worker -V` or `dm-master -V`):

        ```
        DM before git commit 5e3aaba3a1e9b186a3c13b00f8e7a0f472d177f0.
        ```

csuzhangxc:
it will not be supported recently, instead to report an error in #656.

lance6716:
maybe we should send split DDL and its TableInfo one-by-one

lance6716:
closed by https://github.com/pingcap/dm/pull/1479

PR #621, opened 2020-04-20
shardddl: add test cases for two scenarios

csuzhangxc:

add two unit test cases for:
 
- #613 
- #620 

- for #613, we don't support it well because we lack the schema diff mechanism.
- fro #620, we need to replicate `DROP INDEX` to downstream directly to make the schema more compatible.

### Check List 

Tests 

 - Unit test

Related changes


csuzhangxc:

PR #622, opened 2020-04-21
worker/: fix worker stuck in unitTransWaitCondition(#589)

lichunzhu:
cherry-pick #589

---

https://github.com/pingcap/dm/issues/576
When subtask is started, it will use `unitTransWaitCondition` to wait for relay log to catch up. But if this task is paused or stopped, we should quit as fast as we can instead of keeping waiting.

1. Use `st.currCtx` to control wait of `unitTransWaitCondition`.
2. Use parent context `st.ctx` and son context `st.currCtx` to make sure task can be shut down.

### Check List 

Tests 

 - Unit test
 - Integration test

Related changes


WangXiangUSTC:
LGTM

PR #635, opened 2020-04-26
sync unit: disable retry for binlog syncer

WangXiangUSTC:

when enabling retry, go-mysql will [retry sync](https://github.com/siddontang/go-mysql/blob/3731d5147201d61e212e7fbd0c83ea14cb6ef0b9/replication/binlogsyncer.go#L577) from the previous GTID set in GTID mode, 
which may get duplicate binlog event after retry success. 

so just disable retry, and the task will exit with error, then auto-resume by DM itself.

### Check List 

Tests 

 - Manual test

### DIFF
--- a/pkg/binlog/common/replication.go
+++ b/pkg/binlog/common/replication.go
@@ -20,22 +20,18 @@ import (
 )
 
 var (
-	// max reconnection times for binlog syncer in go-mysql
-	maxBinlogSyncerReconnect = 60
+	// MaxBinlogSyncerReconnect is the max reconnection times for binlog syncer in go-mysql
+	MaxBinlogSyncerReconnect = 60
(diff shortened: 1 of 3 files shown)


pkg/binlog/common/replication.go
```diff
@@ -28,10 +28,14 @@ var (
 )
 
 // SetDefaultReplicationCfg sets some default value for BinlogSyncerConfig
-func SetDefaultReplicationCfg(cfg *replication.BinlogSyncerConfig) {
+func SetDefaultReplicationCfg(cfg *replication.BinlogSyncerConfig, reConnect bool) {
```
csuzhangxc, line 31: How about replcaing `reConnect` with `RetryCount`?

rest LGTM.

WangXiangUSTC: update in e856f66

Issue #637, opened 2020-04-27
syncer: data race for `collectMetrics`


csuzhangxc:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do? If possible, provide a recipe for reproducing the error.

    run CI

2. What did you expect to see?

    passed as normal.

3. What did you see instead?

    data race detected, see https://internal.pingcap.net/idc-jenkins/blue/rest/organizations/jenkins/pipelines/dm_ghpr_test/runs/4757/nodes/73/log/?start=0.

    this should be caused by https://github.com/pingcap/dm/issues/629.

4. Versions of the cluster

    - DM version (run `dmctl -V` or `dm-worker -V` or `dm-master -V`):

        ```
        https://github.com/pingcap/dm/pull/615
        ```

WangXiangUSTC:
fix in https://github.com/pingcap/dm/pull/644

Issue #629, opened 2020-04-23
`resume-task` may resume a not real paused (still running) task


csuzhangxc:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do? If possible, provide a recipe for reproducing the error.

   `pause-task` then `resume-task`

2. What did you expect to see?

    the task paused and then resumed.

3. What did you see instead?

   some errors may appear, like (in stdout)

    ```
    [2020-04-22T07:15:32.788Z] [2020/04/22 15:15:10.105 +08:00] [INFO] [ddl.go:457] ["[ddl] start DDL job"] [job="ID:69, Type:drop table, State:none, SchemaState:none, SchemaID:46, TableID:59, RowCount:0, ArgLen:0, start time: 2020-04-22 15:15:10.098 +0800 CST, Err:<nil>, ErrCount:0, SnapshotVersion:0"] [query="CREATE TABLE IF NOT EXISTS `sharding1`.`t1` (`id` BIGINT(20) NOT NULL AUTO_INCREMENT,`uid` INT(11) DEFAULT NULL,`name` VARCHAR(80) DEFAULT NULL,`info` VARCHAR(100) DEFAULT NULL,`age` INT(11) DEFAULT NULL,PRIMARY KEY(`id`),UNIQUE `uid`(`uid`)) ENGINE = InnoDB DEFAULT CHARACTER SET = UTF8MB4 DEFAULT COLLATE = UTF8MB4_BIN AUTO_INCREMENT = 1157477880792380722"]
[...]

PR #644, opened 2020-04-29
.*: fix bug that after execute pause-task the task may still running 

WangXiangUSTC:

fix issue https://github.com/pingcap/dm/issues/637 and https://github.com/pingcap/dm/issues/629

when executing pause task the task is not real paused, and then execute resume task, maybe the old unit is still process, and then will cause some problem.

1. add two stage `pausing` and `resuming`:
- when pausing task, first the stage is `pausing`, and update stage to `paused` at the last.
- when resuming task, first the stage is `resuming`, and update stage to `running` at the last.

2. a minor fix, `fetchResult` should wait for the unit's process result.

### Check List 

Tests 

 - Integration test


dm/worker/subtask.go
```diff
@@ -236,18 +236,16 @@ func (st *SubTask) callCurrCancel() {
 func (st *SubTask) fetchResult(pr chan pb.ProcessResult) {
 	defer st.wg.Done()
 
-	st.RLock()
-	ctx := st.currCtx
-	st.RUnlock()
-
 	select {
-	case <-ctx.Done():
+	case <-st.ctx.Done():
```
csuzhangxc, line 240: Should we need to fix this in release-1.0? cc @lichunzhu

dm/unit/unit.go
```diff
@@ -77,3 +78,11 @@ func NewProcessError(err error) *pb.ProcessError {
 	}
 	return result
 }
+
+// IsCtxCanceledProcessErr returns true if the err's
```
csuzhangxc, line 82: Do some comments lost at the end?

dm/worker/subtask.go
lichunzhu: I think it's not correct. We will start multi `fetchResult` in the end. `fetchResult` should be ended when one turn is finished.

csuzhangxc: but we still waiting for `case result := <-pr:`, is that enough? @lichunzhu

dm/worker/subtask.go
```diff
@@ -462,6 +472,7 @@ func (st *SubTask) Resume() error {
 		st.setStage(pb.Stage_Paused)
 		return err
 	} else if ctx.Err() != nil {
+		st.setStage(pb.Stage_Paused)
```
lichunzhu, line 475: I think this should not be set to `pb.Stage_Paused`, see https://github.com/pingcap/dm/pull/589#discussion_r409342791

dm/worker/subtask.go
lichunzhu: It's enough, and we mush assure `pr` will get a result in the future.

dm/worker/subtask.go
WangXiangUSTC: ok, I remove this and add a comment e34ec49

tests/sharding/run.sh
```diff
@@ -90,12 +90,21 @@ function run() {
...
     # wait really paused
-    # FIXME: `if !st.stageCAS(pb.Stage_Running, pb.Stage_Paused)` in `subtask.go` is not enough to indicate the real stage.
-    sleep 2
+    run_dm_ctl_with_retry $WORK_DIR "127.0.0.1:$MASTER_PORT" \
+        "query-status test" \
+        "Paused" 2
+    run_dm_ctl_with_retry $WORK_DIR "127.0.0.1:$MASTER_PORT" \
+        "resume-task test"\
+        "\"result\": true" 3
+    # resume twice
```
lichunzhu, line 107: Should we add a test to test `stop-task` twice?

WangXiangUSTC: add in ecd45e6

PR #650, opened 2020-05-07
.*: remove mydumper binary config

GMHDBJD:
- remove mydumper-path in task config
- remove mydumper check in Makefile
- simply rename `tests/retry_cancel/run.sh` to `tests/retry_cancel/run.sh.todo` to bypass test in `make integration_test`

### Notice
- `MydumperPath` in source code, `mydumper binary` in Dockerfile and ansible config have no changed

### Check List 

Tests 

 - Unit test
 - Integration test


GMHDBJD:
> Should we also update pingcap/docs-cn?
Yes, pr has been committed

PR #655, opened 2020-05-11
binlog replication: add config DisableRetrySync for binlog syncer

WangXiangUSTC:

relate with https://github.com/pingcap/dm/pull/635

although set MaxReconnectAttempts = 1 for binlog syncer, binlog syncer will still retry, the  `MaxReconnectAttempts` only works when re-sync failed https://github.com/siddontang/go-mysql/blob/fba9c9655f7cff606e91f387352561ec5a77de4b/replication/binlogsyncer.go#L679.

need to set `DisableRetrySync = true`, and binlog syncer will return at https://github.com/siddontang/go-mysql/blob/fba9c9655f7cff606e91f387352561ec5a77de4b/replication/binlogsyncer.go#L665

### Check List 

Tests 

 - Manual test (switch master and slave)


pkg/binlog/common/replication.go
```diff
@@ -32,6 +32,9 @@ func SetDefaultReplicationCfg(cfg *replication.BinlogSyncerConfig, retryCount in
 	cfg.UseDecimal = true // must set true. ref: https://github.com/pingcap/tidb-enterprise-tools/pull/272
 	cfg.VerifyChecksum = true
 	cfg.MaxReconnectAttempts = retryCount
+	if retryCount <= 1 {
```
csuzhangxc, line 35: NOTE: in go-mysql, `zero or negative number means infinite retry. this configuration will not work if DisableRetrySync is true`.

WangXiangUSTC: add comment in https://github.com/pingcap/dm/pull/655/commits/c352270d08dadf560cc5b2c32a6664bcc03d4d6f

PR #656, opened 2020-05-11
shardddl: report error for conflicts between old and new table

csuzhangxc:

report error for the case mentioned in #613 

compare old table info with new table info.

### Check List 

Tests 

 - Unit test


csuzhangxc:

PR #661, opened 2020-05-13
refine dump unit log

lichunzhu:
dump unit log is confusing

refine dump unit log

### Check List 

Tests 

 - Unit test
 - Integration test


dumpling/dumpling.go
```diff
@@ -104,7 +104,11 @@ func (m *Dumpling) Process(ctx context.Context, pr chan pb.ProcessResult) {
 		errs = append(errs, unit.NewProcessError(err))
 	}
 
-	m.logger.Info("dump data finished", zap.Duration("cost time", time.Since(begin)))
+	if err == nil {
```
csuzhangxc, line 107: How about checking `errs` as we send it to `pb.ProcessResult`.

lichunzhu: addressed in https://github.com/pingcap/dm/pull/661/commits/8beb823b1ddc11c32a40345d363bf3168b2a1157

PR #663, opened 2020-05-14
update tools dependency

lichunzhu:
update tidb-tools to solve table name changing after applyOn black-white-list, details at pingcap/tidb-tools#342

change tidb-tools to latest version

### Check List 

Tests 

 - Unit test
 - Integration test


csuzhangxc:
@GMHDBJD PTAL

PR #682, opened 2020-05-25
update dm-portal default name

lichunzhu:
dm-portal default name `test-task` will be judged as invalid name.

Change `test-task` to `test`


csuzhangxc:
@lichunzhu should we need to cherry-pick it to `release-1.0`.

lichunzhu:
> @lichunzhu should we need to cherry-pick it to `release-1.0`.

Yes we should

Issue #677, opened 2020-05-21
Support setting session variables


csuzhangxc:
## Feature Request

**Is your feature request related to a problem? Please describe:**

TiDB has some special variables, like [tidb_constraint_check_in_place](https://pingcap.com/docs-cn/stable/tidb-specific-system-variables/#tidb_constraint_check_in_place), and we may need to only set them for some sessions (or tasks).

NOTE: we can also use this to support set `SQL_MODE` for the session.

**Describe the feature you'd like:**

Support setting session variables

**Describe alternatives you've considered:**

add a new config item in `DBConfig` for session variables.

**Teachability, Documentation, Adoption, Migration Strategy:**

GMHDBJD:
support in #687

PR #687, opened 2020-05-27
config: support setting session variable

GMHDBJD:
Support setting session variables [#677 ](https://github.com/pingcap/dm/issues/677)

add SessionConfig in TargetDB, set session variables in db.Apply.

Tests 

 - Unit test
 - Integration test

Code changes


pkg/conn/basedb.go
```diff
@@ -56,10 +56,16 @@ func (d *DefaultDBProviderImpl) Apply(config config.DBConfig) (*BaseDB, error) {
 		}
 		maxIdleConns = rawCfg.MaxIdleConns
 	}
+
+	for key, val := range config.Session {
+		dsn += fmt.Sprintf("&%s=\"%s\"", key, val)
```
csuzhangxc, line 61: Do we need to use different formats for types of `val`? see https://github.com/go-sql-driver/mysql#system-variables

tests/all_mode/conf/dm-task.yaml
```diff
@@ -13,6 +13,11 @@ target-database:
   port: 4000
   user: "root"
   password: ""
+  session:
+    sql_mode: "ONLY_FULL_GROUP_BY,STRICT_TRANS_TABLES,NO_ZERO_IN_DATE,NO_ZERO_DATE,ERROR_FOR_DIVISION_BY_ZERO,NO_AUTO_CREATE_USER,NO_ENGINE_SUBSTITUTION"
+    tidb_skip_utf8_check: 1
+    tidb_disable_txn_auto_retry: off
+    tidb_retry_limit: "10"
```
csuzhangxc, line 20: Is there any method to verify any or all of them have taken effect?

pkg/conn/basedb.go
GMHDBJD: Because we don't restrict user config, it's hard to use different formats between int and string.
I use `'` for all types and url.QueryEscape to escape string in 9ec51cb. MySQL extracts the relevant type from a string.

tests/all_mode/conf/dm-task.yaml
GMHDBJD: Unit test need launch tidb, so I add test in integration test in 9ec51cb.

pkg/conn/basedb.go
WangXiangUSTC: can it work if add "" for number value?

GMHDBJD: yes，see  https://github.com/pingcap/dm/blob/9ec51cb2ace7e31d8c61112134f5b0a795bf938f/tests/all_mode/conf/dm-task.yaml#L20

csuzhangxc: > MySQL extracts the relevant type from a string.

Are there any docs/links explain about this?

GMHDBJD: Seems not official. https://github.com/go-sql-driver/mysql/issues/405#issuecomment-256257156

csuzhangxc: 👍 you can add this as a comment above the code.

GMHDBJD: add commet in 5d57a5b

Issue #674, opened 2020-05-20
DM-master resign from leader when start scheduler is failed


lichunzhu:
## Feature Request

**Is your feature request related to a problem? Please describe:**

When DM-master fails to start scheduler it will still serve as leader but will refuse scheduler related service. This resign should be retried several times.

**Describe the feature you'd like:**

Retire leader directly when election is failed.

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

csuzhangxc:
one case why we needed it below.

1. 5 DM-master members, kill 2 of them.
2. DM-master-3 became the leader
3. kill DM-master-3 
    1. DM-master-3 resign from the leader
    2. DM-master-4 became the leader
    3. embed etcd in DM-master-3 shutdown
    4. DM-master-4 can't bootstrap as a leader because only 2/5 etcd members alive now

csuzhangxc:
done in #695.

Issue #527, opened 2020-03-09
Add a mechanism to transfer leader role when any necessary component fail to start


lichunzhu:
## Feature Request

**Is your feature request related to a problem? Please describe:**

https://github.com/pingcap/dm/pull/522#discussion_r389794217
When dm-master becomes leader, some components need to be started. If they fail to start, we should transfer leader role.
**Describe the feature you'd like:**

Add a mechanism to transfer leader role to another DM-master member when any necessary components (pessimist and scheduler in currenlty) started fail later.

PR #695, opened 2020-06-01
dm-master: resign from leader when fail to start

GMHDBJD:
resign from leader when fail to start, #527 , #674

when fail to start scheduler/pessimist/optimist, resign the leader

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

### DIFF
--- a/dm/master/election.go
+++ b/dm/master/election.go
@@ -57,42 +57,21 @@ func (s *Server) electionNotify(ctx context.Context) {
 				continue
 			}
 
-			failToStart := false
-
 			if leaderInfo.ID == s.cfg.Name {
 				// this member become leader
 				log.L().Info("current member become the leader", zap.String("current member", s.cfg.Name))
-				err := s.scheduler.Start(ctx, s.etcdClient)
-				if err != nil {
-					log.L().Error("scheduler do not started", zap.Error(err))
-					failToStart = true
-				}
 
-				if !failToStart {
-					err = s.pessimist.Start(ctx, s.etcdClient)
-					if err != nil {
-						log.L().Error("pessimist do not started", zap.Error(err))
-						failToStart = true
-					}
-				}
-
-				if !failToStart {
-					err = s.optimist.Start(ctx, s.etcdClient)
-					if err != nil {
-						log.L().Error("optimist do not started", zap.Error(err))
-						failToStart = true
-					}
-				}
+				ok := s.startLeaderComponent(ctx)
 
 				failpoint.Inject("FailToStartLeader", func(val failpoint.Value) {
 					masterStrings := val.(string)
 					if strings.Contains(masterStrings, s.cfg.Name) {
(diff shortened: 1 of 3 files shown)


dm/master/election.go
```diff
@@ -55,21 +57,46 @@ func (s *Server) electionNotify(ctx context.Context) {
...
 				log.L().Info("current member become the leader", zap.String("current member", s.cfg.Name))
 				err := s.scheduler.Start(ctx, s.etcdClient)
 				if err != nil {
 					log.L().Error("scheduler do not started", zap.Error(err))
+					failToStart = true
 				}
 
 				err = s.pessimist.Start(ctx, s.etcdClient)
-				if err != nil {
+				if err != nil && !failToStart {
```
csuzhangxc, line 72: if `failToStart` is `true`, should we still try to `start` `pessimist` and `optimist`?

GMHDBJD: fix it in 5911e58

dm/master/election_test.go
```diff
@@ -0,0 +1,100 @@
...
+	c.Assert(err, check.IsNil)
+	defer client.Close()
+
+	// s1 is still the leader
+	_, leaderID, _, err := s2.election.LeaderInfo(ctx)
+	c.Assert(err, check.IsNil)
+	c.Assert(leaderID, check.Equals, cfg1.Name)
+
+	// fail to start scheduler/pessimism/optimism
+	c.Assert(failpoint.Enable("github.com/pingcap/dm/dm/master/FailToStartLeader", `return("dm-master-2")`), check.IsNil)
```
lichunzhu, line 82: Do we need to add a defer after this?
defer failpoint.Disable("github.com/pingcap/dm/dm/master/FailToStartLeader")

dm/master/election.go
```diff
@@ -55,21 +57,51 @@ func (s *Server) electionNotify(ctx context.Context) {
...
-				if err != nil {
-					log.L().Error("optimist do not started", zap.Error(err))
+
+				if !failToStart {
+					err = s.optimist.Start(ctx, s.etcdClient)
+					if err != nil {
+						log.L().Error("optimist do not started", zap.Error(err))
+						failToStart = true
+					}
+				}
```
lichunzhu, line 85: How about put these into a function? It looks clearer to me.

dm/master/election_test.go
GMHDBJD: L92 disable it. Should we also add it here?

lichunzhu: If we don't defer here and this test is failed, will it affect other unit tests?

GMHDBJD: got it. add in 8e98233

dm/master/election.go
GMHDBJD: change in 8e98233

Issue #672, opened 2020-05-20
Possibly data race when starting several tasks on one worker


lichunzhu:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do? If possible, provide a recipe for reproducing the error.
Run serveral tasks at one time on a worker.

2. What did you expect to see?
Run normally.

3. What did you see instead?
Data race at collecting metrics.
![image](https://user-images.githubusercontent.com/21104942/82419603-3ce8cc80-9ab1-11ea-950c-a42095de56a8.png)
To fix this, we should move cpu usage metrics to worker server level but not unit level.
DM v1.0 and v2.0 both have this problem.

GMHDBJD:
fix in #696

PR #696, opened 2020-06-01
dm-worker: move cpu metrics to worker server level

GMHDBJD:
Data race at collecting metrics. #672 

 move cpu usage metrics to worker server level

Tests 

 - Unit test
 - Integration test


csuzhangxc:
@lichunzhu If you have time, please take a look.

PR #703, opened 2020-06-02
fix wrong usage of failpoint in #695

GMHDBJD:
fix wrong usage of failpoint in #695, which will cause unit test fail.

failpoint generate `ok` variable and overide outer `ok`
```
ok:=true
failpoint.Inject("", func(val failpoint.Value) {
     ok=false
})
fmt.Println(ok) // ok still true
```
this pr rewrite the logic of failpoint in #695 
### Check List 

Tests 

 - Unit test
 - Integration test

### DIFF
--- a/dm/master/election.go
+++ b/dm/master/election.go
@@ -60,12 +60,14 @@ func (s *Server) electionNotify(ctx context.Context) {
 					if strings.Contains(masterStrings, s.cfg.Name) {
 						log.L().Info("fail to start leader", zap.String("failpoint", "FailToStartLeader"))
 						s.retireLeader()
+						s.election.Resign()
 						failpoint.Continue()
 					}
 				})
 
 				if !ok {
 					s.retireLeader()
+					s.election.Resign()
 					continue
 				}
 
--- a/dm/master/election.go
+++ b/dm/master/election.go
@@ -55,16 +55,6 @@ func (s *Server) electionNotify(ctx context.Context) {
 
 				ok := s.startLeaderComponent(ctx)
 
-				failpoint.Inject("FailToStartLeader", func(val failpoint.Value) {
-					masterStrings := val.(string)
-					if strings.Contains(masterStrings, s.cfg.Name) {
-						log.L().Info("fail to start leader", zap.String("failpoint", "FailToStartLeader"))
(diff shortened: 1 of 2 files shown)


GMHDBJD:
@csuzhangxc @lichunzhu PTAL 🤦

dm/master/election.go
```diff
@@ -63,23 +55,9 @@ func (s *Server) electionNotify(ctx context.Context) {
 
 				ok := s.startLeaderComponent(ctx)
 
-				failpoint.Inject("FailToStartLeader", func(val failpoint.Value) {
-					masterStrings := val.(string)
-					if strings.Contains(masterStrings, s.cfg.Name) {
-						log.L().Info("fail to start leader", zap.String("failpoint", "FailToStartLeader"))
-						ok = false
```
lichunzhu, line 70: Just curious 😕 Why failpoint can't change ok to false here?

GMHDBJD: en.... After investigation, failpoint generate
```
if _, ok := failpoint.Eval(_curpkg_("failpointName")); ok {
		ok=false
	}
```
and the `ok` override the outer ok. 😂
so actually we can just rename the variable....

lichunzhu: emmmm 🤔 get it

csuzhangxc: the name of `ok` is not OK. :scream:

dm/master/election.go
```diff
@@ -152,5 +130,25 @@ func (s *Server) startLeaderComponent(ctx context.Context) bool {
...
+		if strings.Contains(masterStrings, s.cfg.Name) {
+			log.L().Info("fail to start leader", zap.String("failpoint", "FailToStartLeader"))
+			failpoint.Return(false)
+		}
+	})
+
 	return true
 }
+
+func (s *Server) retireLeader() {
```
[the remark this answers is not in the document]
GMHDBJD: closeLeaderComponent? 🤔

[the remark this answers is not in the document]
csuzhangxc: maybe `retireLeader` is fine, because we also reset `s.leader` and `s.closeLeaderClient()`. you can just keep it.

Issue #694, opened 2020-05-29
A large replication lag for heartbeat


SE-Bin:
## Question
If Mysql and DM have lags before enabling the heartbeat in DM. After enabling the heartbeat, the Prometheus report 50 years lag in replicate between Mysql and DM 。It's not easy to understand. 
![image](https://user-images.githubusercontent.com/20888467/83252707-50491580-a1de-11ea-9cf8-b453a4b628c4.png)

Before asking a question, make sure you have:

- Searched existing Stack Overflow questions.
- Googled your question.
- Searched open and closed [GitHub issues](https://github.com/pingcap/dm/issues?q=is%3Aissue)
- Read the documentation:
    - [Detailed documentation](https://pingcap.com/docs/tools/dm/overview/)
    - [简体中文文档](https://pingcap.com/docs-cn/tools/dm/overview/)

PR #704, opened 2020-06-03
syncer: refine heartbeat

csuzhangxc:

- #706
- #694 
- #707

- compare heartbeat schema and table as case-insensitive
- do not update replication lag metric if no valid slave ts exit
- use `UTC_TIMESTAMP(6)` to replace `UTC_TIMESTAMP()`

### Check List 

Tests 

 - Unit test
 - Integration test

Related changes

### DIFF
--- a/syncer/heartbeat.go
+++ b/syncer/heartbeat.go
@@ -308,6 +308,9 @@ func (h *Heartbeat) calculateLag(ctx context.Context) error {
 	select {
 	case h.lock <- struct{}{}:
 		for taskName, ts := range h.slavesTs {
+			if ts == 0 {
+				continue // do not update metrics if no valid slave TS exists.
+			}
 			lag := masterTS - ts
 			reportLagFunc(taskName, lag)
 		}
--- a/syncer/heartbeat.go
+++ b/syncer/heartbeat.go
@@ -18,6 +18,7 @@ import (
 	"database/sql"
 	"fmt"
 	"reflect"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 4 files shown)


csuzhangxc:
@GMHDBJD PTAL

syncer/heartbeat.go
```diff
@@ -292,8 +293,7 @@ func (h *Heartbeat) createTable() error {
 
 // updateTS use `REPLACE` statement to insert or update ts
 func (h *Heartbeat) updateTS() error {
-	// when we not need to support MySQL <=5.5, we can replace with `UTC_TIMESTAMP(6)`
```
WangXiangUSTC, line 295: Does this mean we don't support MySQL <=5.5 after this pr?

tests/all_mode/run.sh
```diff
@@ -101,8 +101,8 @@ function run() {
     # use sync_diff_inspector to check data now!
     check_sync_diff $WORK_DIR $cur/conf/diff_config.toml
 
-    check_metric $WORKER1_PORT 'dm_syncer_replication_lag{task="test"}' 3 0 1
-    check_metric $WORKER2_PORT 'dm_syncer_replication_lag{task="test"}' 3 0 1
+    check_metric $WORKER1_PORT 'dm_syncer_replication_lag{task="test"}' 3 0 2
```
WangXiangUSTC, line 104: why need to increase the `upper` to `2`? It seems that use `UTC_TIMESTAMP(6)` to replace `UTC_TIMESTAMP` when update ts will decrease the replication lag.

syncer/heartbeat.go
csuzhangxc: YES, but we already highlighted this in docs (`5.5 < MySQL 版本 < 8.0`).

tests/all_mode/run.sh
csuzhangxc: I saw lag with 1.x (but < 1.5) in some cases. in the old `UTC_TIMESTAMP` it `==1`, but now it `> 1`.

WangXiangUSTC:
LGTM

PR #725, opened 2020-06-08
*: fix new shard DDL handle when the previous one has not complete

### LINKED ISSUE #628: pessimistic shard DDL can't handle new DDL when the old one has not been resolved
## Bug Report

    execute multiple DDL consecutively in close time in pessimistic shard DDL mode.

    replicate these DDLs normally.

3. What did you see instead?

    DDL coordination blocked.

    all tables in DM-workers (returned by `query-status`) are _synced_, but some sources in DM-master (returned by `show-ddl-locks`) are _unsynced_.

    this should be caused by `fail to try sync shard DDL lock, sharding ddls in ddl lock *** is different with ***`, then the DDL info received from DM-worker is ignored (and DM-worker will not resend again).

    **now, we can `pause-task` and `resume-task` to let DM-worker resend the DDL info to handle it manually.**

4. Versions of the cluster

        ```
        DM master before git commit hash b2a8fb85e0e93540382518b487d934cd2e392066
        ```

csuzhangxc:

fix #628 

do not put shard DDL info if the _done_ operation exists

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

 - Increased code complexity


pkg/shardddl/pessimism/info.go
```diff
@@ -95,6 +96,46 @@ func PutInfo(cli *clientv3.Client, info Info) (int64, error) {
...
+		return 0, false, err
+	} else if opBefore.Done {
+		// the operation with `done` exist before, abort the PUT.
+		return rev, false, nil
+	} else if utils.CompareShardingDDLs(opBefore.DDLs, info.DDLs) {
+		// TODO: try to handle put the same `done` DDL later.
+	}
+
+	// NOTE: try to PUT info if the operation still not done.
+	opNotDone := clientv3.Compare(clientv3.Value(opKey), "=", string(opsResp.Kvs[0].Value))
```
WangXiangUSTC, line 131: why this mean op not done?

syncer/shardddl/pessimist.go
```diff
@@ -65,8 +65,37 @@ func (p *Pessimist) ConstructInfo(schema, table string, DDLs []string) pessimism
...
+	}
+
+	p.logger.Warn("the previous shard DDL operation still exists, waiting for it to be deleted", zap.Stringer("info", info))
+
+	// wait for the operation to be deleted.
+	ctx2, cancel2 := context.WithCancel(ctx)
+	defer cancel2()
+	ch := make(chan pessimism.Operation, 1)
+	errCh := make(chan error, 1)
+	go pessimism.WatchOperationDelete(ctx2, p.cli, info.Task, info.Source, rev, ch, errCh)
```
WangXiangUSTC, line 88: when exiting the goroutine?

pkg/shardddl/pessimism/info.go
csuzhangxc: `opsResp.Kvs[0].Value` is an operation but not done because we have checked in L123.

To be precise, here we want to ensure that the operation has not changed to done between two transactions.

syncer/shardddl/pessimist.go
csuzhangxc: `ctx` or `ctx2` has done (`defer cancel2()`), it can exist, just we not to wait for it to exit.

PR #726, opened 2020-06-09
fix(shardddl): fix init schema fetch before applied DDL to downstream

### LINKED ISSUE #583: fetch table schema from downstream before applying optimistic shard DDL
## Feature Request

fetch schema for all upstream tables before applying any optimistic shard DDL so we can fetch the right initial schema.

to solve the problem:

0. start incrementally replicating
1. table-A apply `ADD COLUMN` to the downstream
2. table-B get row binlog event
3. table-B get the schema from downstream TiDB
4. table-B try to construct SQL from row binlog event and the schema, but mismatch because one more column in the schema

csuzhangxc:

fix #583.

- record the schema if not exist
- get the previously recorded schema if no one in schema tracker.
    - drop the tracked table if no checkpoint exists
- delete the schema when removing the lock

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity


syncer/syncer.go
```diff
@@ -1714,6 +1730,13 @@ func (s *Syncer) handleQueryEvent(ev *replication.QueryEvent, ec eventContext) e
 		}
 	})
 
+	// NOTE: this DDL may be the first event for the table (no DML replicated before), so we try to save the table point (before this event) for it, then we can achieve:
+	// - can rollback tracked table info based on this checkpoint.
+	for db, tables := range sourceTbls {
+		for table := range tables {
+			s.saveTablePoint(db, table, lastLocation)
```
csuzhangxc, line 1737: we only try to rollback the tracked table info if it has checkpoint now, see https://github.com/pingcap/dm/blob/604de8d89a78d153cecdd3b0ea3543ab30a20503/syncer/checkpoint.go#L550.

dm/master/shardddl/optimist.go
```diff
@@ -483,6 +483,17 @@ func (o *Optimist) handleLock(info optimism.Info, tts []optimism.TargetTable, sk
 	} else {
 		o.logger.Info("the shard DDL lock returned some DDLs",
 			zap.String("lock", lockID), zap.Strings("ddls", newDDLs), zap.Stringer("info", info), zap.Bool("is deleted", info.IsDeleted))
+
+		// try to record the init schema before applied the DDL to the downstream.
+		initSchema := optimism.NewInitSchema(info.Task, info.DownSchema, info.DownTable, info.TableInfoBefore)
+		rev, putted, err2 := optimism.PutInitSchemaIfNotExist(o.cli, initSchema)
+		if err2 != nil {
+			o.logger.Warn("fail to record the initial schema", zap.Stringer("info", info), log.ShortError(err2))
```
WangXiangUSTC, line 491: do we need to return err2?

syncer/syncer.go
```diff
@@ -567,32 +567,52 @@ func (s *Syncer) getTable(origSchema, origTable, renamedSchema, renamedTable str
 		return nil, terror.ErrSchemaTrackerCannotGetTable.Delegate(err, origSchema, origTable)
 	}
 
-	// if the table does not exist (IsTableNotExists(err)), continue to fetch the table from downstream and create it.
-
 	if err = s.schemaTracker.CreateSchemaIfNotExists(origSchema); err != nil {
 		return nil, terror.ErrSchemaTrackerCannotCreateSchema.Delegate(err, origSchema)
 	}
 
+	// in optimistic shard mode, we should try to get the init schema (the one before modified ty other tables) first.
```
WangXiangUSTC, line 574:
```suggestion
	// in optimistic shard mode, we should try to get the init schema (the one before modified by other tables) first.
```

dm/master/shardddl/optimist.go
csuzhangxc: changed to return `err2` in https://github.com/pingcap/dm/pull/726/commits/053fb6fb6b39c89e22a2ca30189abc4aead522b5.

PR #753, opened 2020-06-18
error_message: add workaround for error

GMHDBJD:
add workaround for some error

Tests 

 - Unit test
 - Integration test


tests/sequence_sharding/run.sh
```diff
@@ -40,6 +40,7 @@ function run() {
     run_sql_file $cur/data/db1.increment2.sql $MYSQL_HOST1 $MYSQL_PORT1 $MYSQL_PASSWORD1
     run_sql_file $cur/data/db2.increment2.sql $MYSQL_HOST2 $MYSQL_PORT2 $MYSQL_PASSWORD2
 
+    read v1
```
WangXiangUSTC, line 43: this is used for debug?

dm/config/subtask_test.go
```diff
@@ -98,47 +98,47 @@ func (t *testConfig) TestSubTaskAdjustFail(c *C) {
 				cfg.Name = ""
 				return cfg
 			},
-			"\\[.*\\] task name should not be empty",
+			".*task name should not be empty.*",
```
WangXiangUSTC, line 101: why update this?

tests/sequence_sharding/run.sh
GMHDBJD: remove it in 675693e 🤣

dm/config/subtask_test.go
GMHDBJD: change it more restrictive in 675693e

dm/worker/status.go
```diff
@@ -203,7 +203,9 @@ func statusProcessResult(pr *pb.ProcessResult) *pb.ProcessResult {
 	result := proto.Clone(pr).(*pb.ProcessResult)
 	if result != nil {
 		for i := range result.Errors {
-			result.Errors[i].Error = nil
+			if result.Errors[i].Error != nil {
+				result.Errors[i].Msg = ""
```
csuzhangxc, line 207: Can we use `TError` to replace `ProcessError` later?

then we can make 

```
                     "errors": [
                            {
                                "msg": "",
                                "error": {
                                    "ErrCode": 10006,
                                    "ErrClass": 1,
                                    "ErrScope": 2,
                                    "ErrLevel": 3,
                                    "Message": "this is the error message",
                                    "RawCause": "this is the raw cause",
                                    "Workaround": "this is the workaround"
                                }
                            }
                        ],
```

become

```
                     "errors": [
                            {
                                "ErrCode": 10006,
                                "ErrClass": 1,
                                "ErrScope": 2,
                                "ErrLevel": 3,
                                "Message": "this is the error message",
                                "RawCause": "this is the raw cause",
                                "Workaround": "this is the workaround"
                            }
                        ],
```

pkg/terror/error_list.go
```diff
@@ -549,512 +549,512 @@ const (
 // Error instances
 var (
 	// Database operation related error
-	ErrDBDriverError = New(codeDBDriverError, ClassDatabase, ScopeNotSet, LevelHigh, "database driver error")
-	ErrDBBadConn     = New(codeDBBadConn, ClassDatabase, ScopeNotSet, LevelHigh, "database driver")
-	ErrDBInvalidConn = New(codeDBInvalidConn, ClassDatabase, ScopeNotSet, LevelHigh, "database driver")
+	ErrDBDriverError = New(codeDBDriverError, ClassDatabase, ScopeNotSet, LevelHigh, "database driver error", "")
```
csuzhangxc, line 552: Can we cherry-pick this PR to `release-1.0`? because we may need to copy these error list to `release-1.0` periodically so that every error code has the same meaning in all DM versions?

dm/worker/status.go
csuzhangxc: BTW, if we can make `ErrClass`, `ErrScope` and `ErrLevel` to be shown as `string` rather than `int`, it must be better.

PR #758, opened 2020-06-23
checkpoint: flush global checkpoint when first time flush checkpoint

GMHDBJD:
### What problem does this PR solve? 
close #757

### What is changed and how it works?
flush global checkpoint when first flush checkpoint

### Check List 

Tests 

 - Unit test

PR #762, opened 2020-06-28
.*: cherrypick add workaround for error

GMHDBJD:
manually cherrypick #758 

flush global checkpoint when first flush checkpoint

### Check List 

Tests 

 - Unit test


lance6716:
LGTM

PR #763, opened 2020-06-28
checkpoint: cherrypick flush global checkpoint when first time flush checkpoint

GMHDBJD:
manually cherrypick #758 

flush global checkpoint when first time flush checkpoint


lance6716:
LGTM

Issue #543, opened 2020-03-15
HA: add and refactor metrics


csuzhangxc:
## Feature Request

**Is your feature request related to a problem? Please describe:**

**Describe the feature you'd like:**

- add more metrics about HA, like metrics in DM-master.
- refactor previous metrics for HA (like updating `instance` to `source` for metrics in DM-worker).

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

lance6716:
part of new metrics[design]:
- number of workers in different state
- number of subtasks in different state by source
- number of operate errors increment (of different type?) by worker, so we could see worker should be scheduled out.

lance6716:
part of new metrics:
- DDL errors will be reported in #790, but usage of these metrics is not very good
- going to report DDL number in un-synced/un-resolve states

PR #772, opened 2020-07-01
worker: add metrics

lance6716:
part of #543 

add/modify two types of metrics:
- add label `source_id` for state of subtasks
- add counter of number of operate errors by (worker, error type), remove all `TODO: add better metrics`

### Check List 

Tests 

 - pass original tests

Code changes

Side effects

Related changes

### DIFF
--- a/dm/worker/metrics.go
+++ b/dm/worker/metrics.go
@@ -48,6 +48,7 @@ var (
 			Help:      "state of task, 0 - invalidStage, 1 - New, 2 - Running, 3 - Paused, 4 - Stopped, 5 - Finished",
 		}, []string{"task", "source_id"})
 
+	// opErrCounter cleans on worker close, which is the same time dm-worker exits, so no explicit clean
 	opErrCounter = metricsproxy.NewCounterVec(
 		prometheus.CounterOpts{
 			Namespace: "dm",
@@ -140,5 +141,3 @@ func InitStatus(lis net.Listener) {
 func (st *SubTask) removeLabelValuesWithTaskInMetrics(task string, source string) {
 	taskState.DeleteAllAboutLabels(prometheus.Labels{"task": task, "source_id": source})
 }
-
-// opErrCounter cleans on worker close, which is the same time dm-worker exits, so no explicit clean
--- a/dm/worker/metrics.go
+++ b/dm/worker/metrics.go
@@ -137,6 +137,8 @@ func InitStatus(lis net.Listener) {
 	}
 }
 
-func (st *SubTask) removeLabelValuesWithTaskInMetrics(task string) {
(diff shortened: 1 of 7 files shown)


dm/worker/metrics.go
```diff
@@ -123,6 +137,8 @@ func InitStatus(lis net.Listener) {
 	}
 }
 
-func (st *SubTask) removeLabelValuesWithTaskInMetrics(task string) {
-	taskState.DeleteAllAboutLabels(prometheus.Labels{"task": task})
+func (st *SubTask) removeLabelValuesWithTaskInMetrics(task string, source string) {
+	taskState.DeleteAllAboutLabels(prometheus.Labels{"task": task, "source_id": source})
 }
+
+// opErrCounter cleans on worker close, which is the same time dm-worker exits, so no explicit clean
```
WangXiangUSTC, line 144: 😕

dm/worker/worker.go
```diff
@@ -330,11 +333,16 @@ func (w *Worker) resetSubtaskStage(etcdCli *clientv3.Client) (int64, error) {
 			// TODO: right operation sequences may get error when we get etcdErrCompact, need to handle it later
 			// For example, Expect: Running -(pause)-> Paused -(resume)-> Running
 			// we get an etcd compact error at the first running. If we try to "resume" it now, we will get an error
-			err = w.operateSubTaskStage(stage, subtaskCfg)
+			op, err := w.operateSubTaskStage(stage, subtaskCfg)
```
WangXiangUSTC, line 336: how about return the operate type string for these functions, so don't need to judge whether `op` is nil.

dm/worker/metrics.go
```diff
@@ -34,14 +34,27 @@ import (
...
 	taskState = metricsproxy.NewGaugeVec(
 		prometheus.GaugeOpts{
 			Namespace: "dm",
 			Subsystem: "worker",
 			Name:      "task_state",
 			Help:      "state of task, 0 - invalidStage, 1 - New, 2 - Running, 3 - Paused, 4 - Stopped, 5 - Finished",
-		}, []string{"task"})
+		}, []string{"task", "source_id"})
+
+	opErrCounter = metricsproxy.NewCounterVec(
```
WangXiangUSTC, line 51: how to use this metric? send an alert message when the value is changed?

lance6716: show in grafana for example, user could thus know worker's state and decide to retire this worker or add more worker.

dm/worker/metrics.go
lance6716: should I delete this comment? or move it to elsewhere

WangXiangUSTC: no code under this comment 😄

PR #776, opened 2020-07-02
master: add metrics for worker state

lance6716:
part of #543 

add metrics in master, to show worker state

### Check List 

Tests 

 - Manual test: pause an integration test and check dm-master-addr/metrics

Code changes

Side effects

Related changes


lance6716:
https://github.com/pingcap/dm/blob/381f20256171292fa67e0fa579b287fb88e18a27/dm/master/server_test.go#L1239-L1257

unit test `testMaster.TestOfflineMember` runs 3 servers in one process, thus read/write on metrics will report data race because they are global variable. @WangXiangUSTC

PR #769, opened 2020-06-30
check_task: ping db when apply dbconfig

GMHDBJD:
### What problem does this PR solve? 
close #766 

### What is changed and how it works?
ping database when apply db config

### Check List 

Tests 
 - Integration test

PR #778, opened 2020-07-02
check_task: cherrypick ping db when apply dbconfig

GMHDBJD:
manually cherrypick #769 
ping database when apply db config

Tests 

 - Integration test


lance6716:
LGTM

PR #787, opened 2020-07-07
master: fix unit test for OfflineMember

WangXiangUSTC:

~~when master receives GRPC request, first will execute `isLeaderAndNeedForward` to judge whether this master is the leader, and will use the `leaderClient` to forward request to the leader. But the `leaderClient` is not protected by lock, it may be close when sending requests.~~

the unit test for `OfflineMember` may failed, refine it

~~add lock~~

make sure `s3` is not leader, and then offline it.

### Check List 

Tests 

 - Unit test


lance6716:
LGTM

lance6716:
suspect these's bug in test `ha_master` 😢 timeout for 2 times

WangXiangUSTC:
> suspect these's bug in test `ha_master` 😢 timeout for 2 times

I will check it

dm/master/server_test.go
```diff
@@ -1287,7 +1287,14 @@ func (t *testMaster) TestOfflineMember(c *check.C) {
 
 	cancel3()
 	s3.Close()
-	time.Sleep(5 * time.Second)
+	// make sure s3 is not the leader, otherwise OfflineMember request will send to s3
```
WangXiangUSTC, line 1290: @lance6716 the unit test may run failed because s3 is the leader

WangXiangUSTC:
It may cause deadlock if add lock for request. For example, A receives request and add lock, and B is leader, so A  forward request to B. When B receives request, the A becomes leader, so B forward request to A, but A is locked. So I revert the change in 5b32952. It doesn't matter without lock, just return an error if the leader connection changed, and the user can retry. @GMHDBJD @lance6716

Issue #767, opened 2020-06-30
sync-diff-inspector support check table by dm's task config


WangXiangUSTC:
## Feature Request

sync-diff-inspector can get DM's task config file from DM-master, and then sync-diff-inspector generate diff task by DM's task config.

## Category

* Usability

## Value

Convenient for users to use diff.

## Value  Score

3

## User

N/A

csuzhangxc:
get subtask config API added in #789, sync-diff-inspector is waiting to be updated.

IANTHEREAL:
go ahead

PR #789, opened 2020-07-08
master: add HTTP interface for getting subtask's config

WangXiangUSTC:

add HTTP interface for getting subtask's config, so sync-diff-inspector can generate check task by subtask's config.
https://github.com/pingcap/dm/issues/767

add HTTP interface

### Check List 

Tests 

 - Integration test


dm/master/server.go
```diff
@@ -512,6 +512,45 @@ func (s *Server) OperateTask(ctx context.Context, req *pb.OperateTaskRequest) (*
 	return resp, nil
 }
 
+// GetSubTaskCfg implements MasterServer.GetSubTaskCfg
+func (s *Server) GetSubTaskCfg(ctx context.Context, req *pb.GetSubTaskCfgRequest) (*pb.GetSubTaskCfgResponse, error) {
+	log.L().Info("", zap.Stringer("payload", req), zap.String("request", "GetSubTaskCfg"))
+
+	isLeader, needForward := s.isLeaderAndNeedForward()
```
lance6716, line 519: don't forget lock, merge this and add according lock in #787 ?

WangXiangUSTC: I revert the lock in PR #787, it may cause dead lock. 😐

lance6716:
LGTM

PR #791, opened 2020-07-08
master: add metric indicate number of pending DDL state (pessimistic only)

lance6716:
part of #543. Note this PR only deal with pessimistic shard DDL lock, optimistic is more complicated and will be cover in another PR

show number of pending DDLs, in types of "Un-synced" and "Synced". Only leader of dm-master will take count, so we could collect from dm-masters and calculate sums to get current pending DDL number.

### Check List 

Tests 

 - pause `shardddl1` and check metrics work
image

Code changes

Side effects

Related changes


dm/master/election.go
```diff
@@ -152,4 +153,5 @@ func (s *Server) retireLeader() {
 	s.leader = ""
 	s.closeLeaderClient()
 	s.Unlock()
+	metrics.RemoveDDLPendingInMetrics()
```
WangXiangUSTC, line 156: this metric will reset to 0 when the leader is changed, may puzzles users  🤔

lance6716: I'm not sure how we supposed to use HA metrics. I think we could collect all dm-master's counter and add them, so non-leader should output zero.

Maybe we could forward metric requests to leader 😂 ?

WangXiangUSTC: how about wait Xuecheng give some suggestion

csuzhangxc: Maybe PD needs to solve the same problem, can we refer to it?

lance6716: PD use a reversed proxy (not understand) to forward every http request to leader, so this PR should be updated (but after I figure it out)

lance6716:
seems there're many PRs editing `dm.json`, which need resolve conflict such as rearrange panel, change panel ID... I prefer waiting other PR merged and then add panel for this PR.

PR #794, opened 2020-07-09
*: detect user cancel and omit according error

### LINKED ISSUE #673: Replace execErrorDetected with execError
## Feature Request

When syncer context is canceled while runing some sqls, it will returns context.Canceled error which will still make `execErrorDetected` become true. Finally, the task is regarded as error task.

Replace execErrorDetected with execError and filter this execError outside. 

lance6716:
fix #673 

add a function `WithUserCancelFlag` to wrap context with context values. on user cancel scenarios, set this flag so we could detect afterwards. when we meet an context cancel error, check this flag so we could omit it.

fix a TODO, BTW

### Check List 

Tests 

 - Integration test (modify case `incremental_mode`)
 - Manual test (grep log to find `filter out error caused by user cancel`)

Code changes

Side effects

Related changes


lance6716:
PTAL @lichunzhu , is this an acceptable fix

pkg/terror/error_list.go
```diff
@@ -1018,6 +1019,7 @@ var (
 	ErrWorkerRelayIsPurging          = New(codeWorkerRelayIsPurging, ClassDMWorker, ScopeInternal, LevelHigh, "relay log purger is purging, cannot start sub task %s", "Please try again later.")
 	ErrWorkerHostPortNotValid        = New(codeWorkerHostPortNotValid, ClassDMWorker, ScopeInternal, LevelHigh, "host:port '%s' not valid", "Please check configs in worker configuration file.")
 	ErrWorkerSourceNotMatch          = New(codeWorkerSourceNotMatch, ClassDMWorker, ScopeInternal, LevelHigh, "source of request does not match with source in worker", "")
+    ErrWorkerDDLLockOpNotFound       = New(codeWorkerDDLLockOpNotFound, ClassDMWorker, ScopeInternal, LevelHigh, "missing shard DDL lock operation for shard DDL info (%s)", "")
```
lichunzhu, line 1022: Do you use `make terror_check` to generate this code?

syncer/syncer.go
```diff
@@ -924,15 +923,15 @@ func (s *Syncer) syncDDL(tctx *tcontext.Context, queueBucket string, db *DBConn,
...
+				err = terror.ErrWorkerDDLLockOpNotFound.Generate(shardInfo)
 			} else {
 				err = s.optimist.DoneOperation(*(s.optimist.PendingOperation()))
 			}
 		}
 		s.jobWg.Done()
 		if err != nil {
-			s.execErrorDetected.Set(true)
 			if !utils.IsContextCanceledError(err) {
+				s.execErrorDetected.Set(true)
```
lichunzhu, line 934: IMO, for `context.Canceled` error we also should set `execErrorDetected` to true so that the other functions who probe this variable can quit ASAP. If we don't set `execErrorDetected` to true syncer may still think everything works fine.

lance6716:
PTAL @lichunzhu

pkg/context/context.go
```diff
@@ -78,3 +78,54 @@ func (c *Context) WithLogger(logger log.Logger) *Context {
...
 }
+
+type userCancelFlag struct{}
+
+// WithUserCancelFlag add an flag to context, to indicate if cancel() is called by user
+func WithUserCancelFlag(ctx context.Context) context.Context {
+	// no need for a lock, because write to this variable should happens before cancel() and read happens
+	// after ctx.Done()
+	cancelByUser := false
+	return context.WithValue(ctx, userCancelFlag{}, &cancelByUser)
```
lichunzhu, line 89: I think we'd better not introduce `context.WithValue`. It will make the code hard to understand 🤔 
WDYT @WangXiangUSTC

lance6716: WithValue but not exposed to caller

lichunzhu: I still think we needn't distinguish `user canceled` error and `error canceled` error. We can filter all the `context.Cancel` error (just log is Okay). The error that calls `context.Cancel` should be returned to user.

lance6716: reasonable, going to check whether there's severe context cancel

lance6716: addressed in https://github.com/pingcap/dm/pull/794/commits/7e1bea484ec08b3236a3ac0e82d5050ba19414ea

syncer/syncer.go
```diff
@@ -526,8 +532,14 @@ func (s *Syncer) Process(ctx context.Context, pr chan pb.ProcessResult) {
 	wg.Wait()           // wait for receive all fatal from s.runFatalChan
 
 	if err != nil {
-		syncerExitWithErrorCounter.WithLabelValues(s.cfg.Name).Inc()
-		errs = append(errs, unit.NewProcessError(err))
+		if terror.ErrSyncerCtxCanceled.Equal(err) {
```
WangXiangUSTC, line 535: can we judge `utils.IsContextCanceledError(err)` here, and then don't need to transform the error like L1816-L1819

lance6716: yes, seems I should use `*Error.Delegate` so `utils.IsContextCanceledError(err)` could work

lance6716: fixed in https://github.com/pingcap/dm/pull/794/commits/38d71eeb6d28657d59375bf20339071a04087277, PTAL

WangXiangUSTC:
LGTM

pkg/atomic2/atomic_test.go
```diff
@@ -0,0 +1,50 @@
...
+		err = errors.New("test")
+	)
+	err2 := e.Get()
+	c.Assert(err2, check.Equals, nil)
+
+	e.Set(err)
+	err2 = e.Get()
+	c.Assert(err2, check.DeepEquals, err)
+
+	err = errors.New("test2")
```
csuzhangxc, line 43: not used below?

lance6716: check it's a deep copy of `err`

csuzhangxc: oh! 👍

PR #786, opened 2020-07-07
dmctl: check master connection when start.

GMHDBJD:
### What problem does this PR solve? 
close #571 

### What is changed and how it works?
- add block and timeout when create masterClient in dmctl.
- remove tests which start dmctl when master has not started.

### Check List 

Tests 

 - Integration test

PR #802, opened 2020-07-14
dmctl: check master connection when start. (#786)

GMHDBJD:
manually cherrypick #786 

- add block and timeout when create masterClient in dmctl.
- remove tests which start dmctl when master has not started.

Tests 
 - Integration test


lance6716:
LGTM

PR #806, opened 2020-07-17
*: detect user cancel and omit according error #794

lance6716:
manually cherry-pick #794, and seperate test case (mainly copied from incremental_mode)

### Check List 

Tests 

 - Integration test

Code changes

Side effects

Related changes

### DIFF
--- a/pkg/atomic2/atomic.go
+++ b/pkg/atomic2/atomic.go
@@ -1,0 +1,38 @@
+// Copyright 2020 PingCAP, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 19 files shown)


tests/filter_user_cancel/run.sh
```diff
@@ -0,0 +1,82 @@
+#!/bin/bash
```
csuzhangxc, line 1: add this case to _others_integration.txt_.

pkg/atomic2/atomic.go
```diff
@@ -0,0 +1,43 @@
...
+
+// AtomicError implements atomic error method
+type AtomicError struct {
+	p unsafe.Pointer
+}
+
+// Get returns error
+func (e *AtomicError) Get() error {
+	failpoint.Inject("UserCancel", func(_ failpoint.Value) {
+		failpoint.Return(context.Canceled)
```
csuzhangxc, line 31: conveniently but strange for adding a failpoint in this pkg. 😭

PR #807, opened 2020-07-17
loader: support enable-ansi-quotes

### LINKED ISSUE #167: load unit can't load data with `"` in the table name
## Bug Report

loading data with `"` in the table.

data loaded.

3. What did you see instead?

_paserpkg_ parse SQL statement fail because not using the correct `sql-mode`.

4. Versions of the cluster

        ```
        before this commit (https://github.com/pingcap/dm/tree/cd743e564306a9ef103460a0a446359ae711e21d)
        ```

GMHDBJD:
close #167 

- move enable-ansi-quotes to global task config
- update tidb-tool with [tidb-tools#371](https://github.com/pingcap/tidb-tools/pull/371)
- support ANSIQUOTES in loader
- #167 actually working in DM now, add a integration test for it

### Check List 

Tests 

 - Unit test
 - Integration test


lance6716:
/run-all-tests

seems timeout errors https://internal.pingcap.net/idc-jenkins/blue/organizations/jenkins/dm_ghpr_test/detail/dm_ghpr_test/5712/pipeline

dm/config/task.go
```diff
@@ -238,7 +238,6 @@ type SyncerConfig struct {
 	EnableGTID       bool `yaml:"enable-gtid" toml:"enable-gtid" json:"enable-gtid"`
 	DisableCausality bool `yaml:"disable-detect" toml:"disable-detect" json:"disable-detect"`
 	SafeMode         bool `yaml:"safe-mode" toml:"safe-mode" json:"safe-mode"`
-	EnableANSIQuotes bool `yaml:"enable-ansi-quotes" toml:"enable-ansi-quotes" json:"enable-ansi-quotes"`
```
csuzhangxc, line 241: Should we need to keep it for compatibility to support recovery a task from a persistent task config? 😢

dm/config/subtask.go
```diff
@@ -189,6 +189,8 @@ type SubTaskConfig struct {
 
 	CleanDumpFile bool `toml:"clean-dump-file" json:"clean-dump-file"`
 
+	EnableANSIQuotes bool `toml:"ansi-quotes" json:"ansi-quotes"`
```
GMHDBJD, line 192: rename to `ansi-quotes` to avoid conflict with deprecated sycncer config `enable-ansi-quotes`

dm/config/task.go
GMHDBJD: address in f4ccf8f

dm/config/task.go
```diff
@@ -305,6 +306,8 @@ type TaskConfig struct {
 	Syncers   map[string]*SyncerConfig   `yaml:"syncers"`
 
 	CleanDumpFile bool `yaml:"clean-dump-file"`
+
+	EnableANSIQuotes bool `yaml:"ansi-quotes" toml:"ansi-quotes" json:"ansi-quotes"`
```
GMHDBJD, line 310: rename to `ansi-quotes` to avoid conflict with deprecated sycncer config `enable-ansi-quotes`

lance6716:
/run-all-tests

LGTM

csuzhangxc:
we may support detect ansi-auotes later, no need to update docs now? @lance6716 @GMHDBJD

lance6716:
we might postpone updating the doc. 

and I wonder if the origin issue https://github.com/pingcap/dm/issues/167 was caused by global variable ANSI_QUOTES so mydumper/dumpling dump data can't fit loader. is this right? so as a integrated data migration tool we could export ANSI_QUOTES to file for loader's using rather than require dumpling adding this feature.

csuzhangxc:
@lance6716

> I wonder if the origin issue #167 was caused by global variable ANSI_QUOTES so mydumper/dumpling dump data can't fit loader. is this right

Right, I think it's different global variable between upstream and downstream.

> as a integrated data migration tool we could export ANSI_QUOTES to file for loader's using rather than require dumpling adding this feature.

:+1: good idea!!!

PR #814, opened 2020-07-22
shardddl/optimism: add `synced` as struct member

lance6716:
Support show optimistic lock number in grafana. After this PR, we could know old state before `TrySync`, and increase new state/decrease old state number in metric.

reuse `(*Lock).syncStatus` to get synced state. `(*Lock).syncStatus` using `(*Lock).table` and `(*Lock).joined` to compare, get result of every table in `(*Lock).table` is same or not. `(*Lock).joined` is only maintained in `(*Lock).TrySync`, `(*Lock).table` is edited in `(*Lock).TrySync` and `(*Lock).addTables`.

for `(*Lock).TrySync`, before returning, when editing of `(*Lock).table` and `(*Lock).joined` finished, update `(*Lock).synced`.

for `(*Lock).addTables`, it only add `(*Lock).joined` into `(*Lock).table` and doesn't change `synced` state.

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes


pkg/shardddl/optimism/lock.go
```diff
@@ -121,6 +123,8 @@ func (l *Lock) TrySync(callerSource, callerSchema, callerTable string,
 	// if any real conflicts after joined exist, they will be detected by the following steps.
 	var cmp int
 	if cmp, err = newTable.Compare(oldJoined); err == nil && cmp == 0 {
+		_, remain := l.syncStatus()
+		l.synced = remain == 0
```
csuzhangxc, line 127: How about doing this in `defer`?

lance6716: addressed in https://github.com/pingcap/dm/pull/814/commits/95408c3290ff554c652fd44f737ce9a82ffbc2df

PR #815, opened 2020-07-22
loader: support enable-ansi-quotes (#807)

GMHDBJD:
manually cherry pick #807 

- support ANSIQUOTES in loader
- fix error workaround for source-id not exist


_utils/terror_gen/errors_release.txt
```diff
@@ -145,7 +145,7 @@ ErrConfigMydumperCfgNotFound,[code=20027:class=config:scope=internal:level=mediu
...
 ErrConfigLoaderCfgNotFound,[code=20029:class=config:scope=internal:level=medium], "Message: mysql-instance(%d)'s loader config %s not exist in loaders, Workaround: Please check the `loader-config-name` config in task configuration file."
 ErrConfigSyncerCfgNotFound,[code=20030:class=config:scope=internal:level=medium], "Message: mysql-instance(%d)'s syncer config %s not exist in syncer, Workaround: Please check the `syncer-config-name` config in task configuration file."
-ErrConfigSourceIDNotFound,[code=20031:class=config:scope=internal:level=medium], "Message: source %s in deployment configuration not found, Workaround: Please use `operate-source create source-config-file-path` to add source."
+ErrConfigSourceIDNotFound,[code=20031:class=config:scope=internal:level=medium], "Message: source %s in deployment configuration not found, Workaround: Please check the `source-id` config in worker configuration file."
```
csuzhangxc, line 148: we have `source-id` both in DM-worker's config and DM-master's config (in _release-1.0_)

dm/config/task.go
```diff
@@ -474,6 +477,11 @@ func (c *TaskConfig) adjust() error {
 			inst.Syncer.WorkerCount = inst.SyncerThread
 		}
 
+		// for backward compatible, set global config `ansi-quotes: true` if any syncer is true
+		if inst.Syncer.EnableANSIQuotes == true {
```
lance6716, line 481: no need for `bool == true` 🤔

csuzhangxc:
> will there be more SQL modes need us processing in future? maybe we could pass SQL mode string instead of adding boolean one by one in future PR.

pass SQL mode should be better, and maybe we need to compatible between these (`sql_mode` under `session`) two config items.

in fact, `enable-ansi-quotes` was added before `session` added, 😢

_utils/terror_gen/errors_release.txt
GMHDBJD: address in df278ac

dm/config/task.go
GMHDBJD: address in 2931f93

lance6716:
LGTM

GMHDBJD:
We use ansi_quotes for parser. If user set upstream and downstream @@GLOBAL.sql_mode="ANIS_QUOTES", then they may not set sql_mode in task config. 🤔

csuzhangxc:
> We use ansi_quotes for parser. If user set upstream and downstream @@GLOBAL.sql_mode="ANIS_QUOTES", then they may not set session sql_mode in task config. 🤔

em, but if they set one of them, they may expect the other place also taken effect?

PR #821, opened 2020-07-23
portal: minor fix on db config

WangXiangUSTC:

in pr https://github.com/pingcap/tidb-tools/pull/364 omit the `Password` for json, and dm-portal can not get the `Password`

use `DBConfig` defined in portal

### Check List 

Tests 

 - Manual test (deploy dm-portal and generate task config file)


csuzhangxc:
should we need to cherry-pick this to _release-1.0_?

WangXiangUSTC:
> should we need to cherry-pick this to _release-1.0_?

yes, I add a cherry-pick label

Issue #773, opened 2020-07-02
test not stable


lance6716:
record broken test case here

#### Unit test

- [x] testServer.TestTaskAutoResume: https://github.com/pingcap/dm/pull/895

- [x] testForEtcd.TestInfoEtcd: #785 dynamic increase waiting time (by retry)

- [ ] testForEtcd.TestOperationEtcd: #785 dynamic increase waiting time (by retry), [Jenkins outdate]

- [x] testOptimist.testOptimist: #785 dynamic increase waiting time (by retry)

- [x] testMockScheduler: #785 increase lease TTL

- [ ] testForEtcd.TestWorkerKeepAlive (added notes)

- [ ] testMockScheduler (added notes)

- [x] testSubTask.TestPauseAndResumeSubtask: https://github.com/pingcap/dm/pull/805

- [x] testMaster.TestOfflineMember: https://github.com/pingcap/dm/pull/824

- [ ] testForEtcd.TestKeepAliveRevokeLease: [Jenkins outdate]

- [ ] testScheduler.TestRestartScheduler: https://github.com/pingcap/dm/pull/812, [Jenkins outdate]

- [x] testMaster.TestOperateSource: https://github.com/pingcap/dm/pull/883

#### Integration test

- [ ] dmctl_basic: [Jenkins outdate]

- [x] ha_cases: fixed by https://github.com/pingcap/tidb/issues/18456

- [ ] shardddl1 (DM-001 failed): https://internal.pingcap.net/idc-jenkins/blue/organizations/jenkins/dm_ghpr_test/detail/dm_ghpr_test/6108/pipeline/88

- [x] shardddl1 (Unknown database): https://internal.pingcap.net/idc-jenkins/blue/organizations/jenkins/dm_ghpr_test/detail/dm_ghpr_test/6133/pipeline/88,
seems nothing error
[dm-filter-out-ddl-error.log](https://github.com/pingcap/dm/files/4985782/dm-grep.log),

- [x] sharding2: #785 dynamic increase waiting time (by retry)

- [ ] shardddl3: [Jenkins outdate]

- [ ] http_apis: [Jenkins outdate]

- [x] sharding: https://github.com/pingcap/dm/pull/871

- [ ] ha: [Jenkins outdate]

- [ ] sequence_sharding: [Jenkins outdate]

- [ ] load_interrupt: [Jenkins outdate]

- [ ] sequence_sharding_optimistic: [Jenkins outdate]

- [x] dmctl_basic: #868

[...]

PR #824, opened 2020-07-23
master: fix unit test for OfflineMember

WangXiangUSTC:

fix the unit test https://github.com/pingcap/dm/issues/773#issuecomment-655873038

I run this test case 30 times in my local computer, and it all run success.

evict leader for master which is to be offline

### Check List 

Tests 

 - Unit test

### DIFF
--- a/dm/master/server.go
+++ b/dm/master/server.go
@@ -271,6 +271,9 @@ func (s *Server) Close() {
 		return
 	}
 	log.L().Info("closing server")
+	defer func() {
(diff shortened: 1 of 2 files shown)


WangXiangUSTC:
https://internal.pingcap.net/idc-jenkins/blue/organizations/jenkins/dm_ghpr_test/detail/dm_ghpr_test/5792/pipeline
failed because of `sharding` 🥶

lance6716:
LGTM

PR #828, opened 2020-07-24
*: change source config format to YAML

### LINKED ISSUE #816: add YAML format support for source config
## Feature Request

we use YAML format for task config, but use TOML for source config now.

we can unify to use YAML for source config.

NOTE: processes config of DM-master, DM-worker and dmctl can keep to use TOML because they often deployed with DM-Ansible or TiUP.

lance6716:
fix #816 

add `(*SourceConfig).ParseYaml`, change `operate-source` and `update-relay` using this method, keep etcd using TOML format.

### Check List 

Tests 

 - Unit test
 - Integration test
- Manual test (check print-sample-config)

Code changes

 - Has interface (user interface) methods change

Side effects

Related changes


dm/config/source_config.go
```diff
@@ -270,8 +292,15 @@ func (c *SourceConfig) AdjustServerID(ctx context.Context, db *sql.DB) error {
...
-	return c.check(&metaData, err)
+	content, err := ioutil.ReadFile(path)
+	if err != nil {
+		return terror.ErrConfigReadCfgFromFile.Delegate(err, path)
+	}
+	if err := yaml.UnmarshalStrict(content, c); err != nil {
+		return terror.ErrConfigYamlTransform.Delegate(err, "decode source config")
+	}
+	c.adjust()
+	return nil
```
WangXiangUSTC, line 303: how about using `ParseYaml`

lance6716: `(*SourceConfig).check(metaData)` also checked the undecoded items, `UnmarshalStrict` is similar to that behaviour

PR #817, opened 2020-07-23
worker, utils: show same master status in one response

lance6716:
### What problem does this PR solve? 
fix https://github.com/pingcap/dm/issues/727

### What is changed and how it works?
when one response contains multiple relay and syncer status, use the larger master binlog position overwrite them and re-check status.

### Check List 

Tests 

 - Unit test

Code changes

 - Has exported function/method change

Side effects

Related changes

 - Need to cherry-pick to the **master** branch

PR #830, opened 2020-07-25
worker, utils: show same master status in one response (#817)

lance6716:
manually cherry-pick #817

when one response contains multiple relay and syncer status, use the larger master binlog position overwrite them and re-check status.

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes


lance6716:
@WangXiangUSTC PTAL

PR #290, opened 2019-09-17
subtask: verify task name

3pointer:
### What problem does this PR solve? 
verify subtask name

### What is changed and how it works?
according to MySQL [identifiers](https://dev.mysql.com/doc/refman/5.7/en/identifiers.html)
It's better to check task name with 
```[0-9,a-z,A-Z$_] (basic Latin letters, digits 0-9, dollar, underscore) ```

### Check List 

Tests 

 - Unit test
 - Integration test
 - Manual test (add detailed scripts or steps below)
 - No code

Code changes

 - Has exported function/method change
 - Has exported variable/fields change
 - Has interface methods change
 - Has persistent data change

Side effects

 - Possible performance regression
 - Increased code complexity
 - Breaking backward compatibility

Related changes

 - Need to cherry-pick to the release branch
 - Need to update the documentation
 - Need to update the `dm/dm-ansible`
 - Need to be included in the release note

PR #835, opened 2020-07-27
*: back quote and escape MySQL identifier

lance6716:
duplicate of #290

back quote and escape MySQL identifier (mainly `task.Name` to construct table name)

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

Related changes


loader/checkpoint.go
```diff
@@ -79,7 +80,7 @@ type RemoteCheckPoint struct {
 	conn           *DBConn
 	id             string
 	schema         string
-	table          string
+	tableName      string
```
WangXiangUSTC, line 83: how about adding a comment, tableName contains schema name

lance6716:
IT is useful 😂 PTAL @WangXiangUSTC @csuzhangxc

WangXiangUSTC:
LGTM

Issue #774, opened 2020-07-02
upgrade DM 1.x to 2.0


GMHDBJD:
## Feature Request

**Is your feature request related to a problem? Please describe:**

- Deploy and mange DM cluster using DM-Ansible is too complex and diffcult to use.
- upgrade any incompatible data (like checkpoint) when bootstrapping a new version of DM

**Describe the feature you'd like:**

- Use TiUP for 2.0 to optimized in terms of topology, configuration files, startup methods, version upgrades, expansion and contraction, resetting components, etc.
- internal upgrade mechanism from 1.0 to 2.0

PR #846, opened 2020-07-30
feat: get subtask config and status of v1.0.x from dm-worker

csuzhangxc:

part of #774.

support to get subtask config and status of v1.0.x from dm-worker if it's deployed in a v1.0.x directory.

**for reviewers**:

- the following files are copied and refined from _release-1.0_ branch
    - _pkg/v1workermeta/meta.go_
    - _pkg/v1workermeta/meta_test.go_
    - _pkg/v1workermeta/db.go_
- _pkg/v1workermeta/v106_data_for_test/kv_ is generated by v1.0.6

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes


dm/worker/v1meta.go
```diff
@@ -0,0 +1,61 @@
...
+
+	"go.uber.org/zap"
+
+	"github.com/pingcap/dm/dm/pb"
+	"github.com/pingcap/dm/pkg/log"
+	"github.com/pingcap/dm/pkg/v1workermeta"
+)
+
+// OperateV1Meta implements WorkerServer.OperateV1Meta.
+func (s *Server) OperateV1Meta(ctx context.Context, req *pb.OperateV1MetaRequest) (*pb.OperateV1MetaResponse, error) {
```
csuzhangxc, line 28: I'll add an integration test case after all parts of #774 ready.

pkg/v1workermeta/api.go
```diff
@@ -0,0 +1,90 @@
...
+	"github.com/pingcap/dm/dm/config"
+	"github.com/pingcap/dm/dm/pb"
+	"github.com/pingcap/dm/pkg/terror"
+	"github.com/pingcap/dm/pkg/utils"
+)
+
+// `var` rather than `const` for testing.
+var (
+	// metaPath is the meta data path in v1.0.x.
+	metaPath = "./dm_worker_meta"
```
lance6716, line 31: user may set another value of `meta-dir` in v1.0.x, so we may change this set by configuration.

pkg/v1workermeta/meta.go
```diff
@@ -0,0 +1,161 @@
...
+
+// encodeTaskMetaKey encodes take name into a task meta key.
+func encodeTaskMetaKey(name string) []byte {
+	key := append([]byte{}, taskMetaPrefix...)
+	return append(key, name...)
+}
+
+// loadTaskMetas loads all task metas from kv db.
+func loadTaskMetas(db *leveldb.DB) (map[string]*pb.V1SubTaskMeta, error) {
+	if helper.IsNil(db) {
```
lance6716, line 63: seems in v1.0 we use `IsNil` to check an interface, but there's no need

pkg/v1workermeta/meta.go
```diff
@@ -0,0 +1,161 @@
...
+	)
+
+	iter := db.NewIterator(util.BytesPrefix(taskMetaPrefix), nil)
+	for iter.Next() {
+		taskBytes := iter.Value()
+		task := &pb.V1SubTaskMeta{}
+		err = task.Unmarshal(taskBytes)
+		if err != nil {
+			err = terror.ErrWorkerLogUnmarshalTaskMeta.Delegate(err, taskBytes)
+			break
```
lance6716, line 79: seems we could `iter.Release()` and `return nil, ...` here

pkg/v1workermeta/meta.go
```diff
@@ -0,0 +1,161 @@
...
+		return nil, terror.ErrWorkerLogFetchTaskFromMeta.Delegate(err, taskMetaPrefix)
+	}
+
+	return tasks, nil
+}
+
+// setTaskMeta saves task meta into kv db.
+// it's used for testing.
+func setTaskMeta(db *leveldb.DB, task *pb.V1SubTaskMeta) error {
+	if helper.IsNil(db) {
```
lance6716, line 100: ditto

pkg/v1workermeta/meta.go
```diff
@@ -0,0 +1,161 @@
...
+		return terror.ErrWorkerLogSaveTaskMeta.Delegate(err, task.Name)
+	}
+
+	return nil
+}
+
+// deleteTaskMeta deletes task meta from kv DB.
+// it's used for testing.
+func deleteTaskMeta(db *leveldb.DB, name string) error {
+	if helper.IsNil(db) {
```
lance6716, line 125: ditto

pkg/v1workermeta/api.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+
+	return meta.TasksMeta(), nil
+}
+
+// RemoveSubtasksMeta removes subtasks' metadata.
+// this is often called after upgraded from v1.0.x to v2.0.x,
+// so no need to handle again after re-started the DM-worker process.
+func RemoveSubtasksMeta() error {
+	return os.RemoveAll(metaPath)
```
lance6716, line 76: since user could config this path by `meta-dir`, we may only remove necessary files to avoid deleting user's file 😂

pkg/v1workermeta/api.go
```diff
@@ -0,0 +1,90 @@
...
+	// so we overwrite it with another filed which has the same struct tag `chunk-filesize` here.
+	// but if set `chunk-filesize: 64` in a YAML file, both v1.0.x (int64) and v2.0.x (string) can support it.
+	ChunkFilesize int64 `yaml:"chunk-filesize" toml:"chunk-filesize" json:"chunk-filesize"` // -F, --chunk-filesize
+}
+
+// GetSubtasksMeta gets all subtasks' meta (config and status) from `dm_worker_meta` in v1.0.x.
+func GetSubtasksMeta() (map[string]*pb.V1SubTaskMeta, error) {
+	// check if meta data exist.
+	if !utils.IsDirExists(metaPath) || !utils.IsDirExists(dbPath) {
+		return nil, nil
```
GMHDBJD, line 53: How about return err if not exist?

pkg/v1workermeta/api.go
```diff
@@ -0,0 +1,90 @@
...
+func SubTaskConfigFromV1TOML(data []byte) (config.SubTaskConfig, error) {
+	var v1Cfg v1SubTaskConfig
+	_, err := toml.Decode(string(data), &v1Cfg)
+	if err != nil {
+		return config.SubTaskConfig{}, terror.ErrConfigTomlTransform.Delegate(err, "decode v1 subtask config from data")
+	}
+
+	cfg := v1Cfg.SubTaskConfig
+	cfg.MydumperConfig.ChunkFilesize = strconv.FormatInt(v1Cfg.ChunkFilesize, 10)
+	return cfg, cfg.Adjust(true)
```
GMHDBJD, line 89: Need divide it into two lines？

pkg/v1workermeta/api.go
csuzhangxc: I plan to get subtask meta only for the first time upgrading from v1.0.x to v2.0.x, and if upgraded, I want to remove the meta.

so if no meta exists, I want to do nothing.

Did you mean return `err` and handle it in the outer caller?

lance6716: Maybe return `err` and let upper function decide the logic (I don't know which is better)

pkg/v1workermeta/api.go
csuzhangxc: YES, users may set `meta-dir`. but I want to only support upgrading from DM-Ansible deployed v1.0.x to v2.0.x, and if deployed with DM-Ansible, this item should not be set.

BTW, I don’t really want to add new configuration items in v2.0.x 😭 .

How about checking if the path is a valid meta directory and only do something if it is?

lance6716: Okay, I think it's fine if DM-Ansible call this upgrading function, and we could warn user in manual upgrading doc.

pkg/v1workermeta/api.go
csuzhangxc: @GMHDBJD what's your opinion?

GMHDBJD: > so if no meta exists, I want to do nothing.

If meta does not exist, will the upgrade fail? If so, I think we can simply return error, otherwise nil will be ok. Because no other calls this func.

csuzhangxc: > If meta does not exist

one of the following two cases:
1. the version is not v1.0.x (not supported)
   - this should be handled in DM-worker/DM-master/TiUP-DM or just noticed in the doc (in fact, I plan to choose this simple one).
2. already upgraded (already removed)
   - this should do nothing.

GMHDBJD: > just noticed in the doc 

That's OK

pkg/v1workermeta/meta.go
csuzhangxc: changed to `db == nil` in https://github.com/pingcap/dm/pull/846/commits/042b647f3d97e76cca0e6318e0d1b7b1b769d75b.

pkg/v1workermeta/meta.go
csuzhangxc: changed to `db == nil` in https://github.com/pingcap/dm/pull/846/commits/042b647f3d97e76cca0e6318e0d1b7b1b769d75b.

pkg/v1workermeta/meta.go
csuzhangxc: changed to `db == nil` in https://github.com/pingcap/dm/pull/846/commits/042b647f3d97e76cca0e6318e0d1b7b1b769d75b.

pkg/v1workermeta/meta.go
csuzhangxc: updated in https://github.com/pingcap/dm/pull/846/commits/042b647f3d97e76cca0e6318e0d1b7b1b769d75b.

pkg/v1workermeta/api.go
csuzhangxc: add check in https://github.com/pingcap/dm/pull/846/commits/042b647f3d97e76cca0e6318e0d1b7b1b769d75b.

pkg/v1workermeta/api.go
csuzhangxc: fixed in https://github.com/pingcap/dm/pull/846/commits/042b647f3d97e76cca0e6318e0d1b7b1b769d75b.

pkg/v1workermeta/api_test.go
```diff
@@ -0,0 +1,155 @@
...
+	defer func() {
+		metaPath = oldMetaPath
+		dbPath = oldDBPath
+	}()
+
+	metaPath = c.MkDir()
+	dbPath = filepath.Join(metaPath, "kv")
+
+	// copy test data to a temp directory.
+	copyDir(c, srcDBPath, dbPath)
```
lance6716, line 50: seems go prefer `copy(dst, src)`

pkg/v1workermeta/api_test.go
```diff
@@ -0,0 +1,155 @@
...
+
+	// get subtasks meta.
+	meta, err := GetSubtasksMeta()
+	c.Assert(err, IsNil)
+
+	// verify tasks meta.
+	// - task_single:
+	//   - no shard task, Running stage.
+	//   - shard task, Paused stage.
```
lance6716, line 59:
```suggestion
	// - task_shard
	//   - shard task, Paused stage.
```

pkg/v1workermeta/meta_test.go
```diff
@@ -0,0 +1,140 @@
...
+	c.Assert(err, ErrorMatches, ".*empty task.*")
+
+	err = setTaskMeta(db, &pb.V1SubTaskMeta{})
+	c.Assert(err, ErrorMatches, ".*empty task.*")
+
+	c.Assert(setTaskMeta(db, testTask1Meta), IsNil)
+	c.Assert(setTaskMeta(db, testTask2Meta), IsNil)
+
+	// load task meta
+	metaDB, err := newMeta(nil)
```
lance6716, line 112: `TestNewMetaDB` has checked `newMeta(nil)`

csuzhangxc: removed in https://github.com/pingcap/dm/pull/846/commits/fd789fa1bb32824556420813439ca510748aa333.

pkg/v1workermeta/api_test.go
csuzhangxc: updated in https://github.com/pingcap/dm/pull/846/commits/a638dcdc19d0311a4af8e42eec4279849b0eb8b3.

lance6716:
LGTM

PR #848, opened 2020-07-30
*: use terror for some error

lance6716:
reduce some TODO: add terror

use terror instead of error

### Check List 

Tests 

Code changes

Side effects

Related changes


pkg/ha/stage.go
```diff
@@ -16,7 +16,7 @@ package ha
 import (
 	"context"
 	"encoding/json"
-	"fmt"
+	"github.com/pingcap/dm/pkg/terror"
```
csuzhangxc, line 19: re-order it.

pkg/ha/subtask.go
```diff
@@ -15,7 +15,7 @@ package ha
 
 import (
 	"context"
-	"fmt"
+	"github.com/pingcap/dm/pkg/terror"
```
csuzhangxc, line 18: ditto.

pkg/ha/task.go
```diff
@@ -15,7 +15,7 @@ package ha
 
 import (
 	"context"
-	"fmt"
+	"github.com/pingcap/dm/pkg/terror"
```
csuzhangxc, line 18: ditto.

pkg/ha/stage.go
lance6716: https://github.com/pingcap/dm/pull/848/commits/c6e4bf9a5c25b7f9cdb9cd4c73ddea768e72f6f2

do you have any tools recommended to handle grouping-import 😂

csuzhangxc: not found yet, 😢

PR #822, opened 2020-07-23
error_message: remove stack error for flag

GMHDBJD:
### What problem does this PR solve? 
close #818 

### What is changed and how it works?
change ErrorStack to terror.Message

Tests 
 - Manual test (add detailed scripts or steps below)

PR #849, opened 2020-07-31
remove stack error for flag (#822)

GMHDBJD:
manually cherrypick #822 
change ErrorStack to terror.Message for flag


lance6716:
LGTM

PR #857, opened 2020-08-04
mydumper, worker: don't auto-resume when can't acquire global lock

### LINKED ISSUE #837: do not "auto resume" for errors in dump unit
## Bug Report

   Run a task with "Couldn't acquire global lock"

   the task keep paused

3. What did you see instead?

    still "auto resume".

    for errors in dump unit, most of them should not succeed after resumed, so no extra workload on upstream MySQL/MariaDB is better.

4. Versions of the cluster

        ```
        v1.0.6
        ```

lance6716:
close #837 

add a terror and check it in `isResumableError`

### Check List 

Tests 

 - Integration test

Code changes

Side effects

Related changes

**- need cherry-pick terror to master**

### DIFF
--- a/mydumper/mydumper.go
+++ b/mydumper/mydumper.go
@@ -168,11 +168,18 @@ func (m *Mydumper) spawn(ctx context.Context) ([]byte, error) {
 				continue
 			case "ERROR":
 				m.logger.Error(string(msg))
+				if strings.HasPrefix(string(msg), "Couldn't acquire global lock") {
+					err = terror.ErrDumpUnitGlobalLock
+				}
 				continue
 			}
 		}
 		stderr.Write(line)
 		stderr.WriteByte('\n')
+		if err != nil {
+			stdout.Write(stderr.Bytes())
+			return stdout.Bytes(), err
+		}
 	}
 	if err = scanner.Err(); err != nil {
 		stdout.Write(stderr.Bytes())
--- a/mydumper/mydumper.go
+++ b/mydumper/mydumper.go
@@ -152,6 +152,11 @@ func (m *Mydumper) spawn(ctx context.Context) ([]byte, error) {
 	// so we parse all these lines and translate into our own logs.
 	scanner := bufio.NewScanner(stderrPipe)
 	for scanner.Scan() {
+		// handle errors in scanning
+		if err != nil {
+			stdout.Write(stderr.Bytes())
+			return stdout.Bytes(), err
+		}
 		line := scanner.Bytes()
 		if loc := mydumperLogRegexp.FindSubmatchIndex(line); len(loc) == 4 {
 			level := string(line[loc[2]:loc[3]])
@@ -176,10 +181,6 @@ func (m *Mydumper) spawn(ctx context.Context) ([]byte, error) {
 		}
 		stderr.Write(line)
 		stderr.WriteByte('\n')
-		if err != nil {
-			stdout.Write(stderr.Bytes())
-			return stdout.Bytes(), err
-		}
 	}
 	if err = scanner.Err(); err != nil {
 		stdout.Write(stderr.Bytes())
--- a/mydumper/mydumper.go
+++ b/mydumper/mydumper.go
@@ -133,6 +133,7 @@ func (m *Mydumper) spawn(ctx context.Context) ([]byte, error) {
 	var (
 		stdout bytes.Buffer
 		stderr bytes.Buffer
+		errMsg []byte
 	)
 	cmd := exec.CommandContext(ctx, m.cfg.MydumperPath, m.args...)
 	cmd.Stdout = &stdout
@@ -167,6 +168,7 @@ func (m *Mydumper) spawn(ctx context.Context) ([]byte, error) {
 				m.logger.Error(string(msg))
 				if strings.HasPrefix(string(msg), "Couldn't acquire global lock") {
 					err = terror.ErrDumpUnitGlobalLock
+					errMsg = msg
 				}
 			}
 		} else {
@@ -176,8 +178,7 @@ func (m *Mydumper) spawn(ctx context.Context) ([]byte, error) {
 
 		// handle errors in scanning
 		if err != nil {
-			stdout.Write(stderr.Bytes())
-			return stdout.Bytes(), err
+			return errMsg, err
 		}
 	}
 	if err = scanner.Err(); err != nil {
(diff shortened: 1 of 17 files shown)


csuzhangxc:
> NOTE: should this terror also be added in master, to avoid future conflict?

YES. all errors in all branches should always be the same, and in fact, before releasing any version, we must ensure errros codes have synced between branches.

csuzhangxc:
BTW, does the master branch (with the newest dumpling) has this problem?

lance6716:
> BTW, does the master branch (with the newest dumpling) has this problem?

if this fail is caused by lack of RELOAD privilege, dumpling could keep pause. 

```
"Error 1227: Access denied; you need (at least one of) the RELOAD privilege(s) for this operation"
```

but we can't simulate other fail reason

csuzhangxc:
> > BTW, does the master branch (with the newest dumpling) has this problem?
> 
> if this fail is caused by lack of RELOAD privilege, dumpling could keep pause.
> 
> ```
> "Error 1227: Access denied; you need (at least one of) the RELOAD privilege(s) for this operation"
> ```
> 
> but we can't simulate other fail reason

OK, we can only fix it in _release-1.0_ now.

pkg/terror/error_list.go
```diff
@@ -808,6 +809,7 @@ var (
 	ErrDumpUnitRuntime        = New(codeDumpUnitRuntime, ClassDumpUnit, ScopeInternal, LevelHigh, "mydumper runs with error", "")
 	ErrDumpUnitGenTableRouter = New(codeDumpUnitGenTableRouter, ClassDumpUnit, ScopeInternal, LevelHigh, "generate table router", "Please check `routes` config in task configuration file.")
 	ErrDumpUnitGenBAList      = New(codeDumpUnitGenBAList, ClassDumpUnit, ScopeInternal, LevelHigh, "generate block allow list", "Please check the `block-allow-list` config in task configuration file.")
+	ErrDumpUnitGlobalLock     = New(codeDumpUnitGlobalLock, ClassDumpUnit, ScopeInternal, LevelHigh, "Couldn't acquire global lock", "Please check upstream privilege about FTWL, or add --no-locks to extra-args of mydumpers")
```
csuzhangxc, line 812: my fault, it's `FTWRL`.

BTW, wrap ``` `--no-locks` ```?

tests/full_mode/run.sh
```diff
@@ -6,7 +6,45 @@ cur=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
...
+
+    cp $cur/conf/dm-task.yaml $WORK_DIR/dm-task.yaml
+    sed -i '/timezone/i\ignore-checking-items: ["dump_privilege"]' $WORK_DIR/dm-task.yaml
+    dmctl_start_task $WORK_DIR/dm-task.yaml
+
+    check_log_contains $WORK_DIR/worker1/log/dm-worker.log "Couldn't acquire global lock"
+    check_log_contains $WORK_DIR/worker2/log/dm-worker.log "Couldn't acquire global lock"
+    run_dm_ctl $WORK_DIR "127.0.0.1:$MASTER_PORT" \
+        "query-status test" \
+        "\"stage\": \"Paused\"" 4
```
GMHDBJD, line 39: check error message and workaround here?

pkg/terror/error_list.go
lance6716: fixed in https://github.com/pingcap/dm/pull/857/commits/8c017f5c03d762303cdddef19a09496da8a01d88

tests/full_mode/run.sh
lance6716: fixed in https://github.com/pingcap/dm/pull/857/commits/8c017f5c03d762303cdddef19a09496da8a01d88

mydumper/mydumper.go
```diff
@@ -168,11 +168,18 @@ func (m *Mydumper) spawn(ctx context.Context) ([]byte, error) {
 				continue
 			case "ERROR":
 				m.logger.Error(string(msg))
+				if strings.HasPrefix(string(msg), "Couldn't acquire global lock") {
+					err = terror.ErrDumpUnitGlobalLock
```
csuzhangxc, line 172: Is it possible that this `err` overwrite by any later one, like in L184?

mydumper/mydumper.go
```diff
@@ -152,6 +152,11 @@ func (m *Mydumper) spawn(ctx context.Context) ([]byte, error) {
 	// so we parse all these lines and translate into our own logs.
 	scanner := bufio.NewScanner(stderrPipe)
 	for scanner.Scan() {
+		// handle errors in scanning
+		if err != nil {
+			stdout.Write(stderr.Bytes())
+			return stdout.Bytes(), err
```
csuzhangxc, line 158: is it possible to lose some remaining log? 😂

mydumper/mydumper.go
lance6716: https://github.com/pingcap/dm/pull/857/commits/433edcb16b8397d6026f2f273afd9168d4520588 

by set `err` function will return earlier than L184, this commit ⬆️ detect `err` in a more early place

mydumper/mydumper.go
lance6716: change this error-check to where after flush line, and correct return output message

lance6716: seems need to drain and output `stderr` of `mydumper` 😂

lance6716: when `terror.ErrDumpUnitGlobalLock` hanppens, there could be some remaining log which mydumper has output to `stderr`. Since go doesn't have a non-blocking IO read, I guess it's not neat to fetch these remaining log. So if we don't output remaining log, behaviour could be just like "stop on first error", is this acceptable?

lance6716: * non-blocking IO could be implement by another goroutine and select, not very hard though

csuzhangxc: Can we just:

1. record the first error
2. out all log as we did before
3. return the recorded error

lance6716: OK, then we depend on mydumper will exit quickly so we could get the error quickly. I'll left a TODO to switch to non-blocking read If we want to detect more error and when that error happens, mydumper continue to run.

lance6716: done in latest two commit

PR #866, opened 2020-08-06
*: don't auto-resume for dump unit error

### LINKED ISSUE #837: do not "auto resume" for errors in dump unit

lance6716:
close #837

move logic together, check PR does take effect

### Check List 

Tests 

 - Integration test

Code changes

Side effects

Related changes

 - Need to cherry-pick to **master**


dm/worker/task_checker.go
```diff
@@ -302,8 +300,10 @@ func (tsc *realTaskStatusChecker) getResumeStrategy(stStatus *pb.SubTaskStatus,
 	// TODO: use different strategies based on the error detail
 	for _, processErr := range stStatus.Result.Errors {
 		if !isResumableError(processErr) {
+			tsc.l.Info("error is not resumable", zap.Stringer("error", processErr))
```
csuzhangxc, line 303: Maybe too many logs? but may not a problem.

lance6716: used to check task_checker does find an error and decide not to resume it. going to enable this only in failpoint.

lance6716:
@GMHDBJD PTAL

PR #875, opened 2020-08-07
dumpling: fix args config

### LINKED ISSUE #862: some args set in `mydumper` don't take effect
## Bug Report

   specify `statement-size` in `mydumper` config section.

   `statement-size` takes effect.

3. What did you see instead?

   `statement-size` does not take effect, and can only be specified in `extra-args` to take effect.

4. Versions of the cluster

        ```
       v2.0.0-beta.2
        ```

GMHDBJD:
close #862 

- Simply use previous config value as default value of flagset. May be we should change export config of dumpling to support flagset directly.
- rename no-sharding.yaml in shardddl1 to make sure test case enter sync stage.

Tests 

 - Unit test


lance6716:
LGTM

Issue #874, opened 2020-08-07
Syncer convert binlog event to sql bit column error


sultan8252:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do? If possible, provide a recipe for reproducing the error.

2. What did you expect to see?

3. What did you see instead?
![image](https://user-images.githubusercontent.com/10906273/89626813-884e9200-d8cc-11ea-9236-8314cfb65807.png)

4. Versions of the cluster

5. current status of DM cluster (execute `query-status` in dmctl)

6. Operation logs
   - Table schema
``` CREATE TABLE `cash_withdraw_request` (
  `id` int(10) unsigned NOT NULL AUTO_INCREMENT,
  `amount` decimal(12,2) NOT NULL COMMENT '提现申请金额',
  `applyTime` datetime NOT NULL COMMENT '提现申请时间',
  `fee` decimal(8,2) NOT NULL COMMENT '手续费',
  `handleTime` datetime DEFAULT NULL COMMENT '处理此请求的时间',
  `resultTIme` datetime DEFAULT NULL COMMENT '请求处理后接收时间',
  `bankCardNo` varchar(25) NOT NULL COMMENT '申请提现的银行卡号',
  `bankCardBranch` varchar(64) NOT NULL COMMENT '申请提现的银行卡的发卡行',
  `userId` int(10) unsigned NOT NULL COMMENT '申请用户id',
  `status` smallint(5) unsigned NOT NULL COMMENT '状态 UserDef.CASH_WITHDRAW_*',
  `operationComment` varchar(64) DEFAULT NULL COMMENT '操作注释，供后台人员使用',
  `bankCode` varchar(5) DEFAULT NULL COMMENT '申请提现的银行卡所在银行的代码，见bank_code.sql',
  `accountBankCity` varchar(64) DEFAULT NULL COMMENT '开户行城市',
  `accountBankId` varchar(12) DEFAULT NULL COMMENT '开户行行号，如001121004527',
  `isOutput` bit(1) DEFAULT b'0' COMMENT '是否已制表导出，默认是false(未导出)',
  `runUpAccount` decimal(14,2) unsigned NOT NULL DEFAULT 0.00 COMMENT '挂账金额',
  `freeWithdrawQuota` decimal(14,2) unsigned NOT NULL DEFAULT 0.00 COMMENT '免费提现配额',
[...]

PR #876, opened 2020-08-07
syncer: fix wrong conversion for mysql BIT

lance6716:
close https://github.com/pingcap/dm/issues/874

BIT type support up to 64 bit so returned int64 in binlog (I guess), don't need to convert to string

**Need to know why choose convert to string orginally** 🤔 

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

Related changes


lance6716:
this change was introduced in https://github.com/pingcap/tidb-enterprise-tools/pull/5/commits/300bfa154021daec805827c7410d0233e6d33306 , seems avoiding overflow. But now it won't overflow.

BTW, jenkins can't open now

Issue #878, opened 2020-08-10
uniform handle for `ansi-quotes` and `ANSI_QUOTES` `sql_mode` in `session` for task config


csuzhangxc:
## Feature Request

**Is your feature request related to a problem? Please describe:**

if we set `ANSI_QUOTES` `sql_mode` in `session` variable, we may need `ansi-quotes` for parser to parse SQLs from upstream and downstream.

**Describe the feature you'd like:**

remove `ansi-quotes` or set it automatically if `ANSI_QUOTES` `sql_mode` in `session` set? but I'm not sure now

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

PR #882, opened 2020-08-11
pkg/binlog: parse statusVars and get ansi-quotes mode

lance6716:
part of #878 , enable DM to get ansi-quotes mode more automaticly, so in future we could remove ansi-quotes in settings

only added some function. `GetAnsiQuotesMode` or `statusVarsToKV` could be used on `replication.QueryEvent.StatusVars` in future

`statusVarsToKV` parses `replication.QueryEvent.StatusVars`, note that the order of KV sequence is undetermined (at least unordered I noticed).

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes


pkg/binlog/event/util.go
```diff
@@ -269,3 +271,170 @@ func writeStringColumnValue(buf *bytes.Buffer, value interface{}) error {
...
+			vars[key] = value
+			continue
+		}
+
+		// get variable-length value of according key and save it in `value`
+		switch key {
+		// 1-byte length + <length> chars of the catalog + '0'-char
+		case QCatalog:
+			if err = appendLengthThenCharsToValue(); err != nil {
+				return generateError()
```
csuzhangxc, line 391: can we carry `err` in the returned error?

lance6716: changed in https://github.com/pingcap/dm/pull/882/commits/7178f200203c5a86b14a2ea8009a6543bf27e990

csuzhangxc:
How about also cherrying pick this to _release-1.0_ now?

lance6716:
> How about also cherrying pick this to _release-1.0_ now?

now or after PR merging?

csuzhangxc:
after PR merged. I mean cherry-pick it even it's just a part of #878.

PR #870, opened 2020-08-07
precheck: support primary/unique key check

GMHDBJD:
### What problem does this PR solve? 
close #788 

### What is changed and how it works?
update tidb-tools and fix some tests

Tests 
 - Manual test (add detailed scripts or steps below)

PR #887, opened 2020-08-12
precheck: support primary/unique key check #870

GMHDBJD:
manually cherrypick #870 

update tidb-tools and fix some tests


lance6716:
how about test `ignore-checking-items` could work

rest LGTM

GMHDBJD:
> how about test ignore-checking-items could work

The origin commit failed because doesn't have primary key. I think it's enough. 🤣

lance6716:
LGTM

PR #888, opened 2020-08-12
pkg/binlog: parse statusVars and get ansi-quotes mode (#882)

lance6716:
manually cherry-pick #882

---

part of #878 , enable DM to get ansi-quotes mode more automaticly, so in future we could remove ansi-quotes in settings

only added some function. `GetAnsiQuotesMode` or `statusVarsToKV` could be used on `replication.QueryEvent.StatusVars` in future

`statusVarsToKV` parses `replication.QueryEvent.StatusVars`, note that the order of KV sequence is undetermined (at least unordered I noticed).

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes


pkg/terror/error_list.go
```diff
@@ -668,6 +684,7 @@ var (
 	ErrBinlogNeedMariaDBGTIDSet      = New(codeBinlogNeedMariaDBGTIDSet, ClassFunctional, ScopeInternal, LevelHigh, "the event should be a MariadbGTIDListEvent, but got %T", "")
 	ErrBinlogParseMariaDBGTIDSet     = New(codeBinlogParseMariaDBGTIDSet, ClassFunctional, ScopeInternal, LevelHigh, "parse MariaDB GTID set", "")
 	ErrBinlogMariaDBAddGTIDSet       = New(codeBinlogMariaDBAddGTIDSet, ClassFunctional, ScopeInternal, LevelHigh, "add set %v to GTID set", "")
+	ErrBinlogStatusVarsParse         = New(codeBinlogStatusVarsParse, ClassFunctional, ScopeInternal, LevelMedium, "fail to parse binglog status_vars: %v, offset: %d", "")
```
GMHDBJD, line 687: Move to the end of `Config related error code list`, otherwise the position in errors_release.txt is strange.

lance6716: done in https://github.com/pingcap/dm/pull/888/commits/9383d2401fb8e6db8a8e1b7c201e05b87181eea8

pkg/terror/error_list.go
```diff
@@ -152,6 +152,22 @@ const (
...
+	codeConnRegistryTLSConfig
+
+	// pkg/upgrade
+	codeUpgradeVersionEtcdFail
+
+	// pkg/v1workermeta
+	codeInvalidV1WorkerMetaPath
+
+	// pkg/v1dbschema
+	codeFailUpdateV1DBSchema
```
csuzhangxc, line 167: How about coping `Err**` for these codes too?

lance6716: seems not used in v1.x.x. Does it break error code compatibility?

csuzhangxc: if codes exist, it doesn't break compatibility. but if we copied them, may easier for cherry-picking other PRs later?

lance6716: done in https://github.com/pingcap/dm/pull/888/commits/105acfb0144a55172c7d8fa913401a7c412135a4

PR #854, opened 2020-08-03
ctl: support using task-file when need task-name

lance6716:
### What problem does this PR solve? 
part of #826

### What is changed and how it works?
support using task-file when need task-name

### Check List 

Tests 

 - Integration test

Code changes

 - Has exported function/method change

Side effects

Related changes

 - Need to cherry-pick to the release branch
 - Need to update the documentation
 - Need to be included in the release note

Issue #826, opened 2020-07-23
both config file and name support for source and task operations


csuzhangxc:
## Feature Request

**Is your feature request related to a problem? Please describe:**

now, for task operations, `start-task` accepts `task.yaml` file but `pause-task`, `resume-task` ... accept _task name_.

for source operations, only `source.yaml` file accepted.

**Describe the feature you'd like:**

add both config file and name (or _source ID_) support for source and task operations.

but only config file supported for the following two commands (because we must have the whole file):

- `start-task`
- `operate-source create`

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

PR #892, opened 2020-08-13
ctl: support using task-file when need task-name (#854)

lance6716:
manually cherry-pick #854 

---

part of #826

support using task-file when need task-name

### Check List 

Tests 

 - Integration test

Code changes

Side effects

Related changes


dm/ctl/common/util.go
```diff
@@ -210,3 +211,22 @@ func IsDDL(sql string) (bool, error) {
 		return false, nil
 	}
 }
+
+// GetTaskNameFromArgOrFile tries to retrieve name field from the file if arg is filename-like, otherwise returns arg directly
+func GetTaskNameFromArgOrFile(arg string) string {
+	if !strings.HasSuffix(arg, ".yaml") {
```
csuzhangxc, line 217: some people may name a YAML file with the suffix of `yml` rather than `yaml`. 😂

lance6716:
> LGTM. do we also need to update the doc of v1.0, right?

https://github.com/pingcap/docs-dm/pull/220 will cherry-pick to 1.0

dm/ctl/common/util.go
lance6716: done in https://github.com/pingcap/dm/pull/892/commits/6d1aede47719764e2eac05009255833c87f349c2

csuzhangxc: we also need to update the master branch now.

PR #894, opened 2020-08-14
fix: redirect stream when get last replace

GMHDBJD:
When replace a DDL into multiple DDLs, eg:
```
1000-1010 --> 1000.0-1000.1, 1000.1-1010.0
```
When we redirect to 1000.1 in resharding, after get replace event, stream still at 1000. It will get origin event again

redirect stream when get last replace event.

### Check List 

Tests 

 - Integration test


syncer/syncer.go
```diff
@@ -1331,11 +1331,21 @@ func (s *Syncer) Run(ctx context.Context) (err error) {
...
-		// we calculate endLocation(currentLocation) for Rows/Query event here
+		// we calculate startLocation and endLocation(currentLocation) for Rows/Query event here
+		// set startLocation empty for other events to avoid misuse
+		startLocation = binlog.Location{}
 		switch e.Event.(type) {
 		case *replication.RowsEvent, *replication.QueryEvent:
+			startLocation = binlog.Location{
+				Position: mysql.Position{
+					Name: lastLocation.Position.Name,
+					Pos:  e.Header.LogPos - e.Header.EventSize,
```
lance6716, line 1343: can't use lastLocation?

lance6716:
we might list these XXXLocations' meaning in comment, with some invarant properties attached to them (such as "at beginning of for-loop, events before XXXLocation event has been synced" )

lance6716:
(please help cherry-pick this https://github.com/pingcap/dm/commit/6d1aede47719764e2eac05009255833c87f349c2 to this PR) 😂

syncer/syncer.go
GMHDBJD: More safe. 😂

GMHDBJD:
> we might list these XXXLocations' meaning in comment, with some invarant properties attached to them (such as "at beginning of for-loop, events before XXXLocation event has been synced" )

I update the comment in the beginning. But most comments are still scattered in different places.

lance6716:
advice some renaming:

```
currentLocation = s.checkpoint.GlobalPoint() // also init to global checkpoint
```
rename it to `nextLocation`, aware there's a `nextLocation`around line 1212, may need rename it to `...2`

--

```
shardingReSync.currLocation
```
ditto

--

```
		s.currentLocationMu.Lock()
		s.currentLocationMu.currentLocation = currentLocation.Clone()
		s.currentLocationMu.Unlock()
```
this is indeed the start location of "current" event, and we use `startLocation` representing "start location of current event". So how about a) renaming it to `s.startLocationMu` or b) renaming `startLocation` to `currentLocation`

And here we forward the "current", how about change "next" location also. So later code could reference a clear variable

--

```
e, err = s.getEvent(tctx, &currentLocation)
```
maybe `s.getEvent(tctx, &s.startLocationMu.startLocation)` (depend on above renaming) to make it clear?

--

```
startLocation = binlog.Location{}
...
			startLocation = binlog.Location{
				Position: mysql.Position{
					Name: lastLocation.Position.Name,
					Pos:  e.Header.LogPos - e.Header.EventSize, // <- e is current event
```
in fact it's `prevLocation` right?

--

also align them with `eventContext`, `ShardingReSync` structure

lance6716:
LGTM

PR #895, opened 2020-08-14
*: fix test and yaml suffix

lance6716:
~~fix test TestTaskAutoResume not waiting so long w.r.t checking interval~~ and adding another yaml suffix

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

Related changes


GMHDBJD:
LGTM

Issue #897, opened 2020-08-17
do not auto resume for "Column count doesn't match value count" error


csuzhangxc:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do? If possible, provide a recipe for reproducing the error.

2. What did you expect to see?

    do not auto-resume for "Column count doesn't match value count" error.

    ```
    "errors": [
        {
            "ErrCode": 36027,
            "ErrClass": "sync-unit",
            "ErrScope": "internal",
            "ErrLevel": "high",
            "Message": "startLocation: [position: (mysql-bin.000001, 17491), gtid-set: ], endLocation: [position: (mysql-bin.000001, 17535), gtid-set: ]: gen insert sqls failed, schema: db_single, table: t1: Column count doesn't match value count: 3 (columns) vs 2 (values)",
             "RawCause": "",
             "Workaround": ""
         }
    ],
    ```

3. What did you see instead?

    ```
    [2020/08/14 17:38:03.235 +08:00] [ERROR] [subtask.go:295] ["unit process error"] [subtask=task_single] [unit=Load] ["error information"="{\"ErrCode\":10006,\"ErrClass\":\"database\",\"ErrScope\":\"downstream\",\"ErrLevel\":\"high\",\"Message\":\"file db_single.t1.0.sql: execute statement failed: INSERT INTO `t1` VALUES\\n(1,1),\\n(2,2),\\n(3,3),\\n(4,4),\\n(5,5),\\n(6,6),\\n(7,7),\\n(8,8),\\n(9,9),\\n(10,10);\",\"RawCause\":\"Error 1136: Column count doesn't match value count at row 1\"}"]
    [2020/08/14 17:38:03.958 +08:00] [INFO] [server.go:507] [request=QueryStatus] [payload="name:\"task_single\" "]
    [2020/08/14 17:38:04.260 +08:00] [WARN] [task_checker.go:342] ["backoff skip auto resume task"] [component="task checker"] [task=task_single] [latestResumeTime=2020/08/14 17:38:04.260 +08:00] [duration=1s]
    [2020/08/14 17:38:09.260 +08:00] [INFO] [worker.go:288] ["auto_resume sub task"] [component="worker controller"] [task=task_single]
    [2020/08/14 17:38:09.260 +08:00] [INFO] [subtask.go:482] ["resume with unit"] [subtask=task_single] [unit=Load]
[...]

Issue #856, opened 2020-08-03
disable some unsupported commands in dmctl


csuzhangxc:
## Feature Request

**Is your feature request related to a problem? Please describe:**

**Describe the feature you'd like:**

some commands like `pause-relay`, `sql-skip` are not supported in DM v2.0 now, it's better to remove them from _dmctl_'s help message (and we can add them back one by one if supported later).

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

PR #901, opened 2020-08-19
dmctl: disable unsupported command

GMHDBJD:
#856 #897 

- disable currently not supported command in dmctl
- do not auto-resume for DMLColumnNotMatch error

### Check List 

Tests 
 - Integration test


lance6716:
Could DMLColumnNotMatch be resolved by some downstream operations? If so, we'd better leave DMLColumnNotMatch auto-resumable

csuzhangxc:
> Could DMLColumnNotMatch be resolved by some downstream operations? If so, we'd better leave DMLColumnNotMatch auto-resumable

This usually means that something unexpected happened. and in some supported cases, the user still needs to use `operate-schema` to fix it.

csuzhangxc (commented):
LGTM

lance6716:
LGTM

lance6716 (approved):

csuzhangxc:
@GMHDBJD `needs-update-docs`.

PR #906, opened 2020-08-20
feat: support to dump data with column name

csuzhangxc:

dump date with column name, so we can support the case:
- more columns exist in downstream than upstream
- different column order between downstream and upstream

- update dumpling
- set `CompleteInsert` to `true`

data dumped before

```sql
INSERT INTO VALUES
(1,1),
(2,2),
(3,3);
```

data dumped now

```sql
INSERT INTO `t1` (`c1`,`c2`) VALUES
(1,1),
(2,2),
(3,3),
```

### Check List 

Tests 

    - dump & load data with more columns exist in downstream

Code changes

 - Has persistent data change

Related changes

### DIFF
--- a/dumpling/dumpling.go
+++ b/dumpling/dumpling.go
@@ -212,6 +212,7 @@ func (m *Dumpling) constructArgs() (*export.Config, error) {
 	}
 	dumpConfig.TableFilter = tableFilter
 	dumpConfig.EscapeBackslash = true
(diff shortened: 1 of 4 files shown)


lance6716:
LGTM

GMHDBJD (approved):
LGTM. Should we support it in extra args?

csuzhangxc:
> LGTM. Should we support it in extra args?

It seems have no side effect, so I think always enable it is fine.

PR #911, opened 2020-08-21
feat: support `get-task-config` for tasks imported from v1.0.x

csuzhangxc:

- support `get-task-config` for tasks imported from v1.0.x.

- merge subtask configs into task config when importing from v1.0.x

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

 - Has persistent data change

### DIFF
--- a/dm/master/bootstrap.go
+++ b/dm/master/bootstrap.go
@@ -19,6 +19,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"sort"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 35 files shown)


dm/master/bootstrap.go
```diff
@@ -369,6 +372,19 @@ outerLoop:
 				}
 			}
 		}
+		sort.Slice(cfgLists, func(i, j int) bool {
+			return cfgLists[i].SourceID < cfgLists[j].SourceID
+		})
+		var mergedCfg config.TaskConfig
+		err2 := mergedCfg.FromSubTaskConfigs(cfgLists...)
```
lance6716, line 379: seems `FromSubTaskConfigs` doesn't have a fail return

csuzhangxc: 👍 removed return value in https://github.com/pingcap/dm/pull/911/commits/7eb2f69decff28977614ca70da29f165d69d70db.

lance6716:
LGTM

PR #923, opened 2020-08-24
test: move status check later

lance6716:
part of https://github.com/pingcap/dm/issues/773, fix `full_mode`

log shows both worker has encountered dumpling error, but sometimes the result of RPC lack one failure, so I guess sometimes RPC returns too fast.

remove the immediately check, later check could has same promise.

### Check List 

Tests 

 - Integration test

Code changes

Side effects

Related changes

### DIFF
--- a/tests/full_mode/run.sh
+++ b/tests/full_mode/run.sh
@@ -40,8 +40,7 @@ function fail_acquire_global_lock() {
     cp $cur/conf/dm-task.yaml $WORK_DIR/dm-task.yaml
     sed -i '/timezone/i\ignore-checking-items: ["dump_privilege"]' $WORK_DIR/dm-task.yaml
     run_dm_ctl $WORK_DIR "127.0.0.1:$MASTER_PORT" \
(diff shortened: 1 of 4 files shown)


tests/full_mode/run.sh
```diff
@@ -40,8 +40,7 @@ function fail_acquire_global_lock() {
     cp $cur/conf/dm-task.yaml $WORK_DIR/dm-task.yaml
     sed -i '/timezone/i\ignore-checking-items: ["dump_privilege"]' $WORK_DIR/dm-task.yaml
     run_dm_ctl $WORK_DIR "127.0.0.1:$MASTER_PORT" \
-        "start-task $WORK_DIR/dm-task.yaml --remove-meta" \
-        "\"result\": true" 1
+        "start-task $WORK_DIR/dm-task.yaml --remove-meta"
```
csuzhangxc, line 43: Do `start-task` check in other cases have the same problem?

lance6716: seems other are "successful start" or "failed at init stage"

PR #933, opened 2020-08-26
Fix auto resume

lance6716:
test fails because injected failpoint only fail for 2 times, when test runs slowly, that's not enough

keep fails until we check, then disable it

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes

### DIFF
--- a/dm/worker/worker_test.go
+++ b/dm/worker/worker_test.go
@@ -118,12 +118,12 @@ func (t *testServer) TestTaskAutoResume(c *C) {
 
 	c.Assert(failpoint.Enable("github.com/pingcap/dm/dumpling/dumpUnitProcessForever", `return(true)`), IsNil)
 	defer failpoint.Disable("github.com/pingcap/dm/dumpling/dumpUnitProcessForever")
-	c.Assert(failpoint.Enable("github.com/pingcap/dm/dumpling/dumpUnitProcessWithError", `2*return("test auto resume inject error")`), IsNil)
-	defer failpoint.Disable("github.com/pingcap/dm/dumpling/dumpUnitProcessWithError")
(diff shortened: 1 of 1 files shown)


dm/worker/worker_test.go
```diff
@@ -118,13 +118,13 @@ func (t *testServer) TestTaskAutoResume(c *C) {
...
 	c.Assert(failpoint.Enable("github.com/pingcap/dm/dumpling/dumpUnitProcessForever", `return(true)`), IsNil)
 	defer failpoint.Disable("github.com/pingcap/dm/dumpling/dumpUnitProcessForever")
-	c.Assert(failpoint.Enable("github.com/pingcap/dm/dumpling/dumpUnitProcessWithError", `2*return("test auto resume inject error")`), IsNil)
-	defer failpoint.Disable("github.com/pingcap/dm/dumpling/dumpUnitProcessWithError")
 	c.Assert(failpoint.Enable("github.com/pingcap/dm/dm/worker/mockCreateUnitsDumpOnly", `return(true)`), IsNil)
 	defer failpoint.Disable("github.com/pingcap/dm/dm/worker/mockCreateUnitsDumpOnly")
 	c.Assert(failpoint.Enable("github.com/pingcap/dm/loader/ignoreLoadCheckpointErr", `return()`), IsNil)
 	defer failpoint.Disable("github.com/pingcap/dm/loader/ignoreLoadCheckpointErr")
 
+	c.Assert(failpoint.Enable("github.com/pingcap/dm/dumpling/dumpUnitProcessWithError", `return("test auto resume inject error")`), IsNil)
```
GMHDBJD, line 126: defer failpoint.Disable() here for exception-safe.

lance6716: in https://github.com/pingcap/dm/pull/933/commits/bd37295f72355b77c8cfdf7923d85b54d2345a7a

PR #945, opened 2020-08-28
basedb: close db when fail ping (#942)

lance6716:
manually cherry-pick #942

---

DB connection leak when successful open DB but failed ping.

close it manually. **alternative is we should return `db, error` to caller instead of `nil, error`, so caller's rollback function could handle closing DB**.

### Check List 

Tests 

 - unit test

Code changes

Side effects

Related changes


csuzhangxc (approved):
LGTM

PR #948, opened 2020-08-28
fix: handle `null` table_info after importing from v1.0.x

csuzhangxc:

    `gen delete sqls failed, schema: import_v10x, table: t2: Column count doesn't match value count: 0 (columns) vs 2 (values)` reported after importing from v1.0.6.

   do not create an empty table if `table_info` column is `null`

### Check List 

Tests 

 - Unit test
 - Integration test

Related changes


lance6716:
how did test generate "null" tableinfo 🤔

GMHDBJD (commented):
LGTM

csuzhangxc:
> how did test generate "null" tableinfo 🤔

add `table_info` column will generate `null` value with json type.

lance6716:
LGTM

lance6716 (approved):

PR #915, opened 2020-08-21
*: use dumpling's finish building connection location to leave safe mode

lance6716:
### What problem does this PR solve? 
close #907

### What is changed and how it works?

when dump unit can't assure consistency (for example migrate Aurora, or maybe we could support more scenario to avoid FTWRL), between start and ending of dumping's creating connections, binlog maybe duplicated with dump result, so we should enable safe mode in this period. originally safe mode was turned on for 5 minutes, this PR assure safe mode will keep for enough time.

**maybe we should save exit location to checkpoint table, to handle worker switch before exit safe mode. This will achieve in another PR including updating strategy**

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

 - Increased code complexity

Related changes

 - Need to cherry-pick to the release branch
 - Need to update the documentation
 - Need to be included in the release note

PR #988, opened 2020-09-03
*: save exitSafeModeLoc in checkpoint, load it, upgrade strategy

lance6716:

save `ExitSafePointLoc` in #915 to checkpoint

change structure of syncer checkpoint, add upgrade strategy

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity
 - Breaking backward compatibility (but with upgrade strategy)

Related changes

  support upstream no lock migration, such as from Aurora


syncer/syncer_test.go
```diff
@@ -1471,6 +1482,34 @@ func (s *testSyncerSuite) TestExitSafeModeByConfig(c *C) {
 	c.Assert(failpoint.Disable("github.com/pingcap/dm/syncer/SafeModeInitPhaseSeconds"), IsNil)
 }
 
+func (s *testSyncerSuite) TestRemoveMetadataIsFine(c *C) {
```
lance6716, line 1485: please help me think more senarios LoadMeta will return metadata file can't found

loader/loader.go
```diff
@@ -1274,8 +1275,12 @@ func (l *Loader) getMydumpMetadata() error {
 	metafile := filepath.Join(l.cfg.LoaderConfig.Dir, "metadata")
 	loc, _, err := dumpling.ParseMetaData(metafile, l.cfg.Flavor)
 	if err != nil {
+		toPrint, err := ioutil.ReadFile(metafile)
```
csuzhangxc, line 1278: Do we need to keep the original `err` returned from `dumpling.ParseMetaData`? if so, ditto for another call to `dumpling.ParseMetaData`.

pkg/upgrade/upgrade.go
```diff
@@ -13,18 +13,47 @@
...
+// NOTE that zero value of Context is nil, be aware of nil-dereference
+type Context struct {
+	context.Context
+	SubTaskConfigs map[string]map[string]config.SubTaskConfig
+}
+
+// NewUpgradeContext creates a Context, avoid nil Context member
+func NewUpgradeContext() Context {
+	return Context{
+		Context: context.Background(),
```
csuzhangxc, line 50: Should we need to `make` a `map` for it?

pkg/upgrade/upgrade.go
```diff
@@ -65,6 +94,66 @@ func TryUpgrade(cli *clientv3.Client) error {
 
 // upgradeToVer1 does upgrade operations from Ver0 to Ver1.
 // in fact, this do nothing now, and just for demonstration.
-func upgradeToVer1(cli *clientv3.Client) error {
+func upgradeToVer1(cli *clientv3.Client, uctx Context) error {
+	return nil
+}
+
+// upgradeToVer2 does upgrade operations from Ver1 to Ver2 (v2.0.0-rc3) to upgrade syncer checkpoint schema
```
csuzhangxc, line 101:
```suggestion
// upgradeToVer2 does upgrade operations from Ver1 to Ver2 (v2.0.0-rc.3) to upgrade syncer checkpoint schema
```

maybe we release `v2.0.0` GA next time? 😄

pkg/upgrade/upgrade.go
```diff
@@ -65,6 +94,66 @@ func TryUpgrade(cli *clientv3.Client) error {
...
+func upgradeToVer2(cli *clientv3.Client, uctx Context) error {
+	upgradeTaskName := "upgradeToVer2"
+	logger := log.L().WithFields(zap.String("task", upgradeTaskName))
```
csuzhangxc, line 104: `task` has a special meaning in DM (and you also use it in L118 below, in order to match `ExecuteSQLWithIgnoreError`?). How about giving it another name like `job`/`op`...

pkg/upgrade/upgrade.go
```diff
@@ -65,6 +94,66 @@ func TryUpgrade(cli *clientv3.Client) error {
...
+
+	// tableName -> DBConfig
+	dbConfigs := map[string]config.DBConfig{}
+	for task, m := range uctx.SubTaskConfigs {
+		for sourceID, subCfg := range m {
+			tableName := dbutil.TableName(subCfg.MetaSchema, cputil.SyncerCheckpoint(subCfg.Name))
+			subCfg2, err := subCfg.DecryptPassword()
+			if err != nil {
+				log.L().Error("skip subconfig when upgrading", zap.String("task", task),
+					zap.String("source id", sourceID), zap.Error(err))
```
csuzhangxc, line 119: Should we `continue` here?

pkg/upgrade/upgrade.go
```diff
@@ -65,6 +94,66 @@ func TryUpgrade(cli *clientv3.Client) error {
...
+	defer func() {
+		for _, db := range toClose {
+			db.Close()
+		}
+	}()
+	for tableName, cfg := range dbConfigs {
+		targetDB, err := conn.DefaultDBProvider.Apply(cfg)
+		if err != nil {
+			logger.Error("skip target DB when upgrading", zap.String("table name", tableName))
+			return err
```
csuzhangxc, line 135: `return` rather than `continue`? if so, how about `return` after L119? if `continue`, log `err` in L134.

syncer/checkpoint.go
```diff
@@ -258,6 +266,8 @@ type RemoteCheckPoint struct {
 	globalPoint         *binlogPoint
 	globalPointSaveTime time.Time
 
+	safeModeExitPoint *binlog.Location
```
csuzhangxc, line 269: How about adding some comments to explain?
- when it will be set and flush
- when it will be unset and flush
- when it will be used to judge

pkg/upgrade/upgrade.go
lance6716: I think internal upgrading is semantic suitable for a task, though other tasks are user specifed in task config. for example, user may see many "task name" running in grafana when upgrade, but in fact that's a internal upgrading not many task running.

and `updateSyncerCheckpoint` of `v1dbschema` uses task name from user's task config, maybe I could change that

loader/loader.go
lance6716: https://github.com/pingcap/dm/pull/988/commits/4f4be9b30ede7f88459670a4de519ff660fab640

pkg/upgrade/upgrade.go
lance6716: https://github.com/pingcap/dm/pull/988/commits/556ffaf48bbaeb9e8868e573908f810bfadb37fb

pkg/upgrade/upgrade.go
lance6716: choose `return`

https://github.com/pingcap/dm/pull/988/commits/fa401af29be9b3bbb7bf5e3cfd6bd7d96f78c800

pkg/upgrade/upgrade.go
csuzhangxc: > and updateSyncerCheckpoint of v1dbschema uses task name from user's task config, maybe I could change that

OK. 👍

lance6716: https://github.com/pingcap/dm/pull/988/commits/e618c7682835a60c6475ae9945cf7ef5b0ed651b

syncer/checkpoint.go
lance6716: https://github.com/pingcap/dm/pull/988/commits/d4705aa46cdda2d136f9812dd22a146d231fa2cc

pkg/upgrade/upgrade.go
lance6716: https://github.com/pingcap/dm/pull/988/commits/3ee0d8c0c3075ff5b5c2cef45e38fbf87a61238c

csuzhangxc (commented):
LGTM

GMHDBJD (approved):
LGTM

PR #899, opened 2020-08-18
fix: TLS support for dumpling and syncer in v2.0

GMHDBJD:
### What problem does this PR solve? 
close #880 

### What is changed and how it works?
- set tls for dumpling
- set tls for streamController

### Check List 

Tests 
 - Manual test (add detailed scripts or steps below)
    start a mysql server with tls config, run an all mode task

PR #1008, opened 2020-09-08
tls: minor fix on http api

WangXiangUSTC:

when setting `cert-allowed-cn`, will get an error  when sending an HTTP request with GRPC gateway, the error log looks like:
```
transport: authentication handshake failed: client certificate authentication failed. The Common Na
me from the client certificate [] was not found in the configuration cluster-verify-cn with value: [dm]`
```

fix it

### Check List 

Tests 

 - Integration test
 - Manual test


csuzhangxc:
@WangXiangUSTC can you explain your fix in the PR's comment? 😂

dm/master/server.go
```diff
@@ -170,11 +170,8 @@ func (s *Server) Start(ctx context.Context) (err error) {
 	if err != nil {
 		return terror.ErrMasterTLSConfigNotValid.Delegate(err)
 	}
-	if tls2 != nil && tls2.TLSConfig() != nil {
-		tls2.TLSConfig().InsecureSkipVerify = true
```
WangXiangUSTC, line 174: if `InsecureSkipVerify` is true, the  argument `verifiedChains` in `VerifyPeerCertificate` will be nil, so will return error here https://github.com/pingcap/tidb-tools/blob/03575b185007b81384dc0728020cf5a95cd28f9e/pkg/utils/security.go#L99

more details refer to https://golang.org/src/crypto/tls/common.go?s=17676:27385#L556

dm/master/server.go
```diff
@@ -170,11 +170,8 @@ func (s *Server) Start(ctx context.Context) (err error) {
...
-	}
 
-	apiHandler, err := getHTTPAPIHandler(ctx, s.cfg.MasterAddr, tls2.ToGRPCDialOption())
```
WangXiangUSTC, line 177: Sometimes `MasterAddr` is not an IP address tls will return error `transport: authentication handshake failed: tls: either ServerName or InsecureSkipVerify must be specified in the tls.Config`. And `AdvertiseAddr`'s format looks like `127.0.0.1:8261`, and tls will set `ServerName` to `127.0.0.1` here https://golang.org/src/crypto/tls/tls.go?h=ServerName#L162

csuzhangxc: `MasterAddr` should be the bind address of the dm-master, it should be an IP address, right?

`AdvertiseAddr` can be a domain.

csuzhangxc (commented):
LGTM

dm/master/server.go
lance6716: If don't set `InsecureSkipVerify`, "client verifies the server's certificate chain and host name". Did we need extra certs to pass this verify? @lichunzhu

lichunzhu: No, I don't think so. Why will we need extra certs?

lance6716: > No, I don't think so. Why will we need extra certs?

Because we set `InsecureSkipVerify` before this PR, I think that behaviour should come from some reason, such as we can't set CA or something

dm/master/server.go
lance6716: what's the required format of `MasterAddr` and `AdvertiseAddr`, might `adjust` them in advance. Currently we will set `AdvertiseAddr` to `MasterAddr` if empty

dm/master/server.go
lichunzhu: Actually from the comment this argument should be used only for test. PD doesn't use this argument but can still works. Maybe after this we need to set the correct URL when we publish CA certs.
https://docs.pingcap.com/zh/tidb-in-kubernetes/stable/enable-tls-between-components#%E4%BD%BF%E7%94%A8-cfssl-%E7%B3%BB%E7%BB%9F%E9%A2%81%E5%8F%91%E8%AF%81%E4%B9%A6

GMHDBJD: I also use InsecureSkipVerify [here](https://github.com/pingcap/dm/blob/255a9c2be64a891d45ad62e0c9f228e191f155f2/syncer/syncer.go#L2652) in #899, any idea?

lance6716: > I also use InsecureSkipVerify [here](https://github.com/pingcap/dm/blob/255a9c2be64a891d45ad62e0c9f228e191f155f2/syncer/syncer.go#L2652) in #899, any idea? If not set, binlog streamerController will return same error `transport: authentication handshake failed: tls: either ServerName or InsecureSkipVerify must be specified in the tls.Config`

I found this problem maybe caused by https://github.com/pingcap/tidb-tools/blob/master/pkg/utils/security.go#L73 , `tls.Config` has a `ServerName` field which says 

```
	// ServerName is used to verify the hostname on the returned
	// certificates unless InsecureSkipVerify is given. It is also included
	// in the client's handshake to support virtual hosting unless it is
	// an IP address.
```

dm/master/server.go
lance6716: Seems we should also set `tls.Config.ServerName` to its domain, (though I'm not sure which address of config this domain should came from)

GMHDBJD (approved):
LGTM

lance6716 (approved):

PR #1045, opened 2020-09-16
tls: use tls in list-member

GMHDBJD:
fix master offline in list-member when use tls.

use tls for http client

### Check List 

Tests 

 - Integration test

### DIFF
--- a/dm/master/server.go
+++ b/dm/master/server.go
@@ -1838,9 +1838,16 @@ func (s *Server) listMemberMaster(ctx context.Context, names []string) (*pb.Memb
 
 	etcdMembers := memberList.Members
 	masters := make([]*pb.MasterInfo, 0, len(etcdMembers))
-	client := http.Client{
-		Timeout: 1 * time.Second,
+
+	client := &http.Client{}
+	if len(s.cfg.SSLCA) != 0 {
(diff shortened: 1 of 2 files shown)


dm/master/server.go
```diff
@@ -1838,9 +1838,16 @@ func (s *Server) listMemberMaster(ctx context.Context, names []string) (*pb.Memb
 
 	etcdMembers := memberList.Members
 	masters := make([]*pb.MasterInfo, 0, len(etcdMembers))
-	client := http.Client{
-		Timeout: 1 * time.Second,
+
+	client := &http.Client{}
+	if len(s.cfg.SSLCA) != 0 {
+		inner, err := toolutils.ToTLSConfigWithVerify(s.cfg.SSLCA, s.cfg.SSLCert, s.cfg.SSLKey, s.cfg.CertAllowedCN)
```
lichunzhu, line 1844: Does http client need `CertAllowedCN`? Is `ToTLSConfig` enough?

GMHDBJD: If user set a wrong common name, we can verify it in http client. We also do it when master create etcd grpc client.
Do you think we don't need to verify the common name for http? It's enough though.

lichunzhu (approved):
LGTM

csuzhangxc:
/run-all-tests

`rpc addr 127.0.0.1:8261 check alive timeout`

csuzhangxc (approved):
LGTM

PR #1032, opened 2020-09-14
tracker: get some session variables from downtream

lance6716:
### What problem does this PR solve? 
enable some test about altering PK, and reduce user config. close #1003 

### What is changed and how it works?

- session variable:
  - if user specified, use it
  - else get `sql_mode`, `tidb_skip_utf8_check` from downstream
- tracker (tidb) config:
  - set `alter-primary-key` true
- max-allow-packet
  - always get from downstream (by feature of driver)

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

 - Has exported function/method change
 - Has exported variable/fields change
 - Has persistent data change

Side effects

 - Breaking backward compatibility

Related changes

 - Need to cherry-pick to the release branch
 - Need to update the documentation
 - Need to be included in the release note

PR #1055, opened 2020-09-17
ddl: pause subtask when exec ddl err

### LINKED ISSUE #984: handle schema tracker newer than downstream TiDB
## Bug Report

update schema tracker in https://github.com/pingcap/dm/pull/982 to enjoy more compatibility, some test failed, such as alter FK without schema.

`query-status` shows nothing

3. What did you see instead?

`query-status` shows that error because downstream TiDB doesn't support

4. Versions of the cluster

       after #982

    - Downstream TiDB cluster version (execute `SELECT tidb_version();` in a MySQL client):

       older than `20200901032733-f82e5320ad75`

GMHDBJD:
close #984 

- report error when execute ddl fail in dowsteam TiDB
- omit error in syncDDL caller to avoid duplicate error message

### Check List 

Tests 

 - Integration test

### DIFF
--- a/syncer/syncer.go
+++ b/syncer/syncer.go
@@ -931,11 +931,18 @@ func (s *Syncer) syncDDL(tctx *tcontext.Context, queueBucket string, db *DBConn,
 			}
 		}
 		if err != nil {
+			s.jobWg.Done()
+			s.execError.Set(err)
+			if !utils.IsContextCanceledError(err) {
+				err = s.handleEventError(err, &sqlJob.startLocation, &sqlJob.currentLocation)
+				s.runFatalChan <- unit.NewProcessError(err)
+			}
 			s.appendExecErrors(&ExecErrorContext{
 				err:      err,
 				location: sqlJob.currentLocation.Clone(),
 				jobs:     fmt.Sprintf("%v", sqlJob.ddls),
 			})
+			continue
 		}
 
 		switch s.cfg.ShardMode {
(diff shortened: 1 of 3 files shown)


syncer/syncer.go
```diff
@@ -931,11 +931,18 @@ func (s *Syncer) syncDDL(tctx *tcontext.Context, queueBucket string, db *DBConn,
 			}
 		}
 		if err != nil {
+			s.jobWg.Done()
```
lance6716, line 934: not sure `s.jobWg.Done()` should be moved later than `s.execError.Set(err)`, to make sure caller will see the error after `s.jobWg.Wait()`

syncer/syncer.go
```diff
@@ -931,11 +931,18 @@ func (s *Syncer) syncDDL(tctx *tcontext.Context, queueBucket string, db *DBConn,
...
+			if !utils.IsContextCanceledError(err) {
+				err = s.handleEventError(err, &sqlJob.startLocation, &sqlJob.currentLocation)
+				s.runFatalChan <- unit.NewProcessError(err)
+			}
 			s.appendExecErrors(&ExecErrorContext{
 				err:      err,
 				location: sqlJob.currentLocation.Clone(),
 				jobs:     fmt.Sprintf("%v", sqlJob.ddls),
 			})
+			continue
```
lance6716, line 945: forward: "can we just rollback checkpoint as we did before?" https://github.com/pingcap/dm/pull/1032#discussion_r489988212

I think that's no need, we could use caller's rollback

lance6716 (commented):
rest LGTM

syncer/syncer.go
GMHDBJD: I move all of it. Actually we now no need `s.execError.get()` in caller. If syncDDL error, subtask will pause and rollback checkpoint in syncer.Process

syncer/syncer.go
GMHDBJD: So no need to change `addjob` in #1032, right?

lance6716:
/run-all-tests

```
[2020-09-17T10:52:23.616Z] FAIL: server_test.go:1293: testMaster.TestOfflineMember
[2020-09-17T10:52:23.616Z] 
[2020-09-17T10:52:23.616Z] server_test.go:1357:
[2020-09-17T10:52:23.616Z]     c.Assert(err, check.IsNil)
[2020-09-17T10:52:23.616Z] ... value *status.statusError = &status.statusError{Code:2, Message:"[code=38043:class=dm-master:scope=internal:level=high], Message: master is not leader, and can't forward request to leader", Details:[]*any.Any(nil), XXX_NoUnkeyedLiteral:struct {}{}, XXX_unrecognized:[]uint8(nil), XXX_sizecache:0} ("rpc error: code = Unknown desc = [code=38043:class=dm-master:scope=internal:level=high], Message: master is not leader, and can't forward request to leader")
```

lance6716:
LGTM

csuzhangxc (approved):
LGTM

PR #1114, opened 2020-09-28
*: improve chaos-mesh actions

csuzhangxc:

improve chaos-mesh tests with GitHub actions.

- enable `pod-kill` action with matrix
- add cron schedule to run chaos tests nighlty
- check failed case faster
- add rand `ALTER TABLE` statements
- add network and IO chaos actions
- add shard DDL test cases (only pessimistic now)

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes


chaos/cases/task.go
```diff
@@ -0,0 +1,395 @@
+// Copyright 2020 PingCAP, Inc.
```
csuzhangxc, line 1: this file copied and refactored from `single.go`.

chaos/cases/cases.go
```diff
@@ -0,0 +1,55 @@
...
+	"golang.org/x/sync/errgroup"
+
+	config2 "github.com/pingcap/dm/dm/config"
+	"github.com/pingcap/dm/dm/pb"
+	"github.com/pingcap/dm/pkg/utils"
+)
+
+var (
+	// NOTE: items in `doSchemas` should be specified in the corresponding task files (`filenames`).
+	// NOTE: can not support optimistic shard DDL now because go-sqlsmith will generated some statements like
```
lance6716, line 29: how about change it to `TODO` to remind us in future

chaos/cases/db.go
```diff
@@ -122,6 +122,8 @@ func ignoreExecSQLError(err error) bool {
 		return true
 	case errno.ErrDupEntry: // HACK: we tolerate `invalid connection`, then `Duplicate entry` may be reported.
 		return true
+	case errno.ErrTooBigRowsize: // HACK:we tolerate `Error 1118: Row size too large. The maximum row size for the used table type, not counting BLOBs, is 65535`
```
lance6716, line 125:
```suggestion
	case errno.ErrTooBigRowsize: // HACK: we tolerate `Error 1118: Row size too large. The maximum row size for the used table type, not counting BLOBs, is 65535`
```

lance6716: is this a connection to MySQL? seems TiDB doesn't have such small length limit

https://docs.pingcap.com/zh/tidb/stable/tidb-limitations

Makefile
```diff
@@ -194,8 +194,8 @@ coverage_fix_cover_mode:
 	sed -i "s/mode: count/mode: atomic/g" $(TEST_DIR)/cov.*.dmctl.*.out
 
 coverage: coverage_fix_cover_mode retool_setup
-	retool do gocovmerge "$(TEST_DIR)"/cov.* | grep -vE ".*.pb.go|.*.__failpoint_binding__.go|.*debug-tools.*|.*portal.*" > "$(TEST_DIR)/all_cov.out"
-	retool do gocovmerge "$(TEST_DIR)"/cov.unit_test*.out | grep -vE ".*.pb.go|.*.__failpoint_binding__.go|.*debug-tools.*|.*portal.*" > $(TEST_DIR)/unit_test.out
+	retool do gocovmerge "$(TEST_DIR)"/cov.* | grep -vE ".*.pb.go|.*.__failpoint_binding__.go|.*debug-tools.*|.*portal.*|.*chaos.*" > "$(TEST_DIR)/all_cov.out"
+	retool do gocovmerge "$(TEST_DIR)"/cov.unit_test*.out | grep -vE ".*.pb.go|.*.__failpoint_binding__.go|.*debug-tools.*|.*portal.*|.*chaos.*" > $(TEST_DIR)/unit_test.out
```
GMHDBJD, line 198: Do we need calculate chaos coverage in unit test?

GMHDBJD (commented):
rest LGTM

chaos/cases/cases.go
csuzhangxc: changed in https://github.com/pingcap/dm/pull/1114/commits/91e5cf704830dbd3fc88d361a660f6b958f43b4d

chaos/cases/db.go
csuzhangxc: applied the suggestion in https://github.com/pingcap/dm/pull/1114/commits/91e5cf704830dbd3fc88d361a660f6b958f43b4d.

> is this a connection to MySQL?

YES

> seems TiDB doesn't have such small length limit

TiDB limited a row to 6MB (https://docs.pingcap.com/zh/tidb/stable/tidb-limitations#%E5%8D%95%E8%A1%8C%E7%9A%84%E9%99%90%E5%88%B6)

Makefile
csuzhangxc: I don't want to add another UT/IT for these testing codes, so I exclude them for coverage.

chaos/scripts/check-case.sh
```diff
@@ -0,0 +1,27 @@
...
+    kubectl wait --for=condition=complete job/chaos-test-case --timeout=1m
+    if [ $? -eq 0 ]; then
+        completed=true
+        echo "chaos-test-case has completed"
+        break
+    else
+        echo "chaos-test-case has not completed" ${i}
+        kubectl get job chaos-test-case
+        if [ $? -ne 0 ]; then
+            echo "chaos-test-case job has been cleared"
```
lance6716, line 14: what's the cause of "cleared" 🤔

chaos/manifests/network-partition-dm.yaml
```diff
@@ -0,0 +1,47 @@
...
+  labels:
+    app: network-partition-dm-master-master
+spec:
+  action: partition
+  mode: one
+  selector:
+    labelSelectors:
+      "app": "dm-master"
+  direction: both
+  target:
```
lance6716, line 40: does this selector (`target` same as above `seletor`) match all dm-master, so each master is isolated from another?

lance6716 (commented):
rest LGTM

chaos/scripts/check-case.sh
csuzhangxc: If the test case job failed, K8s may clear the pod (and also the job) **sometimes** (I don't know how to disable this behavior yet, 😢 )

chaos/manifests/network-partition-dm.yaml
csuzhangxc: > does this selector (target same as above seletor) match all dm-master

YES. but we use `mode: one` here, so only one DM-master member will be isolated with another one (leader or follow, maybe we can change to `mode: all` to isolate the selected one with all other members?).

lance6716: Ah we have 3 master now. ~maybe we could use `fixed-percent` or `random-max-percent`~

I think either `one` or `all` is fine, they will test different scenario 😂  what's your opinion @GMHDBJD

lance6716:
LGTM

chaos/manifests/network-partition-dm.yaml
GMHDBJD: > I think either one or all is fine, they will test different scenario.
Agree. Maybe add more cases later.

GMHDBJD (approved):
LGTM

Issue #998, opened 2020-09-07
test: import unit test coverage


GMHDBJD:
## Feature Request

**Is your feature request related to a problem? Please describe:**

improve unit test coverage
**Describe alternatives you've considered:**

- [ ] github.com/pingcap/dm/checker   0.209s  coverage: 4.4% of statements
- [ ] github.com/pingcap/dm/debug-tools/binlog-event-blackhole        [no test files]
- [ ] github.com/pingcap/dm/dm/command        [no test files]
- [ ] github.com/pingcap/dm/dm/common 0.126s  coverage: 65.6% of statements
- [ ] github.com/pingcap/dm/dm/config 0.478s  coverage: 61.8% of statements
- [ ] github.com/pingcap/dm/dm/ctl    [no test files]
- [ ] github.com/pingcap/dm/dm/ctl/common     [no test files]
- [ ] github.com/pingcap/dm/dm/ctl/master     0.186s  coverage: 3.4% of statements
- [ ] github.com/pingcap/dm/dm/master 72.952s coverage: 61.1% of statements
- [ ] github.com/pingcap/dm/dm/master/shardddl        82.999s coverage: 80.4% of statements
- [ ] github.com/pingcap/dm/dm/master/workerrpc       [no test files]
- [ ] github.com/pingcap/dm/dm/portal 0.260s  coverage: 69.1% of statements
- [ ] github.com/pingcap/dm/dm/portal/statik  [no test files]
- [ ] github.com/pingcap/dm/dm/tracer 0.215s  coverage: 71.0% of statements
- [ ] github.com/pingcap/dm/dm/unit   0.144s  coverage: 0.0% of statements [no tests to run]
- [ ] github.com/pingcap/dm/dm/worker 42.585s coverage: 67.7% of statements
- [ ] github.com/pingcap/dm/dumpling  0.216s  coverage: 52.4% of statements
- [ ] github.com/pingcap/dm/loader    4.170s  coverage: 26.7% of statements
- [ ] github.com/pingcap/dm/mydumper  0.149s  coverage: 39.5% of statements
- [ ] github.com/pingcap/dm/pkg/atomic2       0.085s  coverage: 100.0% of statements
- [ ] github.com/pingcap/dm/pkg/backoff       0.098s  coverage: 100.0% of statements
- [ ] github.com/pingcap/dm/pkg/binlog        0.158s  coverage: 78.6% of statements
[...]

PR #1115, opened 2020-09-28
add dumpling unit test

GMHDBJD:
part of #998

add dumpling ut

### Check List 

Tests 

 - Unit test


dumpling/dumpling_test.go
```diff
@@ -0,0 +1,132 @@
...
+	c.Assert(err, IsNil)
+	resultCh := make(chan pb.ProcessResult, 1)
+
+	var wg sync.WaitGroup
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		dumpling.Process(ctx, resultCh)
+	}()
+	wg.Wait()
```
csuzhangxc, line 90: start a goroutine then wait for it immediately?

dumpling/dumpling.go
```diff
@@ -85,7 +85,6 @@ func (m *Dumpling) Process(ctx context.Context, pr chan pb.ProcessResult) {
 	failpoint.Inject("dumpUnitProcessForever", func() {
 		m.logger.Info("dump unit runs forever", zap.String("failpoint", "dumpUnitProcessForever"))
 		<-ctx.Done()
-		failpoint.Return()
```
lance6716, line 88: though its name is "run forever", I guess it should return to skip below logic here, because in this way so `ProcessResult` is sent so imitate "run forever"

dumpling/dumpling_test.go
```diff
@@ -0,0 +1,132 @@
...
+	c.Assert(result.IsCanceled, IsFalse)
+	c.Assert(len(result.Errors), Equals, 0)
+
+	c.Assert(failpoint.Enable("github.com/pingcap/dm/dumpling/dumpUnitProcessWithError", `return("unknown error")`), IsNil)
+	// nolint:errcheck
+	defer failpoint.Disable("github.com/pingcap/dm/dumpling/dumpUnitProcessWithError")
+
+	// return error
+	wg.Add(1)
+	go func() {
```
lance6716, line 102: ditto

lance6716 (commented):
rest LGTM

dumpling/dumpling.go
GMHDBJD: Dumpling not support context before, so return here. Now we support context.Cancel, below logic will deal with it?

dumpling/dumpling_test.go
```diff
@@ -0,0 +1,122 @@
...
+
+	// cancel
+	var wg sync.WaitGroup
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		dumpling.Process(ctx, resultCh)
+	}()
+	cancel()
+	wg.Wait()
```
csuzhangxc, line 117: the extra goroutine can still be removed?

dumpling/dumpling_test.go
```diff
@@ -0,0 +1,116 @@
...
+		LoaderConfig: config.LoaderConfig{
+			Dir: dir,
+		},
+	}
+	c.Assert(log.InitLogger(&log.Config{}), IsNil)
+}
+
+func (d *testDumplingSuite) TestDumpling(c *C) {
+	dumpling := NewDumpling(d.cfg)
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
```
lance6716, line 77: how about only use `WithTimeout` in third `dumpling.Process`, I'm afraid when CI runs slowly it will fail

lance6716 (commented):
rest LGTM

csuzhangxc (commented):
LGTM

lance6716:
LGTM

lance6716 (approved):

PR #1116, opened 2020-09-29
*: remove dm-tracer

lance6716:
tracer in DM has not been updated to 2.0 for a period, and has not been used in public. we choose to remove it for now

remove related code

### Check List 

Tests 

 - Pass original test

Code changes

Side effects

 - Breaking backward compatibility (But config is not exposed to user)

Related changes


csuzhangxc (commented):
LGTM

GMHDBJD:
Can also remove [this](https://github.com/pingcap/dm/blob/84b12c5fce5d8fddc985909339a837dda7aaa4bf/Makefile#L206-L207)

GMHDBJD (approved):
LGTM

PR #262, opened 2019-09-02
test: add compatibility test

WangXiangUSTC:
### What problem does this PR solve? 

add compatibility test

### What is changed and how it works?
1. for local test
need prepare two version's binary:
bin/dm-master.test.current
bin/dm-worker.test.current
bin/dm-master.test.provious
bin/dm-worker.test.provious
you can get them by `cp bin/dm-master.test  bin/dm-master.test.current`...
and then run `make compatibility_test` to run test case `compatibility`.
If want run other case, run `make compatibility_test CASE={case_name}`, case_name can be `all`.

2. for ci
comment `/run-compatibility-test` to trigger test
the args include `case` and `pre_commit`, `case` means test case, can be file name under `tests`, or empty(will only run case `compatibility`), or `all` to run all the case. `pre_commit` means the previous version's commit, values can be a commit hash, or `HEAD`, `HEAD~5`, `HEAD` is the head of branch master. `pre_commit`'s default value is `HEAD`.

SRE pr: https://github.com/pingcap/SRE/pull/271

### Check List 

Tests 

 - Integration test

PR #1128, opened 2020-10-09
test: bring `run-compatibility-test` back

lance6716:
reveal bug that break upgrading.

compatibility test will run a 1-master-2-worker task, try kill-and-replace master, worker, another worker to check task running well.

default behaviour:
- old master and worker are `v2.0.0-rc`
- only run `compatibility`

apart from https://github.com/pingcap/dm/pull/262, use `/run-compatibility-test break_compatibility=true` to bypass compatibility test if PR require break compatibility

### Check List 

Tests 

 - Integration test

Code changes

Side effects

Related changes


GMHDBJD (commented):
LGTM

csuzhangxc (approved):
LGTM

PR #1134, opened 2020-10-10
loader: fix bug that SQL Mode didn't take effect

lance6716:
fix bug that loader didn't use SQL Mode when create database connections, this bug is not revealed because `all_mode` has manually set SQL Mode related config

if `to` database doesn't have a `sql_mode` item, we use SQL Mode discovered by dumpling's `Init`.

### Check List 

Tests 

 - Integration test

Code changes

Side effects

Related changes

### DIFF
--- a/loader/loader.go
+++ b/loader/loader.go
@@ -460,7 +460,20 @@ func (l *Loader) Init(ctx context.Context) (err error) {
 	dbCfg.RawDBCfg = config.DefaultRawDBConfig().
 		SetMaxIdleConns(l.cfg.PoolSize)
 
-	l.toDB, l.toDBConns, err = createConns(tctx, l.cfg, l.cfg.PoolSize)
+	// used to change loader's specified DB settings, currently SQL Mode
+	lcfg, err := l.cfg.Clone()
+	if err != nil {
+		return err
+	}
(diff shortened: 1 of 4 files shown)


loader/loader.go
```diff
@@ -460,7 +460,20 @@ func (l *Loader) Init(ctx context.Context) (err error) {
...
+	// used to change loader's specified DB settings, currently SQL Mode
+	lcfg, err := l.cfg.Clone()
+	if err != nil {
+		return err
+	}
+	lowerM := map[string]struct{}{}
+	for k := range l.cfg.To.Session {
+		lowerM[strings.ToLower(k)] = struct{}{}
+	}
+	if _, ok := lowerM["sql_mode"]; !ok {
```
csuzhangxc, line 472: Is a `bool` value like `hasSQLMode` enough rather than a `map`? (check `strings.ToLower(k) == sql_mode` in the `for` loop)

lance6716: https://github.com/pingcap/dm/pull/1134/commits/7fb7fdc2aca8695f0878d1ff98d274f1046fb693

csuzhangxc (commented):
LGTM

GMHDBJD (approved):
LGTM

Issue #1130, opened 2020-10-09
set suitable default value for process, source and task config items


csuzhangxc:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do? If possible, provide a recipe for reproducing the error.

2. What did you expect to see?

3. What did you see instead?

we do not set a suitable value for some config items (like `statement-size`, this may dump too large txn to make TiDB OOM)

4. Versions of the cluster

    - DM version (run `dmctl -V` or `dm-worker -V` or `dm-master -V`):

        ```
        DM v2.0.0-rc.2
        ```

csuzhangxc:
dumpling has set the default value of `statement-size` in https://github.com/pingcap/dumpling/pull/160

lance6716:
some default config which maybe dangerous

use dumpling default value: 
- if `mydumper` in config is not null but lack some items:
  - `FileSize` is `UnspecifiedSize`, but seems one big file could have multiple insert statements, will check if it's fine
  - `Rows` is `UnspecifiedSize`, don't split table every `Rows` into multiple files, seems fine because we didn't set it before
  - `SessionParams` is empty, maybe we could make use of it in future
  - `DumpEmptyDatabase`, `NoViews`

overwrite dumpling to DM specified default:
- `EscapeBackslash` is true (what's this)

[WIP]

PR #1142, opened 2020-10-12
dumpling: properly enable EscapeBackslash with upstream SQL mode

lance6716:
part of #1130

because loader will use upstream's SQL mode, if upstream has `NO_BACKSLASH_ESCAPES`, we should not escape `\` in dump files.

### Check List 

Tests 

 - Integration test

Code changes

Side effects

Related changes

### DIFF
--- a/dumpling/dumpling.go
+++ b/dumpling/dumpling.go
@@ -59,6 +59,7 @@ func (m *Dumpling) Init(ctx context.Context) error {
 	var err error
 	m.dumpConfig, err = m.constructArgs()
 	m.detectSQLMode()
+	m.logger.Info("create dumpling", zap.Stringer("config", m.dumpConfig))
 	return err
 }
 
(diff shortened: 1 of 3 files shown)


dumpling/dumpling.go
```diff
@@ -277,4 +276,9 @@ func (m *Dumpling) detectSQLMode() {
 	}
 	m.logger.Info("found upstream SQL mode", zap.String("SQL mode", sqlMode))
 	m.cfg.LoaderConfig.SQLMode = sqlMode
+	if strings.Contains(sqlMode, "NO_BACKSLASH_ESCAPES") {
```
csuzhangxc, line 279: we may also use `HasNoBackslashEscapesMode` in TiDB.

csuzhangxc (commented):
LGTM

dumpling/dumpling.go
lance6716: If it's Aurora upstream, some SQL mode can't be supported using TiDB parser

lichunzhu (approved):
LGTM

PR #1170, opened 2020-10-15
Improve performence

GMHDBJD:
fix performance issue

- improve performance
- fix grafana no data


syncer/syncer.go
```diff
@@ -1671,7 +1671,11 @@ func (s *Syncer) handleRowsEvent(ev *replication.RowsEvent, ec eventContext) err
 }
 
 func (s *Syncer) handleQueryEvent(ev *replication.QueryEvent, ec eventContext) error {
-	sql := strings.TrimSpace(string(ev.Query))
+	query := string(ev.Query)
+	if query == "BEGIN" {
+		return nil
+	}
```
csuzhangxc, line 1677: `bytes.Equal`?

pkg/metricsproxy/summaryvec.go
```diff
@@ -53,7 +53,7 @@ func (c *SummaryVecProxy) WithLabelValues(lvs ...string) prometheus.Observer {
 			labels[c.LabelNames[index]] = label
 		}
 		c.mu.Lock()
-		noteLabelsInMetricsProxy(c, labels)
+		noteLabelsInMetricsProxy(c, labels, lvs)
```
csuzhangxc, line 56: Can we remove above

```
labels := make(map[string]string, len(lvs))
for index, label := range lvs {
	labels[c.LabelNames[index]] = label
}
```

GMHDBJD:
I'm working on location.Clone, will create a new pr because it seems a big change. PTAL for this pr now

pkg/metricsproxy/countervec.go
```diff
@@ -44,12 +48,8 @@ func NewCounterVec(opts prometheus.CounterOpts, labelNames []string) *CounterVec
 //     myVec.WithLabelValues("404", "GET").Add(42)
 func (c *CounterVecProxy) WithLabelValues(lvs ...string) prometheus.Counter {
 	if len(lvs) > 0 {
-		labels := make(map[string]string, len(lvs))
-		for index, label := range lvs {
-			labels[c.LabelNames[index]] = label
-		}
 		c.mu.Lock()
-		noteLabelsInMetricsProxy(c, labels)
+		noteLabelsInMetricsProxy(c, lvs)
```
csuzhangxc, line 52: How about only lock `c.Labels` (not lock two-level function call and `strings.Join`)? or maybe you can try `sync.Map`.

syncer/syncer.go
```diff
@@ -1669,6 +1670,9 @@ func (s *Syncer) handleRowsEvent(ev *replication.RowsEvent, ec eventContext) err
 }
 
 func (s *Syncer) handleQueryEvent(ev *replication.QueryEvent, ec eventContext) error {
+	if bytes.Equal(ev.Query, []byte("BEGIN")) {
```
lance6716, line 1673: not sure if all binlog is upper-case. (it's OK till now)

lance6716 (commented):
will review later

lance6716:
LGTM

pkg/conn/baseconn.go
```diff
@@ -160,9 +160,11 @@ func (conn *BaseConn) ExecuteSQLWithIgnoreError(tctx *tcontext.Context, hVec *me
 			arg = args[i]
 		}
 
-		tctx.L().Debug("execute statement",
-			zap.String("query", utils.TruncateString(query, -1)),
-			zap.String("argument", utils.TruncateInterface(arg, -1)))
+		if tctx.L().Check(zap.DebugLevel, "test") != nil {
```
lance6716, line 163: how does this change take effect?

GMHDBJD: If not debug level, this func will return nil, and will not call TruncateInterface

csuzhangxc: @GMHDBJD you may add a comment here to explain it (for the future).

syncer/checkpoint.go
```diff
@@ -139,14 +139,14 @@ func (b *binlogPoint) outOfDate() bool {
 func (b *binlogPoint) MySQLLocation() binlog.Location {
 	b.RLock()
 	defer b.RUnlock()
-	return b.location.Clone()
+	return b.location
```
csuzhangxc, line 142: How about adding some comments to tell the caller to clone it for writer operations? if so, ditto for `FlushedMySQLLocation`. may also for other public methods?

pkg/binlog/position.go
```diff
@@ -346,3 +338,19 @@ func compareIndex(lhs, rhs int) int {
 func (l *Location) ResetSuffix() {
 	l.Suffix = 0
 }
+
+// SetGTID set new gtid for location
+func (l *Location) SetGTID(gset mysql.GTIDSet) error {
```
csuzhangxc, line 343: How about adding a UT for read/write with a shallow copy? (to simulate passing a location to `job` or `handle-error`)

syncer/checkpoint.go
GMHDBJD: Use SetGTID rather than GITD.Set always. We may change attributes in Location to lowercase later and don't need to clone.

pkg/binlog/position.go
GMHDBJD: add ut in f1b2f52. Pointers need to be treated with care.

pkg/binlog/position_test.go
```diff
@@ -723,3 +723,45 @@ func (t *testPositionSuite) TestVerifyBinlogPos(c *C) {
...
+	loc2.Position.Pos++
+	c.Assert(loc.Position.Pos, Equals, uint32(2333))
+	c.Assert(CompareLocation(loc, loc2, false), Equals, -1)
+
+	loc2.Position.Name = "mysql-bin.00001"
+	c.Assert(loc.Position.Name, Equals, "mysql-bin.00002")
+	c.Assert(CompareLocation(loc, loc2, false), Equals, 1)
+
+	// WARN: will change other location's gtid
+	err := loc2.GTIDSet.Set(mysqlSet2)
```
lance6716, line 756: agree to make member `GTIDSet` private, to avoid this usage (if need much change, later is fine)

lance6716 (commented):
LGTM

csuzhangxc (approved):
LGTM

PR #1189, opened 2020-10-20
loader: fix nil map after clone

lance6716:
fix nil map after clone. We may found more map or develop a safe clone in future

### Check List 

Tests 

Code changes

Side effects

Related changes

### DIFF
--- a/loader/loader.go
+++ b/loader/loader.go
@@ -465,9 +465,12 @@ func (l *Loader) Init(ctx context.Context) (err error) {
 	if err != nil {
 		return err
 	}
-	// fix nil map after clone
-	// TODO: we may develop `TrueClone` in future
-	lcfg.To.Session = make(map[string]string)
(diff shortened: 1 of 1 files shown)


csuzhangxc (approved):
LGTM

loader/loader.go
```diff
@@ -465,6 +465,9 @@ func (l *Loader) Init(ctx context.Context) (err error) {
 	if err != nil {
 		return err
 	}
+	// fix nil map after clone
+	// TODO: we may develop `TrueClone` in future
+	lcfg.To.Session = make(map[string]string)
```
csuzhangxc, line 470: if `lcfg.To.Session` has some valid values, this `make` will clear them, right?

csuzhangxc (commented):
LGTM

GMHDBJD (approved):
LGTM

PR #1195, opened 2020-10-21
fix: fix wrong shorthand for `statement-size`

csuzhangxc:

in dumpling, we use `-s` as the shorthand for `--statement-size` (https://github.com/pingcap/dumpling/blob/270170441a02b0a31ea927ae3ffb13c4acb8dc23/cmd/dumpling/main.go#L102), but we use `-S` in DM.

change `S` to `s`.

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes


GMHDBJD (approved):
LGTM

PR #1247, opened 2020-10-30
test: fix unstable handle-error test case

GMHDBJD:
[jenkins](https://internal.pingcap.net/idc-jenkins/blue/organizations/jenkins/test_dm_master/detail/test_dm_master/1634/pipeline/)
after execute `handle-error revert`, task will auto resume.
when execute `handle-error skip`, task is still running so the case fail
add `query-status` between revert and skip, make sure task is pause by error

### Check List 

Tests 

 - Integration test


csuzhangxc (commented):
LGTM

lance6716 (commented):
LGTM

lance6716 (approved):

PR #1297, opened 2020-11-25
handle-error: only support to handle query error

GMHDBJD:
only support handle query error to avoid misuse
add a check when handle-error

### Check List 

Tests 
 - Integration test

### DIFF
--- a/syncer/handle_error.go
+++ b/syncer/handle_error.go
@@ -33,9 +33,12 @@ func (s *Syncer) HandleError(ctx context.Context, req *pb.HandleWorkerErrorReque
 	pos := req.BinlogPos
 
 	if len(pos) == 0 {
-		startLocation := s.getErrLocation()
+		startLocation, isQueryEvent := s.getErrLocation()
 		if startLocation == nil {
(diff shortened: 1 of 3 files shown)


syncer/handle_error.go
```diff
@@ -33,10 +33,13 @@ func (s *Syncer) HandleError(ctx context.Context, req *pb.HandleWorkerErrorReque
 	pos := req.BinlogPos
 
 	if len(pos) == 0 {
-		startLocation := s.getErrLocation()
+		startLocation, isQueryEvent := s.getErrLocation()
 		if startLocation == nil {
 			return fmt.Errorf("source '%s' has no error", s.cfg.SourceID)
 		}
+		if !isQueryEvent {
+			return fmt.Errorf("only support to handle query error currently")
```
lance6716, line 41: How about change to "handle DDL error" (query is unfamiliar to me), and provide some guidance to handle non-DDL error

lance6716 (commented):
rest LGTM

lance6716:
LGTM

csuzhangxc (approved):
LGTM

PR #1306, opened 2020-11-26
sycner: automatically use upstream sql_mode

GMHDBJD:
Inconsistent upstream and downstream sql_mode leads to inconsistent data

get dsn scope sql_mode from upstream database if sql_mode is not specified in task config.

**Still can't use different session sql_mode base on different binlog event like mysql master-slave replication. One solution is get sql_mode from `begin` event and apply it for rows event in same txn, which is ineffective.**
### Check List 

Tests 

 - Integration test

### DIFF
--- a/syncer/syncer.go
+++ b/syncer/syncer.go
@@ -2351,23 +2351,25 @@ func (s *Syncer) createDBs(ctx context.Context) error {
 		return err
 	}
 
+	hasSQLMode := false
 	// get sql_mode from upstream db
 	if s.cfg.To.Session == nil {
 		s.cfg.To.Session = make(map[string]string)
-	}
-	hasSQLMode := false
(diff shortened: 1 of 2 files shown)


syncer/syncer.go
```diff
@@ -2351,6 +2351,25 @@ func (s *Syncer) createDBs() error {
 		return err
 	}
 
+	// get sql_mode from upstream db
+	if s.cfg.To.Session == nil {
+		s.cfg.To.Session = make(map[string]string)
+	}
+	hasSQLMode := false
+	for k := range s.cfg.To.Session {
```
lance6716, line 2359: could add an `else` in L2357 and put this into `else`

lance6716 (commented):
rest LGTM

csuzhangxc (approved):
LGTM

Issue #1320, opened 2020-12-02
precheck didn't work so dumpling created illegal metadata


lance6716:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do? If possible, provide a recipe for reproducing the error.

https://asktug.com/t/topic/64011/7

2. What did you expect to see?

since this error is `Error 1227: Access denied; you need (at least one of) the SUPER, REPLICATION CLIENT privilege(s) for this operation`, missing of REPLICATION CLIENT should be deteced early

3. What did you see instead?

parse mydumper metadata error: EOF, metadata: Started dump at: 2020-12-01 18:16:30\nSHOW MASTER STATUS:\nSHOW MASTER STATUS: /* AFTER CONNECTION POOL ESTABLISHED */\nFinished dump at: 2020-12-01 23:07:35\n

4. Versions of the cluster

    - DM version (run `dmctl -V` or `dm-worker -V` or `dm-master -V`):

        2.0.0

    - Upstream MySQL/MariaDB server version:

        MySQL 8.0.14

    - Downstream TiDB cluster version (execute `SELECT tidb_version();` in a MySQL client):

        ```
        (paste TiDB cluster version here)
        ```

    - How did you deploy DM: DM-Ansible or manually?

        ```
        (leave DM-Ansible or manually here)
        ```

    - Other interesting information (system version, hardware config, etc):

        >
        >

5. current status of DM cluster (execute `query-status` in dmctl)

6. Operation logs
   - Please upload `dm-worker.log` for every DM-worker instance if possible
   - Please upload `dm-master.log` if possible
   - Other interesting logs
   - Output of dmctl's commands with problems
   
7. Configuration of the cluster and the task
   - `dm-worker.toml` for every DM-worker instance if possible
   - `dm-master.toml` for DM-master if possible
   - task config, like `task.yaml` if possible
   - `inventory.ini` if deployed by DM-Ansible

8. Screenshot/exported-PDF of Grafana dashboard or metrics' graph in Prometheus for DM if possible

glkappe:
@lance6716 This issue will be cherry picked in which version of dm

lance6716:
> @lance6716 This issue will be cherry picked in which version of dm

the root cause is still not located, maybe it's a MySQL 8.0 problem. If so, DM v1.0 didn't support MySQL 8.0 so that's out of our support

lance6716:
> @lance6716 This issue will be cherry picked in which version of dm

[...]

PR #1321, opened 2020-12-02
load: load should not need metadata

lance6716:
part of #1320 

since full mode doesn't not need REPLICATION related privilege and doesn't check it, mydumper/dumpling won't generate a valid `metdata` file containing "SHOW MASTER STATUS:" at then. And when loader parse it, load task will stop.

Now load should ignore that error

### Check List 

Tests 

 - pass original test

Code changes

Side effects

Related changes


lance6716:
Why didn't relay unit start sync from metadata's position, thus we need not `unitTransWaitCondition` 🤔

GMHDBJD:
> Why didn't relay unit start sync from metadata's position, thus we need not `unitTransWaitCondition` 🤔

I think relay should be source level but not task level. So it may start from now, or start from pos/gtid which we specify

loader/loader.go
```diff
@@ -1307,6 +1307,11 @@ func (l *Loader) getMydumpMetadata() error {
 	metafile := filepath.Join(l.cfg.LoaderConfig.Dir, "metadata")
 	loc, _, err := dumpling.ParseMetaData(metafile, l.cfg.Flavor)
 	if err != nil {
+		if terror.ErrMetadataNoBinlogLoc.Equal(err) {
+			l.logger.Warn("dumped metadata doesn't have binlog location, it's OK if DM doesn't enter incremental mode")
```
GMHDBJD, line 1311: How about check task_mode here? 
It is annoying to report errors after a long time import is completed

lance6716: We're checking REPLICATION CLIENT in pre-check wrongly, (should make sure it's grants on `*.*`). So for all mode and incremental mode, pre-check could make sure there's enough privilege generating correct metadata. For full mode, it's no harm

GMHDBJD: Are there any other cases that may cause this error? 🤔

GMHDBJD (commented):
LGTM

csuzhangxc (approved):
LGTM

PR #1323, opened 2020-12-03
feat: add get-config command

GMHDBJD:
use get-config command to get task/worker/master/source cfg.

- [x] refine get-task-config to support merge from subtask
- [x] get worker/master/source cfg

### Check List 

Tests 

 - Unit test
 - Integration test


dm/ctl/ctl.go
```diff
@@ -72,7 +72,7 @@ func NewRootCmd() *cobra.Command {
 		master.NewOperateLeaderCmd(),
 		master.NewListMemberCmd(),
 		master.NewOperateSchemaCmd(),
-		master.NewGetTaskCfgCmd(),
```
lance6716, line 75: I'm afraid some users have remembered this command, could we keep it as an alias of `get-config` or give redirect message?

dm/master/scheduler/scheduler.go
```diff
@@ -336,126 +323,6 @@ func (s *Scheduler) RemoveSourceCfg(source string) error {
...
-		return nil
-	}
-
-	// 1. check whether exists.
-	// return nil because we may use `start-task -s source` for subtask
-	if _, ok := s.taskCfgs[cfg.Name]; ok {
-		return nil
-	}
-
-	// 2. put the config into etcd.
```
lance6716, line 369: we don't need this function now?

lance6716 (commented):
will review later 😭

chaos/cases/task.go
```diff
@@ -363,6 +365,10 @@ func (t *task) genIncrData(ctx context.Context) (err error) {
 				})
 			}
 			if err = eg.Wait(); err != nil {
+				if utils.IsMySQLError(err, mysql.ErrDupFieldName) {
```
csuzhangxc, line 368: `eg.Wait()` will only return the **first** error, but there may be other errors at the same time. How about doing this filter in `eg.Go(func() error {`?

dm/ctl/ctl.go
GMHDBJD: alias it for both interactionMode and commandMode

dm/master/scheduler/scheduler.go
GMHDBJD: No need to save them into etcd now.

dm/config/task.go
```diff
@@ -628,6 +615,21 @@ func (c *TaskConfig) SubTaskConfigs(sources map[string]DBConfig) ([]*SubTaskConf
 	return cfgs, nil
 }
 
+func getGenerateName(rule interface{}, nameIdx int, namePrefix string, nameMap map[string]string) (string, int) {
```
lance6716, line 618: please add a comment about dealing arguments and returns of this function

dm/config/task.go
```diff
@@ -656,39 +658,48 @@ func (c *TaskConfig) FromSubTaskConfigs(stCfgs ...*SubTaskConfig) {
 	c.Loaders = make(map[string]*LoaderConfig)
 	c.Syncers = make(map[string]*SyncerConfig)
 
+	BAListMap := make(map[string]string, len(stCfgs))
```
lance6716, line 661: `baList`

dm/config/task_test.go
```diff
@@ -429,7 +440,7 @@ func (t *testConfig) TestGenAndFromSubTaskConfigs(c *C) {
 				Security:         &security,
 				RawDBCfg:         &rawDBCfg,
 			},
-			RouteRules:  []*router.TableRule{&routeRule1, &routeRule2},
+			RouteRules:  []*router.TableRule{&routeRule2, &routeRule1},
```
lance6716, line 443: we could test different subtask has different route rules, and merged task won't lost any of them for new implementation of `FromSubTaskConfigs `

lance6716 (commented):
will review later

tests/dmctl_basic/check_list/get_config.sh
```diff
@@ -0,0 +1,125 @@
...
+        "operate-source create $WORK_DIR/get_source2.yaml" \
+        "\"result\": true" 2
+
+    # start task with get config
+    run_dm_ctl $WORK_DIR "127.0.0.1:$MASTER_PORT" \
+        "start-task $WORK_DIR/get_task.yaml" \
+        "\"result\": true" 3
+
+    # restart master with get config
+    ps aux | grep dm-master |awk '{print $2}'|xargs kill || true
```
lance6716, line 102: we might try `killall` (not sure it's supported in CI, it's OK not changing)

tests/dmctl_basic/conf/dm-master.toml
```diff
@@ -1,4 +1,5 @@
 # Master Configuration.
 
+name = "master1"
```
lance6716, line 3: if user didn't set this field, could `get-config` find this config? if so, should we set a default value

dm/master/server.go
```diff
@@ -1781,27 +1768,140 @@ func (s *Server) OperateSchema(ctx context.Context, req *pb.OperateSchemaRequest
...
+			return subCfgList[i].SourceID < subCfgList[j].SourceID
+		})
 
-	if len(cfg) == 0 {
-		return &pb.GetTaskCfgResponse{
-			Result: false,
-			Msg:    "task not found",
-		}, nil
+		var taskCfg config.TaskConfig
+		taskCfg.FromSubTaskConfigs(subCfgList...)
```
lance6716, line 1845: How about move reciever of this function to return value
And update the comment of this function (not only used when import from v1.0)

lance6716 (commented):
rest LGTM

cmd/dm-ctl/main.go
```diff
@@ -240,3 +243,20 @@ func loop() {
...
 }
+
+func aliasArgs(args []string) []string {
+	args = aliasGetTaskCfgCmd(args)
+	return args
+}
+
+func aliasGetTaskCfgCmd(args []string) []string {
+	for i, arg := range args {
+		if arg == "get-task-config" {
```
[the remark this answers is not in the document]
GMHDBJD: Not for command mode. `./dmctl --master-addr get-task-config`

csuzhangxc (commented):
LGTM

dm/config/task.go
GMHDBJD: We use BAList in task config. https://github.com/pingcap/dm/blob/4ebd3b5de1e4e7ba49faaae508d671e0b4df03df/dm/config/task.go#L318

tests/dmctl_basic/conf/dm-master.toml
GMHDBJD: We adjust master name for user. 
https://github.com/pingcap/dm/blob/master/dm/master/config.go#L250-L257
They may use list-member to check it.

dm/config/task.go
lance6716: IMO that's public accessible so could use uppercase

lance6716 (approved):
LGTM

PR #1316, opened 2020-11-30
relay: support relay log in v2.0 (#1272)

ti-srebot:
cherry-pick #1272 to release-2.0

---

### What problem does this PR solve? 
part of #1234 

### What is changed and how it works?
- [x]  support TLS in relay
- [x]  support gtid in relay

### Check List 

Tests 

 - Unit test
 - Integration test
 - Manual test (add detailed scripts or steps below)

PR #1326, opened 2020-12-10
load: load should not need metadata (#1321)

ti-srebot:
cherry-pick #1321 to release-2.0

---

part of #1320 

since full mode doesn't not need REPLICATION related privilege and doesn't check it, mydumper/dumpling won't generate a valid `metdata` file containing "SHOW MASTER STATUS:" at then. And when loader parse it, load task will stop.

Now load should ignore that error

### Check List 

Tests 

 - pass original test

Code changes

Side effects

Related changes


lance6716:
waiting https://github.com/pingcap/dm/pull/1316 to resolve conflicts

csuzhangxc (commented):
LGTM

csuzhangxc (approved):

PR #1340, opened 2020-12-15
parser: split ALTER TABLE ADD COLUMN

### LINKED ISSUE #1339: should split DDL for `ADD COLUMN (c1 int, c2 int)`
## Bug Report

in upstream. 

```
ALTER TABLE `all_mode`.`t1` ADD COLUMN (`t6` INT, `t7` INT)
```

sync to downstream, split same as

```
alter table t1 add column t6 int, add column t7 int;
```

3. What did you see instead?

```

                                "Message": "startLocation: [position: (mysql-bin.000001, 1902), gtid-set: 09bec856-ba95-11ea-850a-58f2b4af5188:1-8], endLocation: [position: (mysql-bin.000001, 2026), gtid-set: ]: execute statement failed: ALTER TABLE `all_mode`.`t1` ADD COLUMN (`t6` INT, `t7` INT)",
                                "RawCause": "Error 8200: Unsupported multi schema change",
```

4. Versions of the cluster

        master

lance6716:
close #1339 

split it

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

Related changes


pkg/parser/common.go
```diff
@@ -271,6 +271,26 @@ func SplitDDL(stmt ast.StmtNode, schema string) (sqls []string, err error) {
...
+					bf.Reset()
+					err = stmt.Restore(ctx)
+					if err != nil {
+						v.Specs = specs
+						v.Table = table
+						return nil, terror.ErrRestoreASTNode.Delegate(err)
+					}
+					sqls = append(sqls, bf.String())
+				}
+				continue
```
csuzhangxc, line 291: why need this `continue` here?

csuzhangxc (commented):
rest LGTM

pkg/parser/common.go
lance6716: it's the continue of `for _, spec := range specs`. because we now know this `spec` is a add column statement and restored its SQL for every column, we could skip below logic about "restore whole spec to SQL" and "handle rename table"

csuzhangxc: you can add a comment to highlight this.

csuzhangxc (commented):
LGTM

GMHDBJD (approved):
LGTM

lance6716:
now the slowest test case is `handle_error`, about 12min

PR #1350, opened 2020-12-21
chaos: fix diffIncrData has no effect

GMHDBJD:
- chaos test will always return true when diff incremental data 😨
- alter table add column(timestamp) will have different default value between mysql:5.7 and tidb.

- change context
- add session variable `explicit_defaults_for_timestamp`

### Check List 

Tests 
 - chaos test

### DIFF
--- a/chaos/cases/task.go
+++ b/chaos/cases/task.go
@@ -37,7 +37,7 @@ import (
 const (
 	tableCount      = 10               // tables count in schema.
 	fullInsertCount = 100              // `INSERT INTO` count (not rows count) for each table in full stage.
-	diffCount       = 10               // diff data check count
+	diffCount       = 20               // diff data check count
 	diffInterval    = 10 * time.Second // diff data check interval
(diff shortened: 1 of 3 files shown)


csuzhangxc (commented):
LGTM. 👍

lance6716 (approved):
LGTM

> chaos test will always return true when diff incremental data 😨

caused by after 20 seconds previous context is done?

GMHDBJD:
> caused by after 20 seconds previous context is done?

genIncData exit until context done. So diffIncrData's ctx is always done.

chaos/cases/task.go
```diff
@@ -37,7 +37,7 @@ import (
 const (
 	tableCount      = 10               // tables count in schema.
 	fullInsertCount = 100              // `INSERT INTO` count (not rows count) for each table in full stage.
-	diffCount       = 10               // diff data check count
+	diffCount       = 20               // diff data check count
```
GMHDBJD, line 40: image

csuzhangxc (approved):

PR #1369, opened 2021-01-11
test: refine some tests

### LINKED ISSUE #129: Use gopl to detect code clones
[gopl](https://github.com/mibk/dupl) can be used to detect duplicate codes.

We can use `gopl -html > dm.html` to create a report with contents similar to the following:

Screen Shot 2019-04-30 at 1 06 56 PM

With the report, we can refactor to eliminate duplicate code.

lance6716:
close #129 

the rest of it doesn't improve a lot IMO, choose to ignore them 👀 

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes


3pointer (approved):
LGTM

PR #1366, opened 2021-01-06
dep: check some privilege should be globally granted

lance6716:
### What problem does this PR solve? 
close #1320 

### What is changed and how it works?
tidb-tools should have fix it

### Check List 

Tests 

 - No code

Code changes

Side effects

Related changes

 - Need to cherry-pick to the release branch
 - Need to be included in the release note

PR #1372, opened 2021-01-12
test: fix test isn't suitable for an unordered output

lance6716:
follow up #1366

change to unorder match

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes


lichunzhu (approved):
LGTM

Issue #1044, opened 2020-09-16
pin source bound even the DM-worker instance offline


csuzhangxc:
## Feature Request

**Is your feature request related to a problem? Please describe:**

the load stage has not supported HA now, if a DM-worker instance is offline temporarily, then the source will be scheduled to another DM-worker instance while the new one will lack the full dumped data.

**Describe the feature you'd like:**

pin the source bound relationship even the DM-worker instance is offline.

then the user can fix & make the DM-worker instance online again and the DM-worker instance can continue the previous processes.

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

PR #1373, opened 2021-01-12
*: last bound source is prior to other source when pick for bound

lance6716:
similar to #1044 

when DM cluster restarts, worker is not likely keepalive when master goes to below logic, because etcd is in masters

https://github.com/pingcap/dm/blob/49223f3aa9c0ee421b398cc5bdd4f5e099415659/dm/master/scheduler/scheduler.go#L924-L936

so worker is likely bounded to another source rather than its last source.

we always save a `lastBound` when do bounding, and not delete it. so when `tryBoundForWorker`, we could prefer selecting the last bound source for worker.

Also, if user didn't deploy more workers than sources, the bound relationship will not change (I hope)

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

 - Has persistent data change

Side effects

Related changes


lance6716:
cc @lichunzhu

lance6716:
PTAL @GMHDBJD @lichunzhu

GMHDBJD:
don't forget to update [keyslen](https://github.com/pingcap/dm/blob/de9de762ef8ef2a6fe88ced759a344ffd4c0a151/dm/common/common.go#L79)

GMHDBJD:
Should [recover](https://github.com/pingcap/dm/blob/de9de762ef8ef2a6fe88ced759a344ffd4c0a151/dm/master/scheduler/scheduler.go#L176) when leader change

dm/common/common.go
```diff
@@ -38,6 +38,10 @@ var (
 	// UpstreamBoundWorkerKeyAdapter used to store address of worker in which MySQL-tasks which are running.
 	// k/v: Encode(name) -> the bound relationship.
 	UpstreamBoundWorkerKeyAdapter KeyAdapter = keyHexEncoderDecoder("/dm-master/bound-worker/")
+	// UpstreamLastBoundWorkerKeyAdapter used to store address of worker in which MySQL-tasks which are running.
+	// different with UpstreamBoundWorkerKeyAdapter, this kv should not be deleted when unbound, to provide a priority
```
GMHDBJD, line 42: What can we do if we actually want to transfer two bounded sources.

GMHDBJD:
Should we delete last bound info when [stop source](https://github.com/pingcap/dm/blob/1fc1f5fcd7b25d0bfd05589a053ccfe1c4f59422/dm/master/scheduler/scheduler.go#L301)?

dm/master/scheduler/scheduler.go
```diff
@@ -1164,14 +1173,22 @@ func (s *Scheduler) handleWorkerOffline(ev ha.WorkerEvent, toLock bool) error {
...
-	// 1. check whether any unbound source exists.
-	var source string
-	for source = range s.unbounds {
-		break // got a source.
+	// 1. check if last bound is still available.
+	// if lastBound not found, or this source has been bounded to another worker (we also check that source still exists
+	// here), randomly pick one from unbounds.
+	// NOTE: if worker isn't in lastBound, we'll get "zero" SourceBound and it's OK, because "zero" string is not in
+	// unbounds
+	source := s.lastBound[w.baseInfo.Name].Source
```
GMHDBJD, line 1184: Maybe panic?

dm/master/scheduler/scheduler.go
```diff
@@ -111,6 +111,9 @@ type Scheduler struct {
 	// - when bounded the source to a worker.
 	unbounds map[string]struct{}
 
+	// a mirror of bounds which not delete. worker -> SourceBound
```
GMHDBJD, line 114: should update comment

dm/master/scheduler/scheduler.go
```diff
@@ -1198,14 +1215,31 @@ func (s *Scheduler) tryBoundForWorker(w *Worker) (bounded bool, err error) {
 // tryBoundForSource tries to bound a source to a random Free worker.
 // returns (true, nil) after bounded.
 func (s *Scheduler) tryBoundForSource(source string) (bool, error) {
-	// 1. try to find a random Free worker.
+	// 1. try to find history workers, then random Free worker.
```
GMHDBJD, line 1218: Seems no need now. Because we now remove lastbound when stop-source. And if a worker offline and call `tryBoundForSource`, this check make no sense

dm/master/scheduler/scheduler.go
lance6716: will return a zero value of string, because `ha.SourceBound` of map value is not a pointer. And in test we `tryBoundForWorker(worker3)` which is not in `lastBound`

dm/master/scheduler/scheduler.go
lance6716: if a worker offline and call `tryBoundForSource`, history worker is still prefered when

(s1, w1), (null, w2), (null, w3)
worker1 offline
(s1, w2), (null, w3)
worker1 online
(null, w1), (s1, w2), (null, w3)

now user could kill w2 to wish w1 bounded to s1. and the priority could ensure that scheduling.

GMHDBJD: > (s1, w2), (null, w3)

Will lastbound be overwritten here?

lance6716: the key of lastbound is worker, so not overwritten

GMHDBJD (commented):
LGTM

lance6716:
ping @lichunzhu

lance6716:
PTAL @lichunzhu

lichunzhu (approved):
LGTM

PR #1376, opened 2021-01-13
chaos: half time per round

GMHDBJD:
chaos unstable

decrease time for incremental round

### Check List 

Tests 

 - chaos


lance6716 (approved):
LGTM

PR #1378, opened 2021-01-14
loader: fix loader lost data when quickly starts/stops task

### LINKED ISSUE #1377: load phase of `all` mode may lost data when worker frequently starts/stops task
## Bug Report

    will reproduce error later.

    This bug is caused by two hidden bugs 

    a) `cleanDumpFiles` is wrongly triggered because it will check `checkPoint.AllFinished()`, but this is a memory cache of DB which may not initialized
    b) load unit didn't return error when there're missing SQL files which should exists according to checkpoint

    when worker is too quickly `Close` because of scheduling of network problems, worker may trigger a). In all mode, `cleanDumpFiles` will clean the SQL files (in full mode `cleanDumpFiles` will clean whole directory including dump metadata thus would cause an error). Then next time when it resume with the task, worker is continued on load unit because `IsFresh` will look into database rather than cache like `checkPoint.AllFinished()`. When load unit starts, it will trigger b) so finish load unit because of no files to load and goes to sync unit.

3. What did you see instead?

4. Versions of the cluster

        ```
        (paste DM version here, and you must ensure versions of dmctl, DM-worker and DM-master are same)
        ```

GMHDBJD:
close #1377 

cleanDumpfiles only finish task

### Check List 

Tests 

 - Integration test


lance6716 (commented):
LGTM

3pointer (approved):
LGTM

Issue #1379, opened 2021-01-15
relay: invalid GTID format, must UUID:interval


GMHDBJD:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do? If possible, provide a recipe for reproducing the error.
worker bound source from mysql to mariadb
2. What did you expect to see?
task runs well
3. What did you see instead?
```
some error occurs in dm-worker: ErrCode:30003 ErrClass:\"relay-unit\" ErrScope:\"internal\" ErrLevel:\"high\" Message:\"initial relay unit: load meta data\" RawCause:\"invalid GTID format, must UUID:interval[:interval]\"
```
[chaos](https://github.com/pingcap/dm/pull/1370/checks?check_run_id=1701081954)

lance6716:
should be closed by #1385

PR #1385, opened 2021-01-18
config: fix empty flavor for source

### LINKED ISSUE #1380: dmctl fail to operate-source when flavor is not specified and GTID is set
## Bug Report

cat source.yaml

```
source-id: "test"

enable-gtid: true
enable-relay: true
relay-binlog-gtid: "8fed3d32-1e6e-11eb-9e0f-0242ac110003:1-2933"
from:
  host: "127.0.0.1"
  user: "root"
  password: ""
  port: 3306
purge:
  remain-space: 15
```

./dmctl operator-source create source.yaml

Add source successfully.

3. What did you see instead?
```
[code=11009:class=dm-worker:scope=internal:level=high], Message: relay-binlog-gtid 8fed3d32-1e6e-11eb-9e0f-0242ac110003:1-2933: parse GTID 8fed3d32-1e6e-11eb-9e0f-0242ac110003:1-2933, RawCause: invalid flavor 
```

4. Versions of the cluster

       `1fc1f5fcd7b25d0bfd05589a053ccfe1c4f59422`

GMHDBJD:
close #1380 #1379 

- adjust source config before verify
- try mysql/mariadb if flavor is empty

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code


dm/worker/source.yaml
```diff
@@ -7,13 +7,14 @@ server-id: 101
...
-flavor: mysql
 
 #directory that used to store relay log
 relay-dir: ./relay_log
 
 #enable gtid in relay log unit
-enable-gtid: false
+enable-gtid: true
+
+relay-binlog-gtid: "e68f6068-53ec-11eb-9c5f-0242ac110003:1-50"
```
lance6716, line 17: there's a case that user didn't specify flavor or gtid, expecting starting relay at position of `SHOW MASTER STATUS`. Will `SHOW MASTER STATUS` returned empty `Executed_Gtid_Set`? If so, error message of "empty flavor with empty gtid is invalid" is misleading because he could only left them empty to use position of `SHOW MASTER STATUS`

GMHDBJD: The `flavor` should not be empty since we have adjusted it or specify it, otherwise it is a bug. So `empty flavor with empty gtid is invalid` should not happen.

pkg/gtid/gtid.go
```diff
@@ -47,16 +48,35 @@ type Set interface {
...
+		err  error
+		gtid mysql.GTIDSet
 	)
 
-	gtid, err := mysql.ParseGTIDSet(flavor, gtidStr)
+	if len(flavor) == 0 && len(gtidStr) == 0 {
+		return nil, errors.Errorf("empty flavor with empty gtid is invalid")
+	}
+
+	fla := flavor
```
3pointer, line 60:
```
if len(flavor) == 0 {
    flavor = mysql.MySQLFlavor
}
gtid, err = mysql.ParseGTIDSet(flavor, gtidStr)
if err != nil {
    flavor = mysql. MariaDBFlavor
    gtid, err = mysql.ParseGTIDSet(flavor, gtidStr)
    if err != nil {
        return nil, terror.ErrNotSupportedFlavor.Generate(flavor)
    }
}

GMHDBJD: If user set `flavor: MySQL` but use MariaDB as upstream,  we may need to report an error?

lance6716: > If user set flavor: MySQL but use MariaDB as upstream, we may need to report an error?

I agress that we should report an error, so we should differ empty flavor and user specified `mysql` flavor when  `mysql.ParseGTIDSet("mysql", gtidStr)` failed. PTAL @3pointer

lance6716:
PTAL @lichunzhu @3pointer

3pointer (approved):
LGTM

PR #1390, opened 2021-01-19
relay: align with MySQL when using GTID

### LINKED ISSUE #1383: relay's behaviour should align with MySQL when using GTID
## Bug Report

https://dev.mysql.com/doc/refman/5.7/en/replication-gtids-auto-positioning.html

> The source responds by sending all transactions recorded in its binary log whose GTID is not included in the GTID set sent by the replica. To do this, the source first identifies the appropriate binary log file to begin working with, by checking the Previous_gtids_log_event in the header of each of its binary log files, starting with the most recent. When the source finds the first Previous_gtids_log_event which contains no transactions that the replica is missing, it begins with that binary log file. This method is efficient and only takes a significant amount of time if the replica is behind the source by a large number of binary log files. The source then reads the transactions in that binary log file and subsequent files up to the current one, sending the transactions with GTIDs that the replica is missing, and skipping the transactions that were in the GTID set sent by the replica. The elapsed time until the replica receives the first missing transaction depends on its offset in the binary log file. This exchange ensures that the source only sends the transactions with a GTID that the replica has not already received or committed. If the replica receives transactions from more than one source, as in the case of a diamond topology, the auto-skip function ensures that the transactions are not applied twice.

1. `getPosByGTID` may choose start position by 
    
    choose a relay log file that its "Previous_gtids_log_event" don't overlap with "complementary set of given GTID set", which could reduce to "Previous_gtids_log_event" is subset or equal to "given GTID set"
    
2. scan the relay log and skip the transactions in "given GTID set", to handle cases when given GTID set is not continuous like `uuid:1-10:20-30`

GMHDBJD:
close #1383 

- [x] getPosByGTID always return the begin pos of the collect relay log file (filename, 4)
- [x] startSyncByGTID should replace event to heartbeat event which already be contained for specified gtid set
- [x] adjust gtid for relay
 
### Check List 

Tests 

 - Unit test


pkg/binlog/event/event.go
```diff
@@ -851,3 +851,19 @@ func GenDummyEvent(header *replication.EventHeader, latestPos uint32, eventSize
...
 	return ev, err
 }
+
+// GenHeartbeatEvent generates a heartbeat event.
+// ref: https://dev.mysql.com/doc/internals/en/heartbeat-event.html
+func GenHeartbeatEvent(header *replication.EventHeader) *replication.BinlogEvent {
+	// modify header
+	headerClone := *header // do a copy
+	headerClone.Flags = 0
+	headerClone.EventSize = 39
```
lance6716, line 861: does MySQL choose not keep size? I think this may cause positions in `query-status` mismatched which may confuse user

lance6716 (commented):
will review later

pkg/streamer/reader.go
```diff
@@ -118,59 +121,21 @@ func (r *BinlogReader) checkRelayPos(pos mysql.Position) error {
...
-			return relayMeta.UUID, nil
-		}
-	}
-	return "", terror.ErrNoUUIDDirMatchGTID.Generate(gset.String())
-}
-
-// GetFilePosByGTID tries to get Pos by GTID for file
-func (r *BinlogReader) GetFilePosByGTID(ctx context.Context, filePath string, gset mysql.GTIDSet) (uint32, error) {
+// CheckFileByGTID check whether gset contains file's previous_gset
+func (r *BinlogReader) CheckFileByGTID(ctx context.Context, filePath string, gset mysql.GTIDSet) (bool, error) {
```
lance6716, line 125: maybe rename it to `CheckStartSyncFromFile` or `IsGTIDCoverPreviousFiles` or others 😂 current function name shows no usage or behaviour

pkg/streamer/reader.go
```diff
@@ -535,9 +474,9 @@ func (r *BinlogReader) parseFile(
 				break
 			}
 			u, _ := uuid.FromBytes(ev.SID)
-			err2 := r.advanceCurrentGtidSet(fmt.Sprintf("%s:%d", u.String(), ev.GNO))
-			if err2 != nil {
-				return errors.Trace(err2)
+			r.replaceWithHeartbeat, err = r.advanceCurrentGtidSet(fmt.Sprintf("%s:%d", u.String(), ev.GNO))
```
lance6716, line 477: could we use a local variable instead of a member of `BinlogReader`? BTW, I think we could reset `replaceWithHeartbeat` before `switch` in L446

pkg/streamer/reader.go
```diff
@@ -562,6 +501,20 @@ func (r *BinlogReader) parseFile(
...
+		// if an event's gtid has been contained by given gset
+		// replace it with HEARTBEAT event
+		// for Mariadb, it will bee replaced with MARIADB_GTID_LIST_EVENT
+		// In DM, we replace both of them with HEARTBEAT event
+		if r.replaceWithHeartbeat {
+			switch e.Event.(type) {
+			case *replication.RowsEvent, *replication.QueryEvent, *replication.GTIDEvent, *replication.XIDEvent, *replication.TableMapEvent:
+				// replace with heartbeat event
+				e = event.GenHeartbeatEvent(e.Header)
+			default:
```
lance6716, line 514: what's the skipped event type and why 🤔

lance6716 (commented):
will review later

pkg/binlog/event/event.go
GMHDBJD: It keeps the LogPos but not EventSize.

GMHDBJD: The meets problem when we calculate startPosition, but we don't calculate it for heartbeat event.

pkg/streamer/reader.go
GMHDBJD: Other events like format_description_event,rotate_event should not be replaced.

pkg/streamer/reader.go
GMHDBJD: address in 91d1adb

pkg/streamer/reader.go
GMHDBJD: address in 91d1adb

pkg/streamer/reader.go
lance6716: 👍 please add a comment for it

lance6716 (commented):
rest LGTM

(I'll commit the changes since author is leaving)

lichunzhu (approved):
LGTM

ti-srebot (approved):

Issue #1388, opened 2021-01-18
Support syncing when disk is very slow


lance6716:
## Feature Request

**Is your feature request related to a problem? Please describe:**

A user deploy DM in an environment that has very slow disk. that has help revealing some BUG of DM such as
- [ ] https://github.com/pingcap/dm/issues/1377
- [ ] worker recieved a bound watch, but failed to read bound information in etcd and didn't retry or kill itself
- [ ] `query-status` shows nothing, while can't add task because of `already exists`

**Describe the feature you'd like:**

- [ ] expose more etcd error and metrics (already in https://github.com/pingcap/dm/issues/1219, https://github.com/pingcap/dm/issues/1218), and warn when disk is bad
- [ ] test DM in slow disk

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

lance6716:
I have used chaosmesh to try imitate a bad disk environment, but not effective to reveal bugs. We might try use failpoint with percent probability to inject into etcd API, and check if it will cause inconsistency in DM.

@zeminzhou

lance6716:
(removed the BUG label because we need further investigating if it's has been fixed)

PR #1396, opened 2021-01-21
worker, ha: For errors that can only be attempted a limited number of times, the number of attempts is limited

zeminzhou:
worker recieved a bound watch, but failed to read bound information in etcd.[part of #1388] (https://github.com/pingcap/dm/issues/1388)

Add reading bound information failed error to RetryableError type, but limit the number of retry.

### Check List 

Tests 

 - Unit test [WIP]
 - Integration test [WIP]

Code changes

Side effects

Related changes


dm/worker/server.go
```diff
@@ -62,6 +63,9 @@ type Server struct {
 	ctx    context.Context
 	cancel context.CancelFunc
 
+	kpaCtx    context.Context
```
lance6716, line 66: I think we could rename it to `kaCtx` or `keepaliveCtx`, also it's OK for other names or not change

dm/worker/server.go
```diff
@@ -142,16 +146,21 @@ func (s *Server) Start() error {
 		defer s.wg.Done()
 		// TODO: handle fatal error from observeSourceBound
```
lance6716, line 147: now this TODO could be removed

dm/worker/server.go
```diff
@@ -142,16 +146,21 @@ func (s *Server) Start() error {
...
 		//nolint:errcheck
-		s.observeSourceBound(s.ctx, revBound)
+		for {
+			s.observeSourceBound(s.ctx, revBound)
```
lance6716, line 150: we might make use of the returned error (for example, if `err == nil` means parent context is done)

dm/worker/server.go
```diff
@@ -142,16 +146,21 @@ func (s *Server) Start() error {
...
+			s.stopKeepAlive()
+			select {
+			case <-s.ctx.Done():
+				return
+			case <-time.After(keepaliveTimeout * 2):
```
lance6716, line 155: we didn't give every variables a proper name 😂 `keepaliveTimeout` is not as you expected. It should be `s.cfg.KeepAliveTTL`

also welcome to give old variables better names

dm/worker/server.go
```diff
@@ -190,6 +199,18 @@ func (s *Server) Start() error {
...
+func (s *Server) startKeepAlive() {
+	defer s.wg.Done()
+	// worker keepalive with master
+	// If worker loses connect from master, it would stop all task and try to connect master again.
+	s.kpaCtx, s.kpaCancel = context.WithCancel(context.Background())
+	s.KeepAlive()
+}
+
+func (s *Server) stopKeepAlive() {
+	s.kpaCancel()
```
lance6716, line 211: `s.kpaCancel()` maybe called uninitialized and panic if line 151 comes too quickly. And also there is data race

dm/worker/server.go
```diff
@@ -190,6 +199,18 @@ func (s *Server) Start() error {
 	return terror.ErrWorkerStartService.Delegate(err)
 }
 
+func (s *Server) startKeepAlive() {
+	defer s.wg.Done()
+	// worker keepalive with master
+	// If worker loses connect from master, it would stop all task and try to connect master again.
+	s.kpaCtx, s.kpaCancel = context.WithCancel(context.Background())
```
lance6716, line 206: we might use `s.ctx` rather than `context.Background()`. because line 163 doesn't have its corresponding `stopKeepAlive`, thus it may never quit.

lance6716 (commented):
will review later

dm/worker/server.go
lance6716: please address above comment

dm/worker/server.go
```diff
@@ -138,19 +142,23 @@ func (s *Server) Start() error {
...
-		// If worker loses connect from master, it would stop all task and try to connect master again.
-		s.KeepAlive()
+		for {
+			//nolint:errcheck
+			s.observeSourceBound(s.ctx, revBound)
+			s.stopKeepAlive()
+			select {
+			case <-s.ctx.Done():
+				return
+			case <-time.After(keepaliveTimeout):
```
lance6716, line 157: sleep time should be related to `s.cfg.KeepAliveTTL`, for example, `2 * s.cfg.KeepAliveTTL`

dm/worker/server.go
lance6716: please note that you haven't fix this

zeminzhou: I have fix this by calling s.startKeepAlive before s.observeSourceBound

dm/worker/server_test.go
```diff
@@ -225,6 +226,131 @@ func (t *testServer) TestServer(c *C) {
...
+	// start worker server
+	s := NewServer(cfg)
+	defer s.Close()
+	go func() {
+		err1 := s.Start()
+		c.Assert(err1, IsNil)
+	}()
+	c.Assert(utils.WaitSomething(30, 100*time.Millisecond, func() bool {
+		return !s.closed.Get()
+	}), IsTrue)
```
lance6716, line 279: maybe if we don't use async start, there's no need to check `s.closed` after `s.Start()` returns nil error

dm/worker/server_test.go
```diff
@@ -225,6 +226,131 @@ func (t *testServer) TestServer(c *C) {
...
+		select {
+		case ev := <-workerEvCh:
+			if !ev.IsDeleted {
+				return true
+			}
+			return false
+		default:
+			return false
+		}
+	}), IsTrue)
```
lance6716, line 292: seems we could move two `return false` out of `select` to eliminate one

dm/worker/server_test.go
```diff
@@ -225,6 +226,131 @@ func (t *testServer) TestServer(c *C) {
...
+	// do check until worker offline
+	c.Assert(utils.WaitSomething(50, 100*time.Millisecond, func() bool {
+		select {
+		case ev := <-workerEvCh:
+			if ev.IsDeleted {
+				return true
+			}
+			return false
+		default:
+			return false
```
lance6716, line 313: ditto

dm/worker/server_test.go
```diff
@@ -225,6 +226,131 @@ func (t *testServer) TestServer(c *C) {
...
+		}
+	}), IsTrue)
+	cancel()
+	wg.Wait()
+
+	//nolint:errcheck
```
lance6716, line 332: I think adding an `c.Assert` to assert below `failpoint.Disable` is better than adding a lint comment

lance6716 (commented):
rest LGTM. good job

lance6716:
and please help check other `TODO: handle fatal error from` need this mechanism

lance6716:
PTAL @3pointer @lichunzhu

dm/worker/join.go
```diff
@@ -110,7 +110,7 @@ func (s *Server) KeepAlive() {
 			return // return if failed to stop the worker.
 		}
 		select {
-		case <-s.ctx.Done():
```
lichunzhu, line 113: Suggest using another variable to save `s.ctx` here to avoid data race.

dm/worker/server.go
```diff
@@ -137,20 +141,24 @@ func (s *Server) Start() error {
...
+		for {
+			err1 := s.observeSourceBound(s.ctx, revBound)
+			if err1 == nil {
+				return
+			}
+			s.stopKeepAlive()
+			select {
+			case <-s.ctx.Done():
+				return
+			case <-time.After(time.Duration(s.cfg.KeepAliveTTL) * time.Second * 2):
```
lichunzhu, line 158: Can we start another keepalive only when we assure the last one is quitted? For example, using another waitgroup.

ti-srebot (approved):

PR #1405, opened 2021-01-27
worker, ha: increase keepalive TTL to 1 minute, and to 30 minutes if relay enabled

lance6716:

make keepalive more rubost

after this PR there're two keepalive TTL: one for relay and one for non-relay. They are 30 minutes and 1 minute seperately by default.
If relay task is assigned, change it to relay-keepalive-ttl, when no relay tasks, chage it to keepalive-ttl

### Check List 

Tests 

 - Integration test

Code changes

Side effects

 - Increased code complexity

Related changes


pkg/ha/keepalive.go
```diff
@@ -112,6 +126,30 @@ func KeepAlive(ctx context.Context, cli *clientv3.Client, workerName string, kee
 		case <-ctx.Done():
 			log.L().Info("ctx is canceled, keepalive will exit now")
 			return nil
+		case newTTL := <-keepAliveUpdateCh:
+			// create a new lease with new TTL, and overwrite original KV
+			cliCtx, cancel := context.WithTimeout(ctx, etcdutil.DefaultRequestTimeout)
+			defer cancel()
+			lease, err = cli.Grant(cliCtx, newTTL)
```
zeminzhou, line 133: The revokeLease function should be called for the old lease, like line 104.

lance6716: lease will expire if not keepalive, so I think it's not needed

lance6716:
PTAL @lichunzhu @3pointer

GMHDBJD:
Our ha integration test still pass after increase keepalive time... 🤔

lance6716:
> Our ha integration test still passed after increasing keepalive ttl... 🤔

I guess it's gracefully exited (not kill -9) so revoke lease function called. will check today

pkg/ha/keepalive.go
```diff
@@ -107,11 +126,47 @@ func KeepAlive(ctx context.Context, cli *clientv3.Client, workerName string, kee
...
+			//nolint:lostcancel
+			keepAliveCtx, keepAliveCancel = context.WithCancel(ctx)
+			ch, err = cli.KeepAlive(keepAliveCtx, leaseID)
+			if err != nil {
+				log.L().Error("meet error when change keepalive TTL", zap.Error(err))
+				keepAliveCancel() // make go vet happy
+				return err
+			}
+			// after new keepalive is succeed, we cancel the old keepalive
+			oldCancel()
```
lichunzhu, line 169: Do we need to revoke the last lease?

tests/incremental_mode/run.sh
```diff
@@ -29,6 +43,19 @@ function run() {
     dmctl_operate_source create $WORK_DIR/source1.yaml $SOURCE_ID1
     dmctl_operate_source create $WORK_DIR/source2.yaml $SOURCE_ID2
 
+    # relay should be started after source bounded
+    sleep 1
+    # and now default keepalive TTL is 30 mimnutes
```
lichunzhu, line 48:
```suggestion
    # and now default keepalive TTL is 30 minutes
```

dm/worker/server.go
```diff
@@ -588,6 +588,13 @@ func (s *Server) startWorker(cfg *config.SourceConfig) error {
 			return err
 		}
 		startRelay = !relayStage.IsDeleted && relayStage.Expect == pb.Stage_Running
+		// change keepalive TTL to 30 minutes if it's default value
+		// is relayStage is not running, we choose to change keepalive here instead of at relayStage switching
+		if s.cfg.KeepAliveTTL == defaultKeepAliveTTL {
```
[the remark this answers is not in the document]
lance6716: I think we should left a chance if user doesn't want this feature, so only increase TTL when default value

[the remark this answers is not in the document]
lichunzhu: Then I think we should add a `configuration` named `relayEnabledKeepAliveTTL`. In the current way, I can't set `keepAliveTTL` and increase `relayKeepAliveTTL` at the same time.

pkg/ha/keepalive.go
```diff
@@ -98,7 +112,12 @@ func KeepAlive(ctx context.Context, cli *clientv3.Client, workerName string, kee
 		}
 	}()
 
-	ch, err := cli.KeepAlive(ctx, lease.ID)
+	keepAliveCtx, keepAliveCancel := context.WithCancel(ctx)
+	defer func() {
+		keepAliveCancel()
+	}()
```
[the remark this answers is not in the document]
lance6716: closure will use the function hold by the variable at exit time, while direct defer will save the function to the stack at call time (we will change that variable, direct defer cannot call changed function)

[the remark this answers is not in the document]
lichunzhu: gotta

pkg/ha/keepalive.go
```diff
@@ -107,11 +126,47 @@ func KeepAlive(ctx context.Context, cli *clientv3.Client, workerName string, kee
 		case _, ok := <-ch:
 			if !ok {
 				log.L().Info("keep alive channel is closed")
+				keepAliveCancel() // make go vet happy
```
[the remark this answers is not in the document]
lance6716: go vet will report an error and fail CI

pkg/ha/keepalive.go
lance6716: after TTL it will be cleaned, I think that's no harm

lichunzhu: What if the last TTL is very big?
For example, we start a source with relay, and then stop this source. After that, we assign a new source without relay to this worker. The new worker will has a `relayEnabledKeepAliveTTL` lease. Is that expected?

lichunzhu (commented):
I still think we can try to revoke the last lease when we "re-keepalive". It doesn't matter even if the job is failed.
Rest LGTM

lance6716:
`revokeLease` is added in https://github.com/pingcap/dm/pull/1405/commits/387d9e1801884cdfe84a9bf74e59718080372fdb, PTAL @zeminzhou @lichunzhu

lichunzhu (commented):
LGTM

zeminzhou:
LGTM

lichunzhu (approved):

PR #1410, opened 2021-01-29
config: add warning, when users set configuration to mydumper/loader/syncer b…

zeminzhou:
…ut not use them

For some configurations like routes/filters/mydumpers/loaders/syncers that users set in global configuration, this pr add “reference count” for these configurations.

When some configurations are set in global configuration, but instances don't use thme. An error will return.

### Tests
+ unit test

### DIFF
--- a/dm/config/task.go
+++ b/dm/config/task.go
@@ -482,7 +482,7 @@ func (c *TaskConfig) adjust() error {
 			*inst.Mydumper = *rule // ref mydumper config
 		}
 		if inst.Mydumper == nil {
-			if len(c.Mydumpers) != 0 {
+			if len(c.Mydumpers) != 0 && inst.MydumperThread == 0 {
 				log.L().Warn("mysql instance don't refer mydumper's configuration with mydumper-config-name, the default configuration will be used", zap.Int("mysql instance", i))
 				return terror.ErrConfigUnreferingCofig.Generate()
 			}
@@ -511,7 +511,7 @@ func (c *TaskConfig) adjust() error {
 			*inst.Loader = *rule // ref loader config
 		}
 		if inst.Loader == nil {
-			if len(c.Loaders) != 0 {
+			if len(c.Loaders) != 0 && inst.LoaderThread == 0 {
 				log.L().Warn("mysql instance don't refer loader's configuration with loader-config-name, the default configuration will be used", zap.Int("mysql instance", i))
 				return terror.ErrConfigUnreferingCofig.Generate()
 			}
@@ -532,7 +532,7 @@ func (c *TaskConfig) adjust() error {
 			*inst.Syncer = *rule // ref syncer config
 		}
 		if inst.Syncer == nil {
-			if len(c.Syncers) != 0 {
+			if len(c.Syncers) != 0 && inst.SyncerThread == 0 {
 				log.L().Warn("mysql instance don't refer syncer's configuration with syncer-config-name, the default configuration will be used", zap.Int("mysql instance", i))
 				return terror.ErrConfigUnreferingCofig.Generate()
 			}
--- a/dm/config/task.go
+++ b/dm/config/task.go
@@ -484,6 +484,7 @@ func (c *TaskConfig) adjust() error {
 		if inst.Mydumper == nil {
 			if len(c.Mydumpers) != 0 {
 				log.L().Warn("mysql instance don't refer mydumper's configuration with mydumper-config-name, the default configuration will be used", zap.Int("mysql instance", i))
+				return terror.ErrConfigUnreferingCofig.Generate()
 			}
 			defaultCfg := defaultMydumperConfig()
 			inst.Mydumper = &defaultCfg
@@ -512,6 +513,7 @@ func (c *TaskConfig) adjust() error {
 		if inst.Loader == nil {
 			if len(c.Loaders) != 0 {
 				log.L().Warn("mysql instance don't refer loader's configuration with loader-config-name, the default configuration will be used", zap.Int("mysql instance", i))
+				return terror.ErrConfigUnreferingCofig.Generate()
 			}
 			defaultCfg := defaultLoaderConfig()
 			inst.Loader = &defaultCfg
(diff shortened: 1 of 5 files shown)


lichunzhu (commented):
I don't think users will check dm-worker's log when they start a task. I think we should print these messages to `result.Msg` or something else.

zeminzhou:
> I don't think users will check dm-worker's log when they start a task. I think we should print these messages to `result.Msg` or something else.

good idea

dm/config/task.go
```diff
@@ -418,6 +418,8 @@ func (c *TaskConfig) adjust() error {
 	}
 
 	iids := make(map[string]int) // source-id -> instance-index
+	gobalConfigReferCount := map[string]int{}
```
lance6716, line 421:
```suggestion
	globalConfigReferCount := map[string]int{}
```

dm/config/task.go
```diff
@@ -472,10 +477,15 @@ func (c *TaskConfig) adjust() error {
...
 				return terror.ErrConfigMydumperCfgNotFound.Generate(i, inst.MydumperConfigName)
 			}
+			gobalConfigReferCount[prefixs[3]+inst.MydumperConfigName]++
 			inst.Mydumper = new(MydumperConfig)
 			*inst.Mydumper = *rule // ref mydumper config
 		}
 		if inst.Mydumper == nil {
+			if len(c.Mydumpers) != 0 && inst.MydumperThread == 0 {
+				log.L().Warn("mysql instance don't refer mydumper's configuration with mydumper-config-name, the default configuration will be used", zap.Int("mysql instance", i))
+				return terror.ErrConfigUnreferingCofig.Generate()
```
lance6716, line 487: If we let DM fallback to default config, we should not return here

and I think check `inst.MydumperThread == 0` is not necessary

dm/config/task.go
```diff
@@ -539,6 +559,42 @@ func (c *TaskConfig) adjust() error {
 		return terror.ErrConfigDuplicateCfgItem.Generate(strings.Join(duplicateErrorStrings, "\n"))
 	}
 
+	for route := range c.Routes {
+		if gobalConfigReferCount[prefixs[0]+route] == 0 {
+			log.L().Error("route in gobal configuration isn't used", zap.String("route", route))
+			return terror.ErrConfigGobalConfigUnused.Generate()
```
lance6716, line 565: we should return the unused configs in error message, grep the names in log is a hard work.

could add the names in a slice and format error message with all of them.

dm/config/task_test.go
```diff
@@ -28,6 +28,314 @@ import (
...
+    syncer-config-name: "global1"
+
+  - source-id: "mysql-replica-02"
+    route-rules: ["route-rule-1"]
+    filter-rules: ["filter-rule-1"]
+    column-mapping-rules: ["column-mapping-rule-1"]
+`
+	taskConfig = NewTaskConfig()
+	err = taskConfig.Decode(errorTaskConfig2)
+	c.Assert(err, NotNil)
```
lance6716, line 336: after we return the unused names, we should use `ErrorMatches` to check them

pkg/terror/error_list.go
```diff
@@ -823,6 +824,7 @@ var (
 	ErrConfigEtcdParse              = New(codeConfigEtcdParse, ClassConfig, ScopeInternal, LevelHigh, "incapable config of %s from etcd", "")
 	ErrConfigMissingForBound        = New(codeConfigMissingForBound, ClassConfig, ScopeInternal, LevelHigh, "source bound %s doesn't have related source config in etcd", "")
 	ErrConfigBinlogEventFilter      = New(codeConfigBinlogEventFilter, ClassConfig, ScopeInternal, LevelHigh, "generate binlog event filter", "Please check the `filters` config in source and task configuration files.")
+	ErrConfigGlobalConfigsUnused    = New(codeConfigGlobalConfigsUnused, ClassConfig, ScopeInternal, LevelHigh, "The configurations as following %v are set in gobal configuration but instances don't use them", "Please check the configuration files.")
```
lance6716, line 827:
```suggestion
	ErrConfigGlobalConfigsUnused    = New(codeConfigGlobalConfigsUnused, ClassConfig, ScopeInternal, LevelHigh, "The configurations as following %v are set in global configuration but instances don't use them", "Please check the configuration files.")
```

dm/config/task.go
```diff
@@ -512,10 +525,14 @@ func (c *TaskConfig) adjust() error {
 			if !ok {
 				return terror.ErrConfigSyncerCfgNotFound.Generate(i, inst.SyncerConfigName)
 			}
+			globalConfigReferCount[prefixs[5]+inst.SyncerConfigName]++
 			inst.Syncer = new(SyncerConfig)
 			*inst.Syncer = *rule // ref syncer config
 		}
 		if inst.Syncer == nil {
+			if len(c.Syncers) != 0 {
+				log.L().Warn("mysql instance don't refer syncer's configuration with syncer-config-name, the default configuration will be used", zap.Int("mysql instance", i))
```
lance6716, line 534: we could use `source-id` to identify it

lance6716 (commented):
rest LGTM

lance6716:
for failed tests, you could sort the slice to get the stable result. rest LGTM

lichunzhu (commented):
/lgtm

dm/config/task_test.go
```diff
@@ -234,7 +234,8 @@ mysql-instances:
 	taskConfig = NewTaskConfig()
 	err = taskConfig.Decode(errorTaskConfig)
 	c.Check(err, NotNil)
-	c.Assert(err, ErrorMatches, "*The configurations as following [route-rule-2 filter-rule-2 column-mapping-rule-2] are set in global configuration but*")
+	//c.Assert(err, ErrorMatches, "*The configurations as following [route-rule-2 filter-rule-2 column-mapping-rule-2] are set in global configuration but*")
```
lance6716, line 237: this test can't detect only one of reference check not working. `sort.Strings` is a better way

and please remove the commented line above to keep codebase clean

lichunzhu (commented):
LGTM

lichunzhu (approved):
LGTM

PR #1420, opened 2021-02-03
auto master-slave switch (#1364)

ti-srebot:
cherry-pick #1364 to release-2.0
You can switch your code base to this Pull Request by using [git-extras](https://github.com/tj/git-extras):
```bash
# In dm repo:
git pr https://github.com/pingcap/dm/pull/1420
```

After apply modifications, you can push your change to this PR via:
```bash
git push git@github.com:ti-srebot/dm.git pr/1420:release-2.0-b04afa753844
```

---

### What problem does this PR solve? 
- [x] syncer auto switch relay subdir when master-slave switch
- [x] auto test master-slave switch
- [x] auto detect master-slave switch

Tests 

 - Unit test
 - github action

PR #1364, opened 2020-12-28
auto master-slave switch

GMHDBJD:
### What problem does this PR solve? 
- [x] syncer auto switch relay subdir when master-slave switch
- [x] auto test master-slave switch
- [x] auto detect master-slave switch

Tests 

 - Unit test
 - github action

PR #1421, opened 2021-02-03
streamer: fix unstable CI introduced in #1364

lance6716:

in #1364, channels are closed in a goroutine when context is done, so in afterwards `select { case v := <- ch` we will read a `v` of zero value, which will cause wrong logic

https://internal.pingcap.net/idc-jenkins/blue/organizations/jenkins/dm_ghpr_test/detail/dm_ghpr_test/8147/pipeline/72/

we don't need to close channels if they are no longer referenced, gc will correctly clean them. (but if the values in channel need a explict `Close`, we should drain the channel and `Close` each one)

https://nanxiao.gitbooks.io/golang-101-hacks/content/posts/need-not-close-every-channel.html

### Check List 

Tests 

 - hard to test

Code changes

Side effects

Related changes

 - Need to cherry-pick to the release branch (will cherry-pick to https://github.com/pingcap/dm/pull/1420 manually)

### DIFF
--- a/pkg/streamer/reader.go
+++ b/pkg/streamer/reader.go
@@ -569,21 +569,13 @@ func (r *BinlogReader) parseFile(
 
 	wg.Add(1)
 	go func(latestPos int64) {
-		defer func() {
-			close(switchCh)
-			close(switchErrCh)
-			wg.Done()
-		}()
+		defer wg.Done()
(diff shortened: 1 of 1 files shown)


lance6716:
@lichunzhu @3pointer @zeminzhou PTAL

lichunzhu (approved):
LGTM

PR #1276, opened 2020-11-05
fix: always wait for the sender to exit in the receiver for load unit

csuzhangxc:
### What problem does this PR solve? 

fix part2 of #1262, the scenario:

1. worker-A reported error via `runFatalChan`
2. cancel the context
3. chan received in worker-B exit
4. chan sender in worker-B blocked

### What is changed and how it works?

always wait for the sender to exit in the receiver for load unit

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

 - Increased code complexity

Related changes

 - Need to cherry-pick to the release branch
 - Need to be included in the release note

PR #1442, opened 2021-02-20
load: don't block when worker cancel

3pointer:
Don't block on send jobs to jobQueue when context is canceled.

Add select statement on jobQueue

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Related changes

### Releasae note
- Fix the issue that the load unit might block when context canceled.


loader/loader.go
```diff
@@ -364,7 +364,10 @@ func (w *Worker) dispatchSQL(ctx context.Context, file string, offset int64, tab
 			}
 			lastOffset = cur
 
-			w.jobQueue <- j
+			select {
```
overvenus, line 367: What about load/loader.go:250? Should we `select ctx.Done` as well?

3pointer: yes, we should add it too, I'll fix it.

GMHDBJD:
Has this issue been fixed in #1276 after 2.0.1?

lance6716:
> Has this issue been fixed in #1276 in 2.0.1?

that pr makes the receiving end of channel could exit, but the sending end may block because the channel is full.

lance6716:
replaced by https://github.com/pingcap/dm/pull/1276

PR #1393, opened 2021-01-20
syncer, relay: support incremental task with gtid specified only

lance6716:
### What problem does this PR solve? 
close #1391 

### What is changed and how it works?
don't update active relay log (only used for relay log purging?) when there's no binlog name which may because user only specifiy gtid only. Active relay log will be updated at a XID or DDL event with valid binlog name afterwards.

### Check List 

Tests 

 - Integration test
 
Code changes

Side effects

Related changes

 - Need to cherry-pick to the release branch
 - Need to be included in the release note

PR #1404, opened 2021-01-27
streamer: add heartbeat for local streamer

lance6716:
### What problem does this PR solve? 

currently relay will ignore heartbeat event before writing to relay logs. so when relay enabled, there's no heartbeat event which causes syncer can't advance its checkpoint when upstream is idle.

### What is changed and how it works?

peroidically send heartbeat event

### Check List 

Tests 

 - Unit test
 - Manually test (only do an insert in upstream, and check checkpoint advanced after 30s)

Code changes

Side effects

Related changes

 - Need to cherry-pick to the release branch
 - Need to be included in the release note

PR #1448, opened 2021-02-22
syncer: update active relay log at fake rotate event

lance6716:

in https://github.com/pingcap/dm/pull/1393, we support only specify GTID with no binlog filename, so we can't update active relay log filename at init time. Although active relay log will be updated when after `checkpoint-flush-interval` seconds and receiving an event, for the first `checkpoint-flush-interval` seconds, purge-relay will fail because of no active relay log, which may confuse user.

when only specify GTID with no binlog filename, use fake rotate event to update active relay log.

also, add a integration test to make sure active relay log is forwarded when there's only DML

### Check List 

Tests 

 - Integration test

Code changes

Side effects

Related changes


lance6716:
in https://github.com/pingcap/dm/pull/1404, heartbeat is sent on 30 seconds no upstream events generated. But user configurates only sync DML, enables relay, and upstream keeps generating binlog events, maybe there's no chance to flush global checkpoint and update active relay log. Will check tomorrow

lichunzhu (commented):
The integration test name only_dml is confusing. Personally suggest changing to purge_relay_active or something else.
Rest LGTM.

lance6716:
> The integration test name only_dml is confusing. Personally suggest changing to purge_relay_active or something else.
> Rest LGTM.

this test aims covering a use case of only synchronize dml, which is specified in release testing doc.

lichunzhu (commented):
/lgtm

GMHDBJD (approved):
/lgtm

Issue #1423, opened 2021-02-03
disable heartbeat feature becauase it may block HA


lance6716:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do? If possible, provide a recipe for reproducing the error.
https://asktug.com/t/topic/68351

2. What did you expect to see?

3. What did you see instead?

4. Versions of the cluster

PR #1467, opened 2021-03-01
config: disable heartbeat feature

lance6716:

close https://github.com/pingcap/dm/issues/1423

overwrite it to `false`

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes


dm/config/task.go
```diff
@@ -626,7 +626,9 @@ func (c *TaskConfig) SubTaskConfigs(sources map[string]DBConfig) ([]*SubTaskConf
 		cfg.Mode = c.TaskMode
 		cfg.CaseSensitive = c.CaseSensitive
 		cfg.MetaSchema = c.MetaSchema
-		cfg.EnableHeartbeat = c.EnableHeartbeat
```
GMHDBJD, line 629: How about set it `false` **explicitly**.

GMHDBJD (commented):
reset LGTM

GMHDBJD:
Will there be the same problem when upgrading from v1 to v2 via tiup？

lance6716:
> Will there be the same problem when upgrading from v1 to v2 via tiup？

I think the `import_v10x` test could ensure it

GMHDBJD:
> I think the `import_v10x` test could ensure it

Seems the `enable-heartbeat` config from [get-config](https://github.com/pingcap/dm/blob/511cea1ea51822274f6ad7fe4acdc8ec948de8f3/tests/import_v10x/conf/task.yaml#L7) still true?

lance6716:
> > I think the `import_v10x` test could ensure it
> 
> Seems the `enable-heartbeat` config from [get-config](https://github.com/pingcap/dm/blob/511cea1ea51822274f6ad7fe4acdc8ec948de8f3/tests/import_v10x/conf/task.yaml#L7) still true?

updated in https://github.com/pingcap/dm/pull/1467/commits/453f2b374105b9c756855c6844fec09db03e464f

GMHDBJD (commented):
LGTM

lichunzhu (approved):
LGTM

PR #1468, opened 2021-03-01
syncer: update active relay log at fake rotate event (#1448)

ti-srebot:
cherry-pick #1448 to release-2.0
You can switch your code base to this Pull Request by using [git-extras](https://github.com/tj/git-extras):
```bash
# In dm repo:
git pr https://github.com/pingcap/dm/pull/1468
```

After apply modifications, you can push your change to this PR via:
```bash
git push git@github.com:ti-srebot/dm.git pr/1468:release-2.0-1ec8439a21fe
```

---

in https://github.com/pingcap/dm/pull/1393, we support only specify GTID with no binlog filename, so we can't update active relay log filename at init time. Although active relay log will be updated when after `checkpoint-flush-interval` seconds and receiving an event, for the first `checkpoint-flush-interval` seconds, purge-relay will fail because of no active relay log, which may confuse user.

when only specify GTID with no binlog filename, use fake rotate event to update active relay log.

also, add a integration test to make sure active relay log is forwarded when there's only DML

### Check List 

Tests 

 - Integration test

Code changes

Side effects

Related changes


GMHDBJD (commented):
/lgtm

lichunzhu (approved):
/lgtm

lichunzhu (approved):
LGTM

PR #1489, opened 2021-03-08
error: show origin ddl when error

GMHDBJD:
show origin ddl when ddl error occur, so user can use handle-error replace with them

return origin ddl when error

### Check List 

Tests 

 - Integration test

### DIFF
--- a/syncer/syncer.go
+++ b/syncer/syncer.go
@@ -1709,7 +1709,7 @@ func (s *Syncer) handleRowsEvent(ev *replication.RowsEvent, ec eventContext) err
 }
 
 func (s *Syncer) handleQueryEvent(ev *replication.QueryEvent, ec eventContext, originSQL string) error {
-	if originSQL == "BEGIN"{
+	if originSQL == "BEGIN" {
 		return nil
 	}
 
--- a/syncer/syncer.go
+++ b/syncer/syncer.go
@@ -941,7 +941,7 @@ func (s *Syncer) syncDDL(tctx *tcontext.Context, queueBucket string, db *DBConn,
 		if err != nil {
 			s.execError.Set(err)
 			if !utils.IsContextCanceledError(err) {
-				err = s.handleEventError(err, sqlJob.startLocation, sqlJob.currentLocation, true)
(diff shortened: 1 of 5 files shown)


syncer/syncer.go
```diff
@@ -1707,26 +1708,27 @@ func (s *Syncer) handleRowsEvent(ev *replication.RowsEvent, ec eventContext) err
 	return nil
 }
 
-func (s *Syncer) handleQueryEvent(ev *replication.QueryEvent, ec eventContext) error {
+func (s *Syncer) handleQueryEvent(ev *replication.QueryEvent, ec eventContext, originSQL *string) error {
 	if bytes.Equal(ev.Query, []byte("BEGIN")) {
 		return nil
 	}
-	sql := strings.TrimSpace(string(ev.Query))
+	*originSQL = strings.TrimSpace(string(ev.Query))
```
lichunzhu, line 1715: What about set originSQL to `strings.TrimSpace(string(ev.Query))` before we entering this function instead of using a pointer?

syncer/syncer.go
```diff
@@ -2792,12 +2794,15 @@ func (s *Syncer) getErrLocation() (*binlog.Location, bool) {
...
 
-func (s *Syncer) handleEventError(err error, startLocation, endLocation binlog.Location, isQueryEvent bool) error {
+func (s *Syncer) handleEventError(err error, startLocation, endLocation binlog.Location, isQueryEvent bool, originSQL string) error {
 	if err == nil {
 		return nil
 	}
 
 	s.setErrLocation(&startLocation, &endLocation, isQueryEvent)
+	if len(originSQL) > 0 {
+		return terror.Annotatef(err, "startLocation: [%s], endLocation: [%s], origin DDL: [%s]", startLocation, endLocation, originSQL)
```
lichunzhu, line 2804:
```suggestion
		return terror.Annotatef(err, "startLocation: [%s], endLocation: [%s], origin SQL: [%s]", startLocation, endLocation, originSQL)
```

lichunzhu (commented):
Rest LGTM

ti-srebot (approved):

PR #1309, opened 2020-11-27
loader: support ANSI_QUOTES when sharing rename

lance6716:
### What problem does this PR solve? 
original code use backquoted scham/table names to do string replace , not supporting upstream has ANSI_QUOTES SQL mode.

### What is changed and how it works?
switch to another quote

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

Related changes

 - Need to cherry-pick to the release branch
 - Need to be included in the release note

PR #1497, opened 2021-03-10
loader: don't use ANSI_QUOTES for dumped data files

lance6716:
in https://github.com/pingcap/dm/pull/1309, condition branch of dumped data file is not tested because of column-mapping, and it has a BUG

don't use ANSI_QUOTES for dumped data files

### Check List 

Tests 

 - Integration test
 
Related changes


GMHDBJD (approved):
/lgtm

PR #1504, opened 2021-03-11
worker: move relay related function together

lance6716:
only a refactoring, support control worker only enable functionality of relay (or subtask, in future), so we could control the behaviour of worker more flexible

move relay-related function into one function`EnableRelay`

### Check List 

Tests 

 - pass original test

Code changes

Side effects

Related changes


lance6716:
replaced by  https://github.com/pingcap/dm/pull/1507

PR #1507, opened 2021-03-12
worker: move relay/subtask related function together

lance6716:
only a refactoring, support control worker only enable functionality of relay or subtask. so we could control the behaviour of worker more flexible

move relay-related function into one function`EnableRelay`, and subtask-relayed function into `EnableHandleSubtasks`

### Check List 

Tests 

 - unit test
 - pass original test

Code changes

Side effects

Related changes


dm/worker/worker.go
```diff
@@ -190,6 +170,114 @@ func (w *Worker) Close() {
...
+// EnableRelay enables the functionality of start/watch/handle relay
+func (w *Worker) EnableRelay() error {
+	// 1. adjust relay starting position, to the earliest of subtasks
+	_, subTaskCfgs, _, err := ha.GetSubTaskStageConfig(w.etcdClient, w.cfg.SourceID)
+	if err != nil {
+		return err
+	}
+	if err = copyConfigFromSourceForEach(subTaskCfgs, w.cfg); err != nil {
+		return err
+	}
```
GMHDBJD, line 182: How about combine them with L246-253

GMHDBJD (commented):
rest LGTM

dm/worker/worker.go
lance6716: OK, and I found in https://github.com/pingcap/dm/blob/8157e33bd21d85504ba533d7ab5dc1cb26417823/dm/worker/worker.go#L301-L309
we forget to `copyConfigFromSourceForEach`.

lance6716: https://github.com/pingcap/dm/pull/1507/commits/c151ea0a21a0cabca26aa272a97821f00a4858d9

ti-srebot (dismissed):

lichunzhu (approved):

PR #1510, opened 2021-03-15
shardddl/optimistic: warn add not fully dropped columns

lichunzhu:
When users adds a column that isn't fully dropped in the downstream database, it will cause a data in-consistent problem.
dm-master will fill to synchronize optimistic shard DDLs after a restart or leader transfer.

1. Report an error to the user when they want to add a not fully dropped column.
When we drop a column, we save this column's name in etcd. When it's fully dropped in downstream database and **we receive a `Done` DDL group from dm-worker**, we can delete this column's name from the map and the etcd.
    Add source, upstreamSchema/upstreamTable to avoid columns affect others.

**Time series:** (+a/-a means add/drop column a)
```
              older ----------------> newer
  tb1: +a +b +c                  -c
  tb2:                                       +a +b +c
  tb3:                 +a +b +c
```
If we added not fully dropped column c for [tb1], tb2 may report an error if we don't identify tb1 and tb2.

2. Fix the bug that dm-master doesn't have complete lock infomation if we meet an error when we recovering locks.
https://github.com/pingcap/dm/pull/1510/files#diff-265ec2690b9027cfdc9cb7f36dda664d0337e6a1b0149e2e058720aee099336eL273-L275
If we stopped recovering locks once we meet an error, dm-master leader may not have the full information for the other normal locks, which will cause an error in dm-worker.
After this PR, we never stop recovering locks even if we met an error in handling locks.

3. Fix the bug that dm-worker may get blocked if dm-master restarted before dm-worker received the optimistic shard ddl operation (after dm-worker put the optimistic shard ddl info).

**Time series:**
a. dm-worker marked done a operation before (with the help of dm-master)
b. dm-worker received ddl
c. dm-worker putted ddl info successfully
d. dm-master restarted before giving the ddl operation to dm-worker
e. dm-master restarted successfully but didn't put operation because this operation is marked as done.
f. dm-worker didn't receive the operation, will get blocked forever.

4. (**Not fixed yet**) dm-master can't return correct DDLs when dm-master rebuild locks after a leader transfer or a restart.

**CASE:**

tb1: a         info: (DDLs: [add column c],  tiBefore: [create table tb1 a int])
tb2: a b c  info: (DDLs: [drop column c], tiBefore: [create table tb1 a int, b int, c int]) operation: done

Now dm-master will use tb2's schema as joined, and use joined for newly added table (tb1). If we recover info for tb2 at first, we will set tb1 table info as tb2's, which will cause dm-master to make the wrong decision (do nothing for `add column c`). Then we will fail in the downstream.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes


lance6716:
there're two failed test case, seems one is tidb again changed the `SHOW CREATE TABLE` behaviour for CLUSTERED INDEX. we could change our matching string

tests/shardddl3/run.sh
```diff
@@ -993,6 +999,92 @@ function DM_RestartMaster() {
...
+    run_sql_source1 "insert into ${shardddl1}.${tb1} values(4);"
+
+    restart_master_on_pos $reset "2"
+
+    run_sql_source2 "alter table ${shardddl1}.${tb1} drop column c;"
+    run_sql_source2 "insert into ${shardddl1}.${tb1} values(5,5);"
+
+    restart_master_on_pos $reset "3"
+
+    # make sure drop column c is synced
```
lance6716, line 1031: may cause a unstable test, could we use some check and retry?

lance6716 (commented):
(review later)

pkg/shardddl/optimism/column.go
```diff
@@ -0,0 +1,101 @@
...
+
+import (
+	"go.etcd.io/etcd/clientv3"
+
+	"github.com/pingcap/dm/dm/common"
+	"github.com/pingcap/dm/pkg/etcdutil"
+)
+
+// GetAllDroppedColumns gets the all dropped columns.
+func GetAllDroppedColumns(cli *clientv3.Client) (map[string]map[string]map[string]map[string]map[string]interface{}, int64, error) {
```
lance6716, line 24: why the final value of that big map is `interface{}` not `struct{}`

dm/master/shardddl/optimist.go
```diff
@@ -245,35 +245,45 @@ func (o *Optimist) rebuildLocks() (revSource, revInfo, revOperation int64, err e
...
 // recoverLocks recovers shard DDL locks based on shard DDL info and shard DDL lock operation.
 func (o *Optimist) recoverLocks(
 	ifm map[string]map[string]map[string]map[string]optimism.Info,
-	opm map[string]map[string]map[string]map[string]optimism.Operation) error {
+	opm map[string]map[string]map[string]map[string]optimism.Operation,
+	colm map[string]map[string]map[string]map[string]map[string]interface{}) error {
 	// construct locks based on the shard DDL info.
-	for task, ifTask := range ifm {
+	o.lk.SetColumnMap(colm)
+	defer o.lk.SetColumnMap(nil)
```
lance6716, line 273: (I guess this does not effect correctness? OK it may release some memory)

pkg/shardddl/optimism/lock.go
```diff
@@ -501,27 +526,123 @@ func (l *Lock) GetVersion(source string, schema string, table string) int64 {
...
+	col, err := GetColumnName(lockID, ddl, ast.AlterTableAddColumns)
+	if err != nil {
+		return col, err
+	}
+	if len(col) > 0 {
+		oldJoinedCols := schemacmp.DecodeColumnFieldTypes(oldJoined)
+		newJoinedCols := schemacmp.DecodeColumnFieldTypes(newJoined)
+		oldCol, ok1 := oldJoinedCols[col]
+		newCol, ok2 := newJoinedCols[col]
+		if ok1 && ok2 {
```
lance6716, line 616: we could combine this two level if

pkg/shardddl/optimism/lock.go
```diff
@@ -501,27 +526,123 @@ func (l *Lock) GetVersion(source string, schema string, table string) int64 {
...
+			delete(l.columns, col)
+		}
+	}
+
+	return nil
+}
+
 // AddDifferentFieldLenColumns checks whether dm adds columns with different field lengths
-func AddDifferentFieldLenColumns(lockID, ddl string, oldJoined, newJoined schemacmp.Table) error {
+func AddDifferentFieldLenColumns(lockID, ddl string, oldJoined, newJoined schemacmp.Table) (string, error) {
```
lance6716, line 606: could we change `schemacmp`, to let it fail different field length join?

pkg/shardddl/optimism/lock_test.go
```diff
@@ -1532,6 +1540,123 @@ func (t *testLock) TestAddDifferentFieldLenColumns(c *C) {
...
+	c.Assert(colm, DeepEquals, colm2)
+
+	// TrySync for the first table, add column b, should succeed, because this column is fully dropped in the downstream
+	DDLs, err = l.TrySync(newInfoWithVersion(task, source, db, tbls[0], downSchema, downTable, DDLs3, ti2, []*model.TableInfo{ti1}, vers), tts)
+	c.Assert(err, IsNil)
+	c.Assert(DDLs, DeepEquals, DDLs3)
+	c.Assert(l.versions, DeepEquals, vers)
+	c.Assert(l.IsResolved(), IsFalse)
+
+	// TrySync for the first table, add column b, should fail, because this column isn't fully dropped in the downstream
```
lance6716, line 1638:
```suggestion
	// TrySync for the first table, add column c, should fail, because this column isn't fully dropped in the downstream
```

pkg/shardddl/optimism/lock_test.go
```diff
@@ -1532,6 +1540,123 @@ func (t *testLock) TestAddDifferentFieldLenColumns(c *C) {
...
+	_, err = l.TrySync(newInfoWithVersion(task, source, db, tbls[0], downSchema, downTable, DDLs4, ti1, []*model.TableInfo{ti0}, vers), tts)
+	c.Assert(err, ErrorMatches, ".*add column c that wasn't fully dropped in downstream.*")
+	c.Assert(l.IsResolved(), IsFalse)
+
+	// TrySync for the second table, drop column b, this column should be dropped
```
lance6716, line 1643: repeated? we didn't add column b after line 1618 🤔

dm/master/shardddl/optimist.go
lichunzhu: No, this will affect correctness. We only use this column map when we recover lock tables.

pkg/shardddl/optimism/lock.go
lichunzhu: If we do so, we will fail when we alter the field length to a large one.

lance6716: again I forget it 😵

dm/master/shardddl/optimist.go
```diff
@@ -245,35 +245,49 @@ func (o *Optimist) rebuildLocks() (revSource, revInfo, revOperation int64, err e
...
+	var firstErr error
```
lance6716, line 274: seems it's not used

pkg/shardddl/optimism/lock.go
lichunzhu: addressed in https://github.com/pingcap/dm/pull/1510/commits/baee6facb5a910f73e6e9462091240a60d7800ec

pkg/shardddl/optimism/lock_test.go
lichunzhu: https://github.com/pingcap/dm/pull/1510/commits/fb5857aef167f500d7f5250f36cf77b11e3d889c

pkg/shardddl/optimism/lock_test.go
lichunzhu: https://github.com/pingcap/dm/pull/1510/commits/fb5857aef167f500d7f5250f36cf77b11e3d889c

dm/master/shardddl/optimist.go
```diff
@@ -291,6 +305,11 @@ func (o *Optimist) recoverLocks(
 					}
 					if op.Done {
 						lock.TryMarkDone(op.Source, op.UpSchema, op.UpTable)
+						err := lock.DeleteColumnsByDDLs(op.DDLs)
```
lance6716, line 308: hope we could left more comment, I think I might forget the logic somedays later

```
if DM-master sent a DROP COLUMN DDL, all shard tables had dropped that column and got synced. So we delete it from paritially dropped columns
```

pkg/shardddl/optimism/lock.go
```diff
@@ -501,27 +526,121 @@ func (l *Lock) GetVersion(source string, schema string, table string) int64 {
...
+		l.columns[col][source] = make(map[string]map[string]interface{})
+	}
+	if _, ok := l.columns[col][source][upSchema]; !ok {
+		l.columns[col][source][upSchema] = make(map[string]interface{})
+	}
+	l.columns[col][source][upSchema][upTable] = struct{}{}
+	return nil
+}
+
+// DeleteColumnsByDDLs deletes the dropped columns by DDLs.
```
lance6716, line 573:
```suggestion
// DeleteColumnsByDDLs deletes the partially dropped columns that extracted from DDLs.
```

pkg/shardddl/optimism/column.go
```diff
@@ -0,0 +1,105 @@
...
+			if _, ok := colm[lockID][column][source][upSchema]; !ok {
+				colm[lockID][column][source][upSchema] = make(map[string]interface{})
+			}
+			colm[lockID][column][source][upSchema][upTable] = struct{}{}
+		}
+	}
+	return colm, rev, nil
+}
+
+// PutDroppedColumn puts the undropped column name into ectd.
```
lance6716, line 66:
```suggestion
// PutDroppedColumn puts the partially dropped column name into ectd.
```

pkg/shardddl/optimism/column.go
```diff
@@ -0,0 +1,105 @@
...
+	op := clientv3.OpPut(key, "")
+
+	resp, rev, err := etcdutil.DoOpsInOneTxnWithRetry(cli, op)
+	if err != nil {
+		return 0, false, err
+	}
+	return rev, resp.Succeeded, nil
+}
+
+// DeleteDroppedColumns tries to delete the dropped columns for the specified lock ID.
```
lance6716, line 81:
```suggestion
// DeleteDroppedColumns tries to delete the paritially dropped columns for the specified lock ID.
```

pkg/shardddl/optimism/column.go
```diff
@@ -0,0 +1,105 @@
...
+// Only when this column is fully dropped in downstream database,
+// in other words, **we receive a `Done` DDL group from dm-worker)**,
```
lance6716, line 83:
```suggestion
// in other words, **we receive a `Done` operation from dm-worker**,
```

pkg/shardddl/optimism/column.go
```diff
@@ -0,0 +1,105 @@
...
+	return rev, resp.Succeeded, nil
+}
+
+// deleteDroppedColumnOp returns a DELETE etcd operation for the specified task and column name.
+func deleteDroppedColumnByColumnOp(task, downSchema, downTable, column string) clientv3.Op {
+	return clientv3.OpDelete(common.ShardDDLOptimismDroppedColumnsKeyAdapter.Encode(task, downSchema, downTable, column), clientv3.WithPrefix())
+}
+
+// deleteDroppedColumnsByLockOp returns a DELETE etcd operation for the specified task.
+func deleteDroppedColumnsByLockOp(task, downSchema, downTable string) clientv3.Op {
```
lance6716, line 103:
```suggestion
func deleteDroppedColumnsByTaskOp(task string) clientv3.Op {
```

pkg/shardddl/optimism/lock_test.go
```diff
@@ -30,6 +31,7 @@ var _ = Suite(&testLock{})
 
 func (t *testLock) SetUpSuite(c *C) {
 	c.Assert(log.InitLogger(&log.Config{}), IsNil)
+	clearTestInfoOperation(c)
```
lance6716, line 34: could we use a defer like other place?

pkg/shardddl/optimism/column.go
lichunzhu: What's the difference? What will we benefit if we use `struct{}`?

lance6716: compared to interface{}, struct{} uses no memory, and don't need a type assertion when assign (don't know if compiler will eliminate this)

and in rest of DM code we use struct{} as map value

pkg/shardddl/optimism/column.go
lichunzhu: This part wants to delete etcd info for a lock. I will update the above code.

pkg/shardddl/optimism/lock_test.go
lichunzhu: I have moved this to TearDownSuite().

dm/master/shardddl/optimist.go
lichunzhu: added under this function's definition.

pkg/shardddl/optimism/lock.go
```diff
@@ -56,12 +59,17 @@ type Lock struct {
 
 	// upstream source ID -> upstream schema name -> upstream table name -> info version.
 	versions map[string]map[string]map[string]int64
+
+	// record the partially dropped columns
+	// column name -> source -> upSchema -> upTable -> interface{}
```
lance6716, line 64:
```suggestion
	// column name -> source -> upSchema -> upTable -> struct{}
```

pkg/shardddl/optimism/lock.go
```diff
@@ -33,6 +34,8 @@ import (
 type Lock struct {
 	mu sync.RWMutex
 
+	cli *clientv3.Client
```
lance6716, line 37: How about we let Optimist do etcd communications, and continue making lock's duty is only recording upstream status. This could be done via TrySync/DeleteColumnsByDDLs returning something to tell Optimist put/delete it

pkg/shardddl/optimism/lock.go
```diff
@@ -207,6 +216,12 @@ func (l *Lock) TrySync(info Info, tts []TargetTable) (newDDLs []string, err erro
 		// if any real conflicts after joined exist, they will be detected by the following steps.
 		var cmp int
 		if cmp, err = nextTable.Compare(oldJoined); err == nil && cmp == 0 {
```
lance6716, line 218: Now this special case causes code duplication, very annoying 😭

pkg/shardddl/optimism/lock.go
```diff
@@ -264,15 +279,25 @@ func (l *Lock) TrySync(info Info, tts []TargetTable) (newDDLs []string, err erro
 
 		cmp, _ = prevTable.Compare(nextTable) // we have checked `err` returned above.
 		if cmp < 0 {
-			// check for add column with different field lengths
-			err = AddDifferentFieldLenColumns(l.ID, ddls[idx], nextTable, newJoined)
-			if err != nil {
+			// check for add column with a smaller field len
```
lance6716, line 282:
```suggestion
			// check for add column with a different field len
```

🤔

pkg/shardddl/optimism/lock.go
```diff
@@ -501,27 +526,123 @@ func (l *Lock) GetVersion(source string, schema string, table string) int64 {
...
+	}
+	if _, ok := l.columns[col][source][upSchema]; !ok {
+		l.columns[col][source][upSchema] = make(map[string]struct{})
+	}
+	l.columns[col][source][upSchema][upTable] = struct{}{}
+	return nil
+}
+
+// DeleteColumnsByDDLs deletes the partially dropped columns that extracted from DDLs.
+// We can remove columns from the partially dropped columns map unless this column is dropped in the downstream database,
```
lance6716, line 574:
```suggestion
// We can not remove columns from the partially dropped columns map unless this column is dropped in the downstream database,
```

pkg/shardddl/optimism/keeper.go
```diff
@@ -46,8 +50,13 @@ func (lk *LockKeeper) TrySync(info Info, tts []TargetTable) (string, []string, e
 	defer lk.mu.Unlock()
 
 	if l, ok = lk.locks[lockID]; !ok {
-		lk.locks[lockID] = NewLock(lockID, info.Task, info.DownSchema, info.DownTable, info.TableInfoBefore, tts)
+		lk.locks[lockID] = NewLock(cli, lockID, info.Task, info.DownSchema, info.DownTable, info.TableInfoBefore, tts)
 		l = lk.locks[lockID]
+		if lk.colm != nil {
```
lance6716, line 55: could we move this "recovering columns" to `recoverLocks`? We could iterate every locks and check `colm`. It could also remove `SetColumnMap`

lance6716 (commented):
rest lgtm

pkg/shardddl/optimism/lock.go
lichunzhu: This comment means we add a column with a larger field length before, and then we add a field len with a smaller one.
For example, `alter table t add column (b varchar(11))`(synced) and `alter table t add column (b varchar(6))` (current one).

lance6716: where's logic we let "larger" field length pass 🤔

lichunzhu: here https://github.com/pingcap/dm/pull/1510/files#diff-110e5d790ae9426fcd1453f3e2a61bb5de5440c0e6ce8dd76d72de4fe2d16432R254

pkg/shardddl/optimism/lock.go
lance6716: (will refactor later)

gozssky:
LGTM

pkg/shardddl/optimism/lock.go
```diff
@@ -207,6 +216,12 @@ func (l *Lock) TrySync(info Info, tts []TargetTable) (newDDLs []string, err erro
...
 		var cmp int
 		if cmp, err = nextTable.Compare(oldJoined); err == nil && cmp == 0 {
+			if col, err := GetColumnName(l.ID, ddls[idx], ast.AlterTableAddColumns); err != nil {
+				return newDDLs, err
+			} else if len(col) > 0 && l.IsDroppedColumn(info, col) {
+				return newDDLs, terror.ErrShardDDLOptimismTrySyncFail.Generate(
+					l.ID, fmt.Sprintf("add column %s that wasn't fully dropped in downstream. ddl: %s", col, ddls[idx]))
+			}
 			newDDLs = append(newDDLs, ddls[idx])
 			continue
```
GMHDBJD, line 226: How do it reach this case?

tests/shardddl3/run.sh
```diff
@@ -984,6 +990,93 @@ function DM_RestartMaster() {
...
+        "\"stage\": \"Running\"" 3 \
+        "\"unit\": \"Sync\"" 2
+
+    run_sql_source1 "alter table ${shardddl1}.${tb1} add column b int after a;"
+
+    restart_master_on_pos $reset "5"
+
+    run_dm_ctl_with_retry $WORK_DIR "127.0.0.1:$MASTER_PORT" \
+        "query-status test" \
+        "because schema conflict detected" 1
```
GMHDBJD, line 1043: Check for specific error messages, "wasn't fully dropped in downstream"

GMHDBJD (commented):
rest LGTM

pkg/shardddl/optimism/keeper.go
lance6716: @lichunzhu here's a comment waiting to address

lance6716:
many conflict 😂

dm/master/shardddl/optimist.go
```diff
@@ -312,26 +320,32 @@ func (o *Optimist) buildLockJoinedAndTTS(ifm map[string]map[string]map[string]ma
...
+	colm map[string]map[string]map[string]map[string]map[string]struct{}) error {
+	o.lk.SetColumnMap(colm)
+	defer o.lk.SetColumnMap(nil)
 	// construct joined table based on the shard DDL info.
 	o.logger.Info("build lock joined and tts")
 	lockJoined, lockTTS := o.buildLockJoinedAndTTS(ifm)
 	// build lock and restore table info
 	o.logger.Info("rebuild locks and tables")
-	o.lk.RebuildLocksAndTables(ifm, lockJoined, lockTTS)
+	o.lk.RebuildLocksAndTables(o.cli, ifm, lockJoined, lockTTS)
```
lance6716, line 332: I mean remove `SetColumnMap`. Instead, iterate through `lk.locks` and `colm` and set them

lance6716 (commented):
rest LGTM

pkg/shardddl/optimism/lock.go
```diff
@@ -514,27 +539,123 @@ func (l *Lock) GetVersion(source string, schema string, table string) int64 {
...
+	}
+	return true
+}
+
+// AddDroppedColumn adds a dropped column name in both etcd and lock's column map
+func (l *Lock) AddDroppedColumn(info Info, col string) error {
+	if l.IsDroppedColumn(info, col) {
+		return nil
+	}
+	log.L().Debug("add partially dropped columns", zap.Stringer("info", info), zap.String("column", col))
```
lichunzhu, line 565:
```suggestion
	log.L().Debug("add partially dropped columns", zap.String("column", col), zap.Stringer("info", info))
```

lichunzhu:
@GMHDBJD PTAL again, thanks!

GMHDBJD (approved):
/lgtm

GMHDBJD:
maybe we should add
```
	mockCluster := integration.NewClusterV3(t, &integration.ClusterConfig{Size: 1})
	defer mockCluster.Terminate(t)

	etcdTestCli = mockCluster.RandClient()
```
in master/optimism_test.go

lichunzhu:
> ```
> 	etcdTestCli = mockCluster.RandClient()
> ```

@GMHDBJD Did this unit test failed because of "etcd race"?

GMHDBJD:
I think so.

lichunzhu:
Okay. I will take a look and fix this.

PR #1512, opened 2021-03-16
sync: return an error for rename table statement in optimistic mode

GMHDBJD:
we do not support rename table now, simply return an error in optimistic mode.
For pessimistic mode, though shardddl group was wrong, dm works well and data are still consistent 🙈.

TODO: support rename table for all mode.
prefilter before handle statement in optimistic mode.

### Check List 

Tests 
 - Integration test

TODO: emphasize don't mixed sharding tables with normal tables in one task, in other words,`sharding-mode` means applying to all tables in this task, not only table-routes tables


pkg/terror/error_list.go
```diff
@@ -998,6 +999,7 @@ var (
 	ErrSyncerOperatorNotExist               = New(codeSyncerOperatorNotExist, ClassSyncUnit, ScopeInternal, LevelLow, "error operator not exist, position: %s", "")
 	ErrSyncerReplaceEventNotExist           = New(codeSyncerReplaceEventNotExist, ClassSyncUnit, ScopeInternal, LevelHigh, "replace event not exist, location: %s", "")
 	ErrSyncerParseDDL                       = New(codeSyncerParseDDL, ClassSyncUnit, ScopeInternal, LevelHigh, "parse DDL: %s", "Please confirm your DDL statement is correct and needed. For TiDB compatible DDL, see https://docs.pingcap.com/tidb/stable/mysql-compatibility#ddl. You can use `handle-error` command to skip or replace the DDL or add a binlog filter rule to ignore it if the DDL is not needed.")
+	ErrSyncerUnsupportedStmt                = New(codeSyncerUnsupportedStmt, ClassSyncUnit, ScopeInternal, LevelHigh, "`%s` statement not supported in %s mode", "")
```
lance6716, line 1002: we could add a workaround 🤔

lance6716:
need to tell user how to handle this situation. rest LGTM

pkg/terror/error_list.go
GMHDBJD: I can't find out a solution. Maybe we should support rename to the same group now. 😥

syncer/syncer.go
```diff
@@ -1837,6 +1837,10 @@ func (s *Syncer) handleQueryEvent(ev *replication.QueryEvent, ec eventContext, o
 			case *ast.TruncateTableStmt:
 				ec.tctx.L().Info("ignore truncate table statement in shard group", zap.String("event", "query"), zap.String("statement", sqlDDL))
 				continue
+			// TODO: Support rename a table to the same shard group
+			// e.g. upstream: tb* -> downstream: tb, rename tb1 to tb2
+			case *ast.RenameTableStmt:
```
lance6716, line 1842: will it error for the non-sharding tables?

GMHDBJD: Yes, it does happen with the above implementation. I'm working on “rename to the same group" and will try to fix it.

GMHDBJD: Actually, if mode is pessimistic, there isn't non-sharding tables.

lance6716: OK, that's related to https://github.com/pingcap/dm/issues/1517. In future we should support specify part of sources to form a sharding group.

syncer/syncer.go
```diff
@@ -1856,6 +1856,8 @@ func (s *Syncer) handleQueryEvent(ev *replication.QueryEvent, ec eventContext, o
 			case *ast.TruncateTableStmt:
 				ec.tctx.L().Info("ignore truncate table statement in shard group", zap.String("event", "query"), zap.String("statement", sqlDDL))
 				continue
+			case *ast.RenameTableStmt:
```
lance6716, line 1859: we heavily rely on "an optimistic sharding task should sync no other tables than `table-route`", if user both sync sharding tables and non sharding tables in one task, non sharding tables will not permitted to RENAME.

How about we give an error when user both specify block-allow list and table routes, to ensure user add a minimal sharding task that will not sync non-sharding tables? we might talk to PM if you think that's reasonable. @GMHDBJD @lichunzhu 

could join us after your vacation @gozssky

lance6716: (but https://github.com/pingcap/dm/issues/1517 is still not solved)

GMHDBJD: Seems there is not associated between #1517 and rename table? We will support rename in both pessimistic and optimistic later. It's no conflict that we can also support #1517 later.

lance6716: yes, what do you think "this PR also forbid user RENAME a non sharding table in a optimistic sharding task", could we raise an error when user both specify block-allow list and table routes (use table routes as allow list)?

GMHDBJD: We have other stmts not supported for a non sharding table in shard mode(truncate statement). So it's not a problem of Rename.
> raise an error when user both specify block-allow list and table routes 

We can support it in another pr or #1517? Maybe task to PM will be better.

PR #1516, opened 2021-03-18
chaos: support optimistic mode for chaos test

GMHDBJD:
run optimistic task in chaos test

- [x] don't execute unsupported DDL for optimistic
- [x] inject manually design test for optimstic
- [x] reload schema after execute DDL, otherwise dml will not include newly added column
- [x] use context.Background as db connection ctx, so we make sure DDL will be executed for all sources when ctx cancel.
- [x] persist log so log still exist after master/worker restart
- [x] reduce max backoff of auto resume, so if a `wasn't fully dropped in downstream` error occur due to the different sync time of DDLs in different sources, we can auto resume it and data consistent finally.
- [ ] ~~Disrupt the order of DDL execution~~ Do it later(in another pr)

Tests 

 - chaos test


chaos/cases/stmt.go
```diff
@@ -44,8 +46,53 @@ func randDML(ss *sqlsmith.SQLSmith) (dml string, t dmlType, err error) {
...
+		return false, nil
+	}
+
+	if len(v.Specs) == 0 || v.Specs[0].Tp != ast.AlterTableAddColumns || len(v.Specs[0].NewColumns) == 0 {
+		return false, nil
+	}
+
+	spec := v.Specs[0]
+	notNull := false
+	for _, newCol := range spec.NewColumns {
```
lance6716, line 86: will it generate multiple new columns?

chaos/cases/task.go
```diff
@@ -82,7 +85,7 @@ func newTask(ctx context.Context, cli pb.MasterClient, taskFile string, schema s
 		if err2 != nil {
 			return nil, err2
 		}
-		conn, err2 := createDBConn(ctx, db, schema)
+		conn, err2 := createDBConn(context.Background(), db, schema)
```
lance6716, line 88: why change context here

lance6716 (commented):
will review later

pkg/retry/errors.go
```diff
@@ -62,8 +62,8 @@ var (
 
 	// UnresumableErrCodes is a set of unresumeable err codes.
 	UnresumableErrCodes = map[int32]struct{}{
-		int32(terror.ErrSyncUnitDDLWrongSequence.Code()):    {},
-		int32(terror.ErrSyncerShardDDLConflict.Code()):      {},
+		int32(terror.ErrSyncUnitDDLWrongSequence.Code()): {},
+		//	int32(terror.ErrSyncerShardDDLConflict.Code()):      {},
```
lance6716, line 66: 🤔  is this a WIP?

chaos/cases/generator.go
```diff
@@ -0,0 +1,208 @@
...
+						fullSqls := make(SQLs, len(sqls))
+						copy(fullSqls, sqls)
+						for idx, sql := range fullSqls {
+							fullSqls[idx].statement = fmt.Sprintf(sql.statement, g.schema, table)
+						}
+						g.sqlsChan <- fullSqls
+					}
+				}
+			}
+			g.sqlsChan <- nil
```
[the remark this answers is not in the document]
GMHDBJD: If close it, we should start it for every increase round. Now we can only start the generator at the beginning of incrLoop.

chaos/cases/stmt.go
GMHDBJD: Not sure. But this range is safer?

chaos/cases/generator.go
```diff
@@ -0,0 +1,257 @@
...
+		testCases:   cases[shardMode],
+		sqlsChan:    make(chan SQLs),
+		testPreSQLs: preSQLs[shardMode],
+	}
+	return g
+}
+
+// Start starts to generate sqls case
+func (g *CaseGenerator) Start(ctx context.Context, schema string, tables []string) {
+	rand.Seed(time.Now().UTC().UnixNano())
```
lance6716, line 212: we use `rand.Seed(time.Now().UnixNano())` elsewhere 🤔

chaos/cases/task.go
GMHDBJD: reverted

chaos/cases/stmt.go
lance6716: In line 90 we don't check rest columns

chaos/cases/task.go
```diff
@@ -316,20 +339,53 @@ func (t *task) genIncrData(ctx context.Context) (err error) {
...
+	}
+
+	defer func() {
+		log.L().Info("complete test case sql")
+		for {
+			if !getNewCase {
+				return
+			}
+
+			if err2 := runCaseSQLs(); err2 != nil {
```
lance6716, line 364: we `runCaseSQLs` in below for loop. Here we want to make sure run it at least once?

lance6716 (commented):
rest lgtm

lance6716:
BTW, still WIP now?

chaos/cases/task.go
GMHDBJD: Make sure it runs at least once, and make sure all cases will be completed.

chaos/cases/generator.go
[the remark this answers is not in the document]
lance6716: we call `genSQLs` only in `Start`, so there's no "`Start` once and `genSQLs` multiple times to avoid re-open the channel" usage. OK to not change

chaos/cases/stmt.go
```diff
@@ -28,24 +30,75 @@ const (
 )
 
 // randDML generates DML (INSERT, UPDATE or DELETE).
-// NOTE: 3 DML types have the same weight now.
+// insert:update:delete=2:1:1
 func randDML(ss *sqlsmith.SQLSmith) (dml string, t dmlType, err error) {
-	t = dmlType(rand.Intn(3) + 1)
+	t = dmlType(rand.Intn(4) + 1)
```
lance6716, line 35: if we remove the +1 in `iota +1`, we could also remove it here 🤔

lance6716 (approved):

loader/loader.go
```diff
@@ -1254,14 +1254,14 @@ func (q *jobQueue) startConsumers(handler func(ctx context.Context, job *restore
 					// test condition for `job.session` means db session still could be controlled outside,
 					// it's used in unit test for now.
 					if session == nil && job.session == nil {
-						baseConn, err := job.loader.toDB.GetBaseConn(q.ctx)
-						if err != nil {
-							return err
+						baseConn, err2 := job.loader.toDB.GetBaseConn(q.ctx)
+						if err2 != nil {
```
gozssky, line 1258: `err` and `err2` may reduce a little readability. I think we'd better remove the unnecessary pre-declared err and return err explicitly to avoid accidental variable shadowing.

gozssky (commented):
(not important, just a code style discussion

lance6716:
> (not important, just a code style discussion

this is a LGTM 🤔

loader/loader.go
GMHDBJD: address in 0670fb4

gozssky:
LGTM

lance6716:
"require-LGT1" added. PR can trigger merge now

PR #1518, opened 2021-03-18
bug-fix: use table-info-before always and fix bug for recover lock in optimistic

GMHDBJD:
#1510 case4

- if table-info-before of an info is different with table saved in master(like worker restart and TrySync twice)
  - use table-info-before always
  - joined tables base on table-info-before as oldJoined
- when recovering lock
  - join all info
  - build lock with joined info, update tables with tableInfoBefore
  - sort the infos, TrySync them one by one

### Check List 

Tests 
 - unit test
 - integration test


lichunzhu:
Is this PR still WIP?

pkg/shardddl/optimism/lock.go
```diff
@@ -72,7 +72,7 @@ func NewLock(ID, task, downSchema, downTable string, ti *model.TableInfo, tts []
 		synced:     true,
 		versions:   make(map[string]map[string]map[string]int64),
 	}
-	l.addTables(tts)
+	l.addTables(l.joined, tts)
```
lichunzhu, line 75: What if `tts` contains the other tables whose schema didn't reach `l.joined`?

GMHDBJD: Here l.joined equals tableInfoBefore. If it's a new lock(first table), tableInfoBefore=l.joined. For other newer tables, they will be added in L163

lichunzhu: The `tts` here is from https://github.com/pingcap/dm/blob/master/dm/master/shardddl/optimist.go#L267. `optimistic` will get all source tables from etcd and this `tts` may have not only the table in this info, I think.

GMHDBJD: em... That's right, I will fix it

pkg/terror/error_list.go
```diff
@@ -1059,6 +1060,7 @@ var (
 	ErrMasterFailToImportFromV10x = New(codeMasterFailToImportFromV10x, ClassDMMaster, ScopeInternal, LevelHigh, "fail to import DM cluster from v1.0.x", "Please confirm that you have not violated any restrictions in the upgrade documentation.")
 
 	ErrMasterInconsistentOptimisticDDLsAndInfo = New(codeMasterInconsistentOptimistDDLsAndInfo, ClassDMMaster, ScopeInternal, LevelHigh, "inconsistent count of optimistic ddls and table infos, ddls: %d, table info: %d", "")
+	ErrMasterOptimisticTableInfoBeforeNotExist = New(codeMasterOptimisticTableInfobeforeNotExist, ClassDMMaster, ScopeInternal, LevelHigh, "table info before not exist in optimistic ddls: %v", "")
```
lance6716, line 1063:
```suggestion
	ErrMasterOptimisticTableInfoBeforeNotExist = New(codeMasterOptimisticTableInfobeforeNotExist, ClassDMMaster, ScopeInternal, LevelHigh, "table-info-before not exist in optimistic ddls: %v", "")
```

the original sentence is confusing

pkg/shardddl/optimism/info.go
```diff
@@ -54,6 +54,10 @@ type Info struct {
 	// only set it when get/watch from etcd
 	Version int64 `json:"-"`
 
+	// only set it when get from etcd
+	// use for sort infos in recoverlock
+	ReVision int64 `json:"-"`
```
lance6716, line 59:
```suggestion
	Revision int64 `json:"-"`
```

dm/master/shardddl/optimist_test.go
```diff
@@ -1029,3 +1030,70 @@ func (t *testOptimist) TestOptimistInitSchema(c *C) {
 	c.Assert(err, IsNil)
 	c.Assert(is.TableInfo, DeepEquals, ti1) // the init schema is ti1 now.
 }
+
+func (t *testOptimist) TestSortInfos(c *C) {
```
lance6716, line 1034:
```suggestion
func (t *testOptimist) testSortInfos(c *C) {
```

Or remove line 151 because a capitalized testing will be called by testing framework I think

pkg/shardddl/optimism/lock.go
```diff
@@ -149,13 +157,19 @@ func (l *Lock) TrySync(info Info, tts []TargetTable) (newDDLs []string, err erro
...
 	// is not in old source tables and current new source tables.
 	// duplicate append is not a problem.
 	tts = append(tts, newTargetTable(l.Task, callerSource, l.DownSchema, l.DownTable,
 		map[string]map[string]struct{}{callerSchema: {callerTable: struct{}{}}}))
 	// add any new source tables.
 	l.addTables(tts)
+	l.receiveTable(callerSource, callerSchema, callerTable, oldTable)
 	if val, ok := l.versions[callerSource][callerSchema][callerTable]; !ok || val < infoVersion {
 		l.versions[callerSource][callerSchema][callerTable] = infoVersion
 	}
```
lance6716, line 175: now we changed the `l.tables` by `l.receiveTable`, so in line 179 `oldJoined` is not a result of "join every l.tables" 🤔

GMHDBJD: Did you mean join all the tables for old joined?

lance6716: I'm just not sure if we depend on "old joined should be joined result of all l.tables", or "old joined of previous TrySync and new joined of next TrySync should keep consistent"

GMHDBJD: If we assume user start-task with same table info in all upstream, tableInfoBefore should be same as oldJoined. Otherwise, the oldjoined is not collect and we may keep it and hope later join may return an error? cc @lichunzhu

lichunzhu: If dm-master/dm-worker meets a restart the table info in all upstream is more likely to be different. What's the problem here? I don't understand this clearly.

GMHDBJD: I think the tableInfoBefore is same as l.tables[source][schema][table], otherwise it's a bug or idempotent retry.

GMHDBJD: But if the tableInfoBefore not equal l.tables[source][schema][table], that means the l.tables[source][schema][table] and oldjoined may wrong, do we need join all the tables with tableInfoBefore as oldjoined?

lichunzhu: > But if the tableInfoBefore not equal l.tables[source][schema][table], that means the l.tables[source][schema][table] and oldjoined may wrong, do we need join all the tables with tableInfoBefore as oldjoined?

I think carefully about this situation. Could you help me check if I'm wrong?

When this happens, it means we add a table whose table schema is not equal to joined table info now. When we init all the tables, we have three situations:
1. This task is new. The pre-check will assure the correctness of the tables' schemas.
2. This dm cluster recovers from a failure. The lock is synced now and all the tables have the same schema. **If we add a table with a different schema, we should report an error.**
3. This dm cluster recovers from a failure. The lock is unsynced now. Some workers didn't put their lock Info into etcd before the restart. These tables' `l.tables` will be set to `l.joined` at first, and if it's different from the `tableInfoBefore` received later I think it's acceptable. Am I right?

pkg/shardddl/optimism/info.go
```diff
@@ -54,6 +54,10 @@ type Info struct {
 	// only set it when get/watch from etcd
 	Version int64 `json:"-"`
 
+	// only set it when get from etcd
+	// use for sort infos in recoverlock
+	Revision int64 `json:"-"`
```
lichunzhu, line 59: I set this field's name to `ModRevision` in my PR. 🤣

pkg/shardddl/optimism/lock.go
GMHDBJD: So you mean if we get a info which we have already receive the table info before, if their schema is different, we should report an error?

lichunzhu: Yes. But I think we shouldn't report an error in situation 3.

GMHDBJD: > Some workers didn't put their lock Info into etcd before the restart. These tables' l.tables will be set to l.joined at first, and if it's different from the tableInfoBefore received later

That means the table haven't received. `l.received[x][x][x]=false`

In unit test, we have some idempotent TrySync, how to deal with them?

GMHDBJD: > Some workers didn't put their lock Info into etcd before the restart. These tables' l.tables will be set to l.joined at first

That means the table haven't been received. `l.received[x][x][x]==false`

> In unit test, we have some idempotent TrySync, how do we deal with the case? 🤔

tests/shardddl1/run.sh
```diff
@@ -589,20 +589,85 @@ function DM_RENAME_COLUMN_OPTIMISTIC() {
...
+#    end=35
+#    except=(024 025 029)
+#    for i in $(seq -f "%03g" ${start} ${end}); do
+#        if [[ ${except[@]} =~ $i ]]; then
+#            continue
+#        fi
+#        DM_${i}
+#        sleep 1
+#    done
+#    DM_RENAME_COLUMN_OPTIMISTIC
```
lichunzhu, line 669: for debug?

dm/master/shardddl/optimist.go
```diff
@@ -255,29 +257,83 @@ func (o *Optimist) rebuildLocks() (revSource, revInfo, revOperation int64, err e
...
+			for _, schemaInfos := range sourceInfos {
+				for _, info := range schemaInfos {
+					lockID := utils.GenDDLLockID(info.Task, info.DownSchema, info.DownTable)
+					if joined, ok := lockJoined[lockID]; !ok {
+						lockJoined[lockID] = schemacmp.Encode(info.TableInfoBefore)
+					} else {
+						newJoined, err := joined.Join(schemacmp.Encode(info.TableInfoBefore))
+						// ignore error, will report it in TrySync later
+						if err != nil {
+							o.logger.Error(fmt.Sprintf("fail to join table info %s with %s, lockID: %s in recover lock", joined, newJoined, lockID), log.ShortError(err))
```
lance6716, line 297: we could add a alerting rule for errors of Optimist initialization in future

dm/master/shardddl/optimist_test.go
```diff
@@ -1029,3 +1031,127 @@ func (t *testOptimist) TestOptimistInitSchema(c *C) {
...
+	rev3, err := optimism.PutInfo(etcdTestCli, i21)
+	c.Assert(err, IsNil)
+	rev4, err := optimism.PutInfo(etcdTestCli, i11)
+	c.Assert(err, IsNil)
+	ifm, _, err = optimism.GetAllInfo(etcdTestCli)
+	c.Assert(err, IsNil)
+	infos = sortInfos(ifm)
+	c.Assert(len(infos), Equals, 3)
+
+	i11.Version = 2
```
lance6716, line 1091: seems we could replace `Info.Version` by `Revision` (could file another PR)

pkg/shardddl/optimism/keeper_test.go
```diff
@@ -391,3 +393,63 @@ func (t *testKeeper) TestTargetTablesForTask(c *C) {
...
+		tblID      int64 = 111
+		ti0              = createTableInfo(c, p, se, tblID, `CREATE TABLE bar (id INT PRIMARY KEY)`)
+		ti1              = createTableInfo(c, p, se, tblID, `CREATE TABLE bar (id INT PRIMARY KEY, c1 INT)`)
+		ti2              = createTableInfo(c, p, se, tblID, `CREATE TABLE bar (id INT PRIMARY KEY, c1 INT, c2 INT)`)
+		ti3              = createTableInfo(c, p, se, tblID, `CREATE TABLE bar (id INT PRIMARY KEY, c2 INT)`)
+
+		i11 = NewInfo(task, source1, upSchema, upTable, downSchema, downTable, DDLs1, ti0, []*model.TableInfo{ti1})
+		i21 = NewInfo(task, source2, upSchema, upTable, downSchema, downTable, DDLs2, ti2, []*model.TableInfo{ti3})
+
+		tts = []TargetTable{
```
lance6716, line 420: "target" in `TargetTable` makes me think of "source/target", where the "target" means downstream tables. We could rename it to "TableMap" or something in future

pkg/shardddl/optimism/lock_test.go
```diff
@@ -1580,3 +1584,81 @@ func newInfoWithVersion(task, source, upSchema, upTable, downSchema, downTable s
 	info.Version = vers[source][upSchema][upTable]
 	return info
 }
+
+func (t *testLock) TestLockTrySyncReceived(c *C) {
```
lance6716, line 1588: which case does it test?

tests/shardddl1/run.sh
```diff
@@ -589,6 +589,70 @@ function DM_RENAME_COLUMN_OPTIMISTIC() {
...
+    # tb1(a,b) tb2(a,b)
+    run_sql_source1 "insert into ${shardddl1}.${tb1} values(1,1);"
+    run_sql_source2 "insert into ${shardddl1}.${tb2} values(2,2);"
+
+    # tb1(a,b,c); tb2(a,b)
+    run_sql_source1 "alter table ${shardddl1}.${tb1} add column c varchar(10);"
+    run_sql_source1 "insert into ${shardddl1}.${tb1} values(3,3,'aaa');"
+    check_log_contain_with_retry "putted a shard DDL.*tb1.*ALTER TABLE .* ADD COLUMN" $WORK_DIR/worker1/log/dm-worker.log $WORK_DIR/worker2/log/dm-worker.log 
+
+    # tb1(a,c); tb2(a)
```
lance6716, line 602:
```suggestion
    # tb1(a,b,c); tb2(a)
```

lance6716 (commented):
will review later

pkg/shardddl/optimism/lock_test.go
GMHDBJD: rename in 814d7a7 . Now we use table-info-before for all table, create a different index will not be in conflict now.

tests/shardddl1/run.sh
```diff
@@ -589,6 +589,70 @@ function DM_RENAME_COLUMN_OPTIMISTIC() {
...
+    run_sql_source2 "insert into ${shardddl1}.${tb2} values(5);"
+    check_log_contain_with_retry "putted a shard DDL.*tb1.*ALTER TABLE .* DROP COLUMN" $WORK_DIR/worker1/log/dm-worker.log $WORK_DIR/worker2/log/dm-worker.log 
+
+    # tb1(a,c); tb2(a,b)
+    run_sql_source2 "alter table ${shardddl1}.${tb2} add column b int;"
+    run_sql_source1 "insert into ${shardddl1}.${tb1} values(6,'ccc');"
+    run_sql_source2 "insert into ${shardddl1}.${tb2} values(7,7);"
+    check_log_contain_with_retry "putted a shard DDL.*tb2.*ALTER TABLE .* ADD COLUMN" $WORK_DIR/worker1/log/dm-worker.log $WORK_DIR/worker2/log/dm-worker.log 
+
+    # recover lock, tb1's info: (a,b,c)->(a,c); tb2's info: (a)->(a,b)
```
lichunzhu, line 624: Do we cover case#4 at this place?

lichunzhu (commented):
Rest LGTM

tests/shardddl1/run.sh
GMHDBJD: I think yes? Though the order cannot determine, but we don't depend on the order now

PR #1521, opened 2021-03-19
master: add etcd compaction and quota config

### LINKED ISSUE #1519: set default AutoCompactionMode
https://github.com/tikv/pd/blob/760fb650a1c12cc73ac48ac23c6563bc43eb5ada/server/config/config.go#L547-L548

like pd

lance6716:
part of #1519 

- [x] set default values and expose etcd config for auto-compaction and backend-quota
- [x] enable auto-compaction for a short value in integration test, to see if DM could handle compaction error

### Check List 

Tests 

 - Unit test
 - Integration test to check DM could handle compaction error

Code changes

 - Has persistent data change

Side effects

Related changes


lance6716:
I tested setting and not setting auto comapction configurations, with writing different value to a same key. The DBSize of endpoint status has no difference most time, while for one configuration old revision key is compacted and other is not.

Will check DBSize of endpoint status in integration test later.

lance6716:
https://github.com/etcd-io/etcd/issues/8496

according to above link, auto compaction does no help to disk problems, and defregment will stop etcd serving thus need carefully trigger, we might let user configurate a big quota to reduce the chance, and write an FAQ on etcd defregment or provide a helper command

lichunzhu:
/lgtm
But if we want to add this in v2.0.2, we'd better and periodically defragment too.

lance6716:
> But if we want to add this in v2.0.2, we'd better and periodically defragment too.

I think adding this and lacking defragment didn't make things worse. defragment will stop etcd's serving, thus we should let master retire leader to run a defragment. I'm not clear about the logic now

lichunzhu:
Users can't achieve their goals if they don't do defragment manually. They may think auto-compaction can avoid `Error: etcdserver: mvcc: database space exceeded` problem. But actually, it's insufficient.

lance6716:
> Users can't achieve their goals if they don't do defragment manually. They may think auto-compaction can avoid `Error: etcdserver: mvcc: database space exceeded` problem. But actually, it's insufficient.

we could tell them auto-compaction features are the same behaviour with etcd/pd, so their expectation is wrong.

lichunzhu:
> > Users can't achieve their goals if they don't do defragment manually. They may think auto-compaction can avoid `Error: etcdserver: mvcc: database space exceeded` problem. But actually, it's insufficient.
> 
> we could tell them auto-compaction features are the same behaviour with etcd/pd, so their expectation is wrong.

OK. We should add this to our documents. I'm curious about how pd handles these problems?

GMHDBJD (commented):
We may no need to config `auto-compaction-retentio` in so many integration test. I think most users will not manualy config it.
rest LGTM

lance6716:
> We may no need to config `auto-compaction-retention in so many integration tests. I think most users will not manually config it.
> rest LGTM

now we set the efault value to `1h` so DM may face etcd compacted error. I set it to a shorter value in integration test to expose failure of etcd compacted error

GMHDBJD (approved):
/lgtm

Issue #1524, opened 2021-03-23
relay didn't get latest "min location from subtasks", and when continue parsing heartbeat logic is wrong


lance6716:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do? If possible, provide a recipe for reproducing the error.

two problems from https://github.com/pingcap/dm/pull/1518/checks?check_run_id=2165916757

1.1 when resuming relay from detecting upstraem switching, relay didn't get the chance to run the code of "get min location from subtasks", so it will pull very old binlogs

1.2 when we continue parsing relay log files because of `needReParse`, we may continue from alocation between "GTIDEvent" and "XIDEvent", so `replaceWithHeartbeat` is wrongly calculated

2. What did you expect to see?

3. What did you see instead?

4. Versions of the cluster

    - DM version (run `dmctl -V` or `dm-worker -V` or `dm-master -V`):

        ```
        (paste DM version here, and you must ensure versions of dmctl, DM-worker and DM-master are same)
        ```

    - Upstream MySQL/MariaDB server version:

        ```
        (paste upstream MySQL/MariaDB server version here)
        ```

    - Downstream TiDB cluster version (execute `SELECT tidb_version();` in a MySQL client):

        ```
        (paste TiDB cluster version here)
        ```

    - How did you deploy DM: DM-Ansible or manually?

        ```
        (leave DM-Ansible or manually here)
        ```

    - Other interesting information (system version, hardware config, etc):

        >
        >

5. current status of DM cluster (execute `query-status` in dmctl)

6. Operation logs
   - Please upload `dm-worker.log` for every DM-worker instance if possible
   - Please upload `dm-master.log` if possible
   - Other interesting logs
   - Output of dmctl's commands with problems
   
7. Configuration of the cluster and the task
   - `dm-worker.toml` for every DM-worker instance if possible
   - `dm-master.toml` for DM-master if possible
   - task config, like `task.yaml` if possible
   - `inventory.ini` if deployed by DM-Ansible

8. Screenshot/exported-PDF of Grafana dashboard or metrics' graph in Prometheus for DM if possible

lance6716:
for 1.1, seems the failure is caused by multiple mysql instance hiding from a same IP/source-id, so there maybe mixed server_uuid/server-id for a relay unit. I'll check later

lance6716:
we can't start parsing a WriteRowsEvent without seeing its previous TableMapEvent @GMHDBJD 

```
[...]

PR #1525, opened 2021-03-23
streamer: fix duplicate event when reparse relay using GTID

lance6716:
part of https://github.com/pingcap/dm/issues/1524

when DM uses GTID and needs to return a reparse location, return last XID location so that `replaceWithHeartbeat` will be correctly calculated

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes


lichunzhu (approved):

GMHDBJD:
Will sync unit get duplicate events? (like gtid_event, table map event, rows event)

lance6716:
> Will sync unit get duplicate events? (like gtid_event, table map event, rows event)

Yes, currently we rely on safe mode. I added a todo of changing the behaviour of sending binlog to "sending DML after see XID",

GMHDBJD:
> Yes, currently we rely on safe mode. I added a todo of changing the behaviour of sending binlog to "sending DML after see XID". Do you think we should do it now?

I think we may simply return both currentGTID and latestPos for parseFile now. Otherwise, if safemode has been closed, we may have duplicate data.

GMHDBJD (approved):
/lgtm

PR #1538, opened 2021-03-28
test: align with TiDB clustered index behaviour

lance6716:
revert some of https://github.com/pingcap/dm/pull/1522

### Check List 

Tests 

 - No code

Related changes


GMHDBJD:
Should we enable `alter-primary-key` in schemaTracker?

lance6716:
> Should we enable `alter-primary-key` in schemaTracker?

I want to set OFF clustered index in another PR, so in future we don't need to change again when update TiDB

GMHDBJD:
> I want to set OFF clustered index in another PR, so in future we don't need to change again when update TiDB

Maybe we should set it the same as downstream?

lance6716:
> > I want to set OFF clustered index in another PR, so in future we don't need to change again when update TiDB
> 
> Maybe we should set it the same as downstream?

downstream may change global variable of clustered index and rebuild the table, at that time schema tracker has no way to know that change. I prefer keep schema tracker maximum compatibility of DDL.

GMHDBJD:
Will a create table stmt `create table tb(id int)` from upstream become `create table tb(id int nonclustered)` in tracker and exec to TiDB if we set `clustered index` OFF

PR #1540, opened 2021-03-29
common, upgrade: return `/` ternimated string when partitially encode keys

lance6716:
fix the bug that etcd watching of source-1 will receive changes of source-10

need manually run a Upgrade-via-TiUP github action to verify after merged

- as title
- add `/v2` to key prefix to upgrade some key adapter
- always use Hex encoder

### Check List 

Tests 

 - Unit test

Code changes

 - Has persistent data change

Side effects

Related changes


lance6716:
```
[2021/03/29 20:01:45.007 +08:00] [ERROR] [subtask.go:311] ["unit process error"] [subtask=sequence_sharding_optimistic] [unit=Sync] ["error information"="{\"ErrCode\":36062,\"ErrClass\":\"sync-unit\",\"ErrScope\":\"internal\",\"ErrLevel\":\"high\",\"Message\":\"startLocation: [position: (test-b4f9e55b-bc73-43b2-96a8-b867d271400f-s6n7n-hf5nb-bin.000001, 5845), gtid-set: ], endLocation: [position: (test-b4f9e55b-bc73-43b2-96a8-b867d271400f-s6n7n-hf5nb-bin.000001, 5983), gtid-set: ], origin SQL: [alter table t2 add column c1 varchar(20)]: fail to handle shard ddl [ALTER TABLE `sharding_target_opt`.`t_target` ADD COLUMN `c1` VARCHAR(20)] in optimistic mode, because schema conflict detected, conflict error: [code=11111:class=functional:scope=internal:level=medium], Message: fail to try sync the optimistic shard ddl lock sequence_sharding_optimistic-`sharding_target_opt`.`t_target`: add column c1 that wasn't fully dropped in downstream. ddl: ALTER TABLE `sharding_target_opt`.`t_target` ADD COLUMN `c1` VARCHAR(20), Workaround: Please use `show-ddl-locks` command for more details.\",\"Workaround\":\"Please use show-ddl-locks command for more details.\"}"]
```

lance6716:
which is caused by 

https://github.com/pingcap/dm/blob/2a4a99f9aa09eacb17780b40691a54c743498cbe/tests/sequence_sharding_optimistic/data/db1.increment.sql#L23-L30

reasonable 🤔

dm/common/common.go
```diff
@@ -173,3 +188,113 @@ func (s keyHexEncoderDecoder) Decode(key string) ([]string, error) {
...
 	return string(s)
 }
+
+// version V2 fixed the bug that XXDecoder.Encode("s1") is the prefix of XXDecoder.Encode("s10"), so watching prefix
+// gets unexpected KV
+// also, always use keyHexEncoderDecoderV2 to avoid `/` in keys
+type keyHexEncoderDecoderV2 string
+
+func quotes(s string) string {
+	return "`" + s + "`"
```
gozssky, line 198: Why don't we just add a separator to the end of the encoded key when watching prefix? It can simply avoid watching incorrect keys.

lance6716: do you mean s + `sep`? I just think quoting something is more familiar to others like an identifier in SQL, while familiar separators such as \0, ctrl+d, are not easy to type

gozssky: Yes. It seems all of the `KeyAdapter`s encode the custom keys to hex and joins them with '/'.  I'm not sure whether I have understood the scenario of watching prefix. If we want to watch the prefix `/dm/s1`, the expected keys should be `/dm/s1/xxx`? If so, adding '/' to the end of key looks more reasonable.

lance6716: Oh I see, will check it later

lance6716: 👍 I'm going to add an `EncodeAsPrefix`, which checks and returns a `/` terminated string

lance6716: still need to check watch `/dm/s1\0` could receive changes of `/dm/s1`, because there maybe `/dm/s10` 🤔

gozssky: does we actually want to watch `/dm/s1` instead of `/dm/s1/xxx`,  in this case,  just watch the key only.

lance6716:
(add "require-LGT1" because not recognized lgtm)

GMHDBJD (approved):
/lgtm

PR #1551, opened 2021-04-01
shardddl/optimistic: use init-schema for table without info when recover lock

gozssky:
In #1518, we always use table-info-before to rebuild locks, but ignore the upstream table who has no optimism.Info. In such a case, we may build an incorrect joined table.

Suppose there are two upstream tables tb1(a int, b int, c int) and tb2(a int, b int, c int).  A possible time series is:
```
t1: tb1 and tb2 are identical.
t2: tb1 drop column b, c.
t3: restart.
t4: joined becomes tb1's table-info-before, which is tb(a int). // The correct joined should be tb(a int, b int, c int).
```
Use init-schema for table without info when recover lock.  If there is no optimism.Info for a upstream table, it indicates the table structure hasn't been changed since last removeLock. So the init schema should be its table info.

### Check List 

Tests 

 - Unit test
 - Integration test

### DIFF
--- a/dm/master/shardddl/optimist_test.go
+++ b/dm/master/shardddl/optimist_test.go
@@ -1169,3 +1169,66 @@ func (t *testOptimist) TestBuildLockJoinedAndTable(c *C) {
 	c.Assert(err, IsNil)
 	c.Assert(cmp, Equals, 0)
 }
+
+func (t *testOptimist) TestBuildLockWithInitSchema(c *C) {
+	defer clearOptimistTestSourceInfoOperation(c)
+
+	var (
+		logger     = log.L()
+		o          = NewOptimist(&logger)
+		task       = "task"
+		source1    = "mysql-replica-1"
+		source2    = "mysql-replica-2"
+		downSchema = "db"
+		downTable  = "tbl"
+		st1        = optimism.NewSourceTables(task, source1)
+		st2        = optimism.NewSourceTables(task, source2)
+		p          = parser.New()
+		se         = mock.NewContext()
+		tblID      = int64(111)
+
+		ti0 = createTableInfo(c, p, se, tblID, `CREATE TABLE bar (a INT PRIMARY KEY, b INT, c INT)`)
+		ti1 = createTableInfo(c, p, se, tblID, `CREATE TABLE bar (a INT PRIMARY KEY, b INT)`)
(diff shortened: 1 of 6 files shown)


GMHDBJD (approved):
/lgtm

dm/master/shardddl/optimist_test.go
```diff
@@ -1165,3 +1169,66 @@ func (t *testOptimist) TestBuildLockJoinedAndTable(c *C) {
...
+		st2        = optimism.NewSourceTables(task, source2)
+		p          = parser.New()
+		se         = mock.NewContext()
+		tblID      = int64(111)
+
+		ti0 = createTableInfo(c, p, se, tblID, `CREATE TABLE bar (a INT PRIMARY KEY, b INT, c INT)`)
+		ti1 = createTableInfo(c, p, se, tblID, `CREATE TABLE bar (a INT PRIMARY KEY, b INT)`)
+		ti2 = createTableInfo(c, p, se, tblID, `CREATE TABLE bar (a INT PRIMARY KEY)`)
+
+		ddlDropB   = "ALTER TABLE bar dROP COLUMN b"
```
lance6716, line 1194:
```suggestion
		ddlDropB   = "ALTER TABLE bar DROP COLUMN b"
```

tests/shardddl3/run.sh
```diff
@@ -1020,20 +1020,24 @@ function DM_DropAddColumn_CASE() {
...
+    run_sql_source2 "insert into ${shardddl1}.${tb1} values(4,4);"
 
-    restart_master_on_pos $reset "3"
+    restart_master_on_pos $reset "2"
 
     # make sure column c is fully dropped in the downstream
     check_log_contain_with_retry 'finish to handle ddls in optimistic shard mode' $WORK_DIR/worker1/log/dm-worker.log
     check_log_contain_with_retry 'finish to handle ddls in optimistic shard mode' $WORK_DIR/worker2/log/dm-worker.log
 
+    run_dm_ctl_with_retry $WORK_DIR "127.0.0.1:$MASTER_PORT" \
```
lance6716, line 1032: could we remove above `check_log_contain_with_retry`?

lance6716 (commented):
rest lgtm

tests/shardddl3/run.sh
gozssky: seems no, `no DDL lock exists` doesn't guarantee that master received info sync `drop column c` to the downstream, probably master didn't received  any ddl infos.

PR #1559, opened 2021-04-06
shardddl: fix optimistic bug

### LINKED ISSUE #1554: optimistic bug found in chaos
## Bug Report

run optimistic chaos test
task succeeded
3. What did you see instead?
record below

GMHDBJD:
fix #1554 

- init miss info with init-schema
- detect SQL execute error before put info
- log short string for info
- delete drop columns after all source done
### Check List 

Tests 

 - Unit test
 - Integration test
 - chaos


GMHDBJD:
PTAL

lance6716 (commented):
will review later

syncer/optimist.go
```diff
@@ -91,6 +91,12 @@ func (s *Syncer) handleQueryEventOptimistic(
 		err      error
 	)
 
+	err = s.execError.Get()
+	if err != nil {
+		ec.tctx.L().Error("error detected when executing SQL job", log.ShortError(err))
+		return nil
```
lance6716, line 97: if we didn't return error, syncer's loop will continue to get the next event. Not sure things will more complex 🤔

tests/shardddl3/run.sh
```diff
@@ -2556,6 +2738,9 @@ function run() {
     DM_RestartMaster
 
     DM_DropAddColumn
+    DM_INIT_SCHEMA
```
lance6716, line 2741: now shardddl3 will take more than 20 minutes, and it's the slowest test. we might move it to another test.

lance6716: https://internal.pingcap.net/idc-jenkins/blue/organizations/jenkins/dm_ghpr_test/detail/dm_ghpr_test/8769/pipeline

so if CI get slower, we might can't merge all PR today 😂

pkg/shardddl/optimism/lock.go
```diff
@@ -28,6 +28,15 @@ import (
 	"github.com/pingcap/dm/pkg/terror"
 )
 
+const (
```
lance6716, line 31: maybe define a type for it, because when I see the value of map is `int` I assume that it means the count of not dropped columns

OK to not change if need change too many place

syncer/syncer.go
```diff
@@ -938,10 +939,18 @@ func (s *Syncer) syncDDL(tctx *tcontext.Context, queueBucket string, db *DBConn,
...
 			if len(sqlJob.ddls) == 0 {
 				ignore = true
-				tctx.L().Info("ignore shard DDLs in optimistic mode", log.WrapStringerField("info", s.optimist.PendingInfo()))
+				tctx.L().Info("ignore shard DDLs in optimistic mode", zap.Stringer("info", s.optimist.PendingInfo()))
 			}
 		}
-		if !ignore {
+
+		fp := false
+		failpoint.Inject("ExecDDLError", func() {
```
lance6716, line 947: or `failpoint.Goto("bypass")` + `failpoint.Label("bypass")`. OK not change

syncer/optimist.go
GMHDBJD: `sync` and `syncDDL` will cancel the loop. Same for [other mode](https://github.com/pingcap/dm/blob/ef6138bf373249f7f9afa8a6d112f555f59e5bc4/syncer/syncer.go#L1940).

GMHDBJD:
ha_master failed, already recorded
/run-all-tests

pkg/shardddl/optimism/info.go
```diff
@@ -84,6 +102,40 @@ func (i Info) String() string {
 	return s
 }
 
+// ShortString returns short string of Info
+func (i *Info) ShortString() string {
```
lance6716, line 106: I mean we could use value receiver, to avoid clone it in this function.

https://play.golang.org/p/qBG5ZRSFK0y

pkg/shardddl/optimism/keeper.go
```diff
@@ -66,10 +68,21 @@ func (lk *LockKeeper) RebuildLocksAndTables(
 			}
 		}
 	}
+
+	// update missTable's table info for locks
```
lance6716, line 72: I think we could remove `missTable`, and pass `initSchemas` here to do same logic. OK to not change because time maybe not enough

pkg/shardddl/optimism/lock.go
```diff
@@ -263,11 +278,23 @@ func (l *Lock) TrySync(info Info, tts []TargetTable) (newDDLs []string, err erro
...
+					return ddls, cols, err2
+				} else if len(col) > 0 && l.IsDroppedColumn(info.Source, info.UpSchema, info.UpTable, col) {
+					return ddls, cols, terror.ErrShardDDLOptimismTrySyncFail.Generate(
+						l.ID, fmt.Sprintf("add column %s that wasn't fully dropped in downstream. ddl: %s", col, ddls[idx]))
+				}
+			} else {
+				if col, err2 := GetColumnName(l.ID, ddls[idx], ast.AlterTableDropColumn); err2 != nil {
+					return ddls, cols, err2
+				} else if len(col) > 0 {
+					err = l.AddDroppedColumn(info, col)
```
lance6716, line 293: from the aspect of "cols is a cache of etcd", i think we should return when error or don't append to cols

GMHDBJD:
testSubTask.TestPauseAndResumeSubtask already recorded.
/run-all-tests

pkg/shardddl/optimism/info.go
GMHDBJD: But the struct between Info and LogInfo is different, unless we define another marshall method for info.

pkg/shardddl/optimism/lock.go
```diff
@@ -559,52 +586,92 @@ func (l *Lock) IsDroppedColumn(info Info, col string) bool {
...
 	return nil
 }
 
-// DeleteColumnsByDDLs deletes the partially dropped columns that extracted from DDLs.
-// We can not remove columns from the partially dropped columns map unless this column is dropped in the downstream database,
-// that is to say, op.Done is true and ddls contains drop column DDL.
-func (l *Lock) DeleteColumnsByDDLs(ddls []string) error {
+// DeleteColumnsByOp deletes the partially dropped columns that extracted from operation.
+// We can not remove columns from the partially dropped columns map unless:
+// this column is dropped in the downstream database,
```
lance6716, line 617: is this needed? the op with DDL also has a `done` upstream shard table

lance6716 (approved):

lance6716:
https://github.com/pingcap/dm/blob/6791a754617dac4d9487f94c088175263d64af5c/chaos/scripts/check-case.sh#L4

https://github.com/pingcap/dm/blob/6791a754617dac4d9487f94c088175263d64af5c/.github/workflows/chaos-mesh.yml#L255

lance6716:
/run-all-tests

unit test of etcd failed

lance6716:
(add require-LGT1 because reviewer is not enough)

GMHDBJD:
DM-147
/run-all-tests

GMHDBJD:
DM-147
/run-all-tests

pkg/shardddl/optimism/column.go
```diff
@@ -22,8 +24,9 @@ import (
 
 // GetAllDroppedColumns gets the all partially dropped columns.
 // return lockID -> column-name -> source-id -> upstream-schema-name -> upstream-table-name
```
lichunzhu, line 26: Please note what does `DropColumnStage` mean in comments.

dm/common/common.go
```diff
@@ -80,7 +80,7 @@ var (
 	// k/v: Encode(task-name, downstream-schema-name, downstream-table-name) -> table schema.
 	ShardDDLOptimismInitSchemaKeyAdapter KeyAdapter = keyHexEncoderDecoder("/dm-master/shardddl-optimism/init-schema/")
 	// ShardDDLOptimismDroppedColumnsKeyAdapter is used to store the columns that are not fully dropped
-	// k/v: Encode(task-name, downstream-schema-name, downstream-table-name, column-name, source-id, upstream-schema-name, upstream-table-name) -> empty
```
lichunzhu, line 83: Maybe we should write what's this `integer`'s usage?

lichunzhu: Besides, if you want to change this key's format, this part must be revised, too.
https://github.com/pingcap/dm/blob/master/pkg/shardddl/optimism/ops.go#L80

lichunzhu (commented):
I will take a further review later.

pkg/shardddl/optimism/column.go
GMHDBJD: comments in https://github.com/pingcap/dm/blob/5ae19df6c531537f61d8b961bbbf526fe09fab80/pkg/shardddl/optimism/lock.go#L31-L40

dm/common/common.go
GMHDBJD: > Maybe we should write what's this `integer`'s usage?

ditto https://github.com/pingcap/dm/blob/5ae19df6c531537f61d8b961bbbf526fe09fab80/pkg/shardddl/optimism/lock.go#L31-L40

GMHDBJD: > Besides, if you want to change this key's format, this part must be revised, too.
> https://github.com/pingcap/dm/blob/master/pkg/shardddl/optimism/ops.go#L80

address in bf05a9d

lichunzhu:
please fix the lint

PR #1412, opened 2021-01-31
Quality reinforcement (9): fix unstable test cases

lance6716:
- [x] testElectionSuite.TestElectionEvictLeader [jenkins](https://internal.pingcap.net/idc-jenkins/blue/organizations/jenkins/dm_ghpr_test/detail/dm_ghpr_test/7989/pipeline/72) https://github.com/pingcap/dm/pull/1412/commits/4d670d9b18f5223e1cbf2a7c26b1b7be2cf33559
  seems "some unit test using etcd [jenkins](https://internal.pingcap.net/idc-jenkins/blue/organizations/jenkins/dm_ghpr_test/detail/dm_ghpr_test/7989/pipeline/72)" is failed because etcd too slow

  ```
  [2021-01-11T08:54:57.388Z] [2021/01/11 16:53:08.828 +08:00] [WARN] [util.go:163] ["apply request took too long"] [took=1.630559395s] [expected-duration=100ms] [prefix=] [request="header:<ID:14726057132519571993 > txn:<compare:<target:CREATE key:\"unit-test/election-evict-leader/4c5d76f0a5efc205\" create_revision:0 > success:<request_put:<key:\"unit-test/election-evict-leader/4c5d76f0a5efc205\" value_size:40 lease:5502685095664796165 >> failure:<request_range:<key:\"unit-test/election-evict-leader/4c5d76f0a5efc205\" > >>"] [response=size:14] []
  ...
  [2021-01-11T08:54:57.388Z] [2021/01/11 16:53:08.830 +08:00] [INFO] [election.go:292] ["get response from election observe"] [component=election] [key=unit-test/election-evict-leader/4c5d76f0a5efc205] [value="{\"id\":\"member1\",\"addr\":\"127.0.0.1:1234\"}"]
  ```

  putting election key spends more than 1 second, so in the test sleeping 1 second and checking the leader is not enough
- [ ] testSubTask.TestPauseAndResumeSubtask subtask_test.go:365 ~[jenkins](https://internal.pingcap.net/idc-jenkins/blue/organizations/jenkins/dm_ghpr_test/detail/dm_ghpr_test/8133/pipeline/)~
   maybe ther's a little chance that the checker just peek when after `setResult` and before `setStage`
- [ ] testSubTask.TestPauseAndResumeSubtask subtask_test.go:394: ~[jenkins](https://internal.pingcap.net/idc-jenkins/blue/organizations/jenkins/dm_ghpr_test/detail/dm_ghpr_test/8205/pipeline)~
[...]

PR #1604, opened 2021-04-16
test: fix unstable shardddl test case DM-143

gozssky:
Fix unstable shardddl test case DM-143. See #1412.

schemacmp doesn't really support comparing partition before, which causes this unstable test.
Partition comparing will be implemented in tidb-tools.
Do not merge this PR before [tidb-tools#431](https://github.com/pingcap/tidb-tools/pull/431).

### Check List 

Tests 

 - Integration test

### DIFF
--- a/checker/checker.go
+++ b/checker/checker.go
@@ -224,7 +224,7 @@ func (c *Checker) Init(ctx context.Context) (err error) {
 				continue
 			}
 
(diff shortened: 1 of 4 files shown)


lance6716:
we might comment the optimistic test for DM-143 before finding a proper solution now, to improve CI stability

lance6716:
replaced by https://github.com/pingcap/dm/pull/1616. Thanks

Issue #1543, opened 2021-03-29
can't remove ddl-lock after stop-task


lance6716:
## Feature Request

**Is your feature request related to a problem? Please describe:**

some users report the error of `task xx which extracted from lock-ID is not found in DM`, at that situation, there's no way to remove a ddl lock

**Describe the feature you'd like:**

check if we could remove ddl-lock when stop-task

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

PR #1605, opened 2021-04-16
master: support unlock-ddl-lock after stop-task

lance6716:
close https://github.com/pingcap/dm/issues/1543

skip this check

### Check List 

Tests 

 - Integration test

Related changes


lichunzhu (commented):
/lgtm

GMHDBJD:
Can we remove lock after stop-task?

lance6716:
> Can we remove lock after stop-task?

since we didn't remove the checkpoint, I think we'd better also not removing ddl-lock in case of user `start-task` later

GMHDBJD (approved):
/lgtm

PR #1610, opened 2021-04-20
check: add error/warn count for check-task

### LINKED ISSUE #1478: grpc has a message limit, so long error message can't be received by dmctl
## Feature Request

some grpc requests may reach the limit, so we might backoff some options like https://pkg.go.dev/google.golang.org/grpc#MaxCallRecvMsgSize, or warn the user to split the input (most of time it's a big `task.yaml`)

GMHDBJD:
close #1478 

add error/warn count for check-task, default is 10

### Check List 

Tests 

 - Integration test


lance6716 (approved):

lichunzhu (commented):
Rest LGTM

dm/ctl/master/check_task.go
```diff
@@ -28,10 +28,12 @@ import (
 // NewCheckTaskCmd creates a CheckTask command.
 func NewCheckTaskCmd() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "check-task <config-file>",
+		Use:   "check-task <config-file> [--error] [--warn]",
```
lance6716, line 31: How about 

`[--error count]`

tests/dmctl_basic/check_list/check_task.sh
```diff
@@ -34,3 +34,33 @@ function check_task_error_database_config() {
...
+    # 1 error
+    run_dm_ctl $WORK_DIR "127.0.0.1:$MASTER_PORT" \
+        "check-task $cur/conf/dm-task3.yaml -e 1" \
+        "\"result\": false" 1 \
+        "\"failed\": 2" 1 \
+        "\"state\": \"fail\"" 1
+
+    # 100 errors
+    run_dm_ctl $WORK_DIR "127.0.0.1:$MASTER_PORT" \
+        "check-task $cur/conf/dm-task3.yaml -e 100 -w 1" \
```
lichunzhu, line 55: Use `--error` and `--warn` here?
```suggestion
        "check-task $cur/conf/dm-task3.yaml --error 100 --warn 1" \
```

lichunzhu (commented):
Rest LGTM

checker/checker.go
```diff
@@ -261,9 +265,21 @@ func (c *Checker) Process(ctx context.Context, pr chan pb.ProcessResult) {
...
 		// remove success result if not pass
 		results := result.Results[:0]
+		var warnCnt int64 = 0
+		var errCnt int64 = 0
 		for _, r := range result.Results {
-			if r.State != check.StateSuccess {
-				results = append(results, r)
+			switch r.State {
+			case check.StateFailure:
+				if c.errCnt >= 0 && errCnt < c.errCnt {
```
glorv, line 273:
```suggestion
				if errCnt < c.errCnt {
```

checker/checker.go
```diff
@@ -261,9 +265,21 @@ func (c *Checker) Process(ctx context.Context, pr chan pb.ProcessResult) {
...
+					errCnt++
+					results = append(results, r)
+				}
+			case check.StateWarning:
+				if c.warnCnt >= 0 && warnCnt < c.warnCnt {
```
glorv, line 278:
```suggestion
				if warnCnt < c.warnCnt {
```

checker/checker.go
```diff
@@ -75,14 +75,18 @@ type Checker struct {
 		sync.RWMutex
 		detail *check.Results
 	}
+	errCnt  int64
```
glorv, line 78: Is it make sense to restrict the total length of the errors or the total numbers of error + warn instead of set two different count separately. e.g. if we think be default we only return 20 messages, then if errors + warns <= 20, we needn't truncate any messages.

lance6716:
we now disable it by passing a big value instead of -1 🤔

checker/checker.go
GMHDBJD: I think error has a higher priority than warning, so if user gets 20 warnings and 1 errors we should make sure we will display that errors.

GMHDBJD:
> we now disable it by passing a big value instead of -1 🤔

cobra doesn't support negative argument... I keep the negative for other callings (like precheck of start-task). But now we use 10 as the default in every call, so I remove it.

lance6716:
wait @glorv to cancel the hold

checker/checker.go
lance6716: we could loop the results twice to make sure errors have higher priority 🤔 I'm OK with current two settings, let's wait glorv

lance6716:
DM-143 is failed because we didn't support optimistic sharding with PARTITION
/run-all-tests

checker/checker.go
glorv: I'm ok, too. I just wonder if this implementation is user-friendly enough to use since the goal is to solve the message too long issue.

BTW, I think maybe there are some cases the user just wants to emit the warning message, thus the `--warn` parameter can be used.

PR #1615, opened 2021-04-20
*: only use go.uber.org/atomic as atomic dependency

lance6716:

remove `github.com/siddontang/go`

### Check List 

Tests 

 - No code

Related changes


loader/loader.go
```diff
@@ -125,7 +124,7 @@ func (w *Worker) Close() {
 		failpoint.Return()
 	})
 
-	if !atomic.CompareAndSwapInt64(&w.closed, 0, 1) {
+	if w.closed.CAS(false, true) {
```
lance6716, line 127:
```suggestion
	if !w.closed.CAS(false, true) {
```

lance6716:
ptal @glorv @Ehco1996

dm/master/workerrpc/rawgrpc.go
```diff
@@ -31,15 +31,15 @@ import (
 type GRPCClient struct {
 	conn   *grpc.ClientConn
 	client pb.WorkerClient
-	closed int32
+	closed *atomic.Bool
```
glorv, line 34:
```suggestion
	closed atomic.Bool
```

Do not use pointer to be consistance with other declares.

pkg/utils/file.go
```diff
@@ -64,3 +67,24 @@ func GetFileSize(file string) (int64, error) {
...
+	}
+	n, err := f.Write(data)
+	f.Close()
+	if err == nil && n < len(data) {
+		err = io.ErrShortWrite
+	} else {
+		err = os.Chmod(f.Name(), perm)
+	}
+	if err != nil {
+		os.Remove(f.Name())
```
glorv, line 86: what if `os.Remove` call fails here, do we need to output a log?

glorv (commented):
rest LGTM

pkg/utils/file.go
lance6716: https://github.com/pingcap/dm/pull/1615/commits/feab697c8a39b13e80ca8eaac0a0b3c8265105a7

PR #1619, opened 2021-04-21
*: only get master status once when query-status

lance6716:
originally we get master status in every sync and relay unit, which is redundant because a worker handles only one source

get master status at worker level

### Check List 

Tests 

 - Pass original tests

Code changes

Related changes


dm/worker/server.go
```diff
@@ -765,11 +765,13 @@ func (s *Server) QueryStatus(ctx context.Context, req *pb.QueryStatusRequest) (*
...
 	}
 
-	resp.SubTaskStatus, sourceStatus.RelayStatus = w.QueryStatus(ctx, req.Name)
-	unifyMasterBinlogPos(resp, w.cfg.EnableGTID)
+	var err error
+	resp.SubTaskStatus, sourceStatus.RelayStatus, err = w.QueryStatus(ctx, req.Name)
 
 	if len(resp.SubTaskStatus) == 0 {
 		resp.Msg = "no sub task started"
+	} else if err != nil {
```
glorv, line 773: Should check `err != nil` first

lichunzhu (commented):
/lgtm

PR #1624, opened 2021-04-22
*: limit checker error at a finer granularity and auto increase max size

lance6716:
close https://github.com/pingcap/dm/issues/1478

- when RPC returned `grpc: received message larger than max (5242885 vs. 4194304)`, increase max message size and retry
- update dependency to limit the warn/error count at a finer granularity, and pretty print precheck errors

### Check List 

Tests 

 - Unit test

Related changes


checker/checker.go
```diff
@@ -262,24 +262,55 @@ func (c *Checker) Process(ctx context.Context, pr chan pb.ProcessResult) {
...
-					errCnt++
+			if r.State == check.StateSuccess {
+				continue
+			}
+
+			// handle results without r.Errors
+			if len(r.Errors) == 0 {
+				switch r.State {
+				case check.StateWarning:
+					if warnLeft == 0 {
```
lichunzhu, line 278: To handle `c.warnCnt == 0`
```suggestion
					if warnLeft <= 0 {
```

dm/ctl/common/util.go
```diff
@@ -19,11 +19,17 @@ import (
...
 	"time"
 
+	"github.com/pingcap/failpoint"
 	"go.etcd.io/etcd/clientv3"
+	"go.uber.org/zap"
+
+	"github.com/pingcap/dm/pkg/log"
 
 	"github.com/pingcap/dm/dm/config"
 	"github.com/pingcap/dm/dm/pb"
```
lichunzhu, line 35: fmt

dm/ctl/common/util.go
```diff
@@ -100,19 +117,38 @@ func (c *CtlClient) sendRequest(ctx context.Context, reqName string, req interfa
...
+	err := GlobalCtlClient.sendRequest(ctx, reqName, req, respPointer)
+	if err == nil {
+		return nil
+	}
+	var opts []interface{}
+	switch status.Code(err) {
+	case codes.ResourceExhausted:
+		matches := re.FindStringSubmatch(err.Error())
+		if len(matches) == 3 {
+			msgSize, err2 := strconv.Atoi(matches[1])
```
lichunzhu, line 129: Should we increase msgSize a little more?

checker/checker.go
lance6716: we can handle `warnLeft == 0` in this implementation 🤔

dm/ctl/common/util.go
lance6716: unit test shows this is enough

dm/ctl/common/util.go
lance6716: I think this is caued by our `make fmt` @Ehco1996 PTAL

checker/checker.go
lichunzhu: But we didn't check this now...

dm/ctl/common/util.go
lichunzhu: Will the result become a little bigger for the second request?

lance6716: we can't decide a reasonable delta 🤔

dm/ctl/common/util.go
Ehco1996: our fmt tools just can't group this imports correctly... but you can formart manually  - -.

dm/ctl/common/util.go
lichunzhu: OK, I'm just afraid it's too strict sometimes.

dm/ctl/common/util.go
lichunzhu: Can you fix these imports manually?

lichunzhu (approved):
LGTM

dm/ctl/common/util.go
glorv: Is there a hard limit of gRPC message size? And how about use next_pow_of_2(msg_size) as the new limit

lance6716: > Is there a hard limit of gRPC message size?

MaxInt. This only has memory affects. next_pow_of_2 seems more like a strategy for container capacity growing, while here I think using exactly the size is OK, because we sent the RPC again in a short time so the response is more likely not changed.

dm/master/server_test.go
```diff
@@ -1738,6 +1741,34 @@ func (t *testMaster) subTaskStageMatch(c *check.C, s *scheduler.Scheduler, task,
...
+	masterAddr := tempurl.Alloc()[len("http://"):]
+	lis, err := net.Listen("tcp", masterAddr)
+	c.Assert(err, check.IsNil)
+	defer lis.Close()
+	server := grpc.NewServer()
+	pb.RegisterMasterServer(server, &Server{})
+	//nolint:errcheck
+	go server.Serve(lis)
+
+	conn, err := grpc.Dial(utils.UnwrapScheme(masterAddr),
```
glorv, line 1761: should add `defer conn.Close()` here? 🤔

Issue #1598, opened 2021-04-15
bug found in chaos


GMHDBJD:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do? If possible, provide a recipe for reproducing the error.
record failed chaos bug.

lance6716:
DM-136, [jenkins](https://internal.pingcap.net/idc-jenkins/blue/organizations/jenkins/dm_ghpr_test/detail/dm_ghpr_test/8788/pipeline/) for DDL "drop index uk, add index uk" 

every worker should DROP the index, and the last worker should ADD the index. But worker1 runs slowly so it DROP after worker2 ADD the new index with same name

GMHDBJD:
Pessimistic
all source exeute ddl1; ddl2;
```
source 1 put ddl1's info
source 2, source 3 put ddl1's info
source1 execute ddl1
source1 put ddl1's done
source2, source3 skip ddl1
source1 crash, haven't flush checkpoint
source1 restart, put ddl1's info
source2, source3 put ddl2's info
lock couldn't  be resolved.
```
We don't support [this case](https://github.com/pingcap/docs-dm/blob/master/en/manually-handling-sharding-ddl-locks.md#scenario-2-some-dm-workers-stop-abnormally-or-the-network-failure-occurs-during-the-ddl-unlocking-process). But it will let chaos failed.

GMHDBJD:
[action](https://github.com/pingcap/dm/runs/2386477745)
```
query-status task_pessimistic -s replica-02
{
    "result": true,
    "msg": "",
    "sources": [
        {
            "result": true,
            "msg": "",
            "sourceStatus": {
                "source": "replica-02",
                "worker": "dm-worker-2",
                "result": null,
                "relayStatus": null
            },
            "subTaskStatus": [
                {
                    "name": "task_pessimistic",
                    "stage": "Paused",
                    "unit": "Load",
                    "result": {
                        "isCanceled": false,
                        "errors": [
                            {
                                "ErrCode": 10006,
                                "ErrClass": "database",
                                "ErrScope": "downstream",
                                "ErrLevel": "high",
[...]

PR #1626, opened 2021-04-23
pessimistic: record latest done ddl to handle conflict

GMHDBJD:
[issue](https://github.com/pingcap/dm/issues/1598#issuecomment-820289755)

when latest DDLs is resolved, put it into etcd.
when conflict DDLs detected in master, ignore DDLs which equals to the latest DDLs.

### Check List 

Tests 

 - Unit test
 - Integration test

### DIFF
--- a/dm/common/common.go
+++ b/dm/common/common.go
@@ -64,6 +64,9 @@ var (
 	// ShardDDLPessimismOperationKeyAdapter is used to store shard DDL operation in pessimistic model.
 	// k/v: Encode(task-name, source-id) -> shard DDL operation.
 	ShardDDLPessimismOperationKeyAdapter KeyAdapter = keyHexEncoderDecoder("/dm-master/shardddl-pessimism/operation/")
+	// ShardDDLPessimismDDLsKeyAdapter is used to store last done DDLs in pessimistic model.
+	// k/v: Encode(lockID) -> DDLs.
+	ShardDDLPessimismDDLsKeyAdapter KeyAdapter = keyHexEncoderDecoder("/dm-master/shardddl-pessimism/ddls/")
 
 	// ShardDDLOptimismSourceTablesKeyAdapter is used to store INITIAL upstream schema & table names when starting the subtask.
 	// In other words, if any Info for this subtask exists, we should obey source tables in the Info.
@@ -94,7 +97,7 @@ func keyAdapterKeysLen(s KeyAdapter) int {
 	switch s {
 	case WorkerRegisterKeyAdapter, UpstreamConfigKeyAdapter, UpstreamBoundWorkerKeyAdapter,
 		WorkerKeepAliveKeyAdapter, StageRelayKeyAdapter, TaskConfigKeyAdapter,
-		UpstreamLastBoundWorkerKeyAdapter, UpstreamRelayWorkerKeyAdapter:
+		UpstreamLastBoundWorkerKeyAdapter, UpstreamRelayWorkerKeyAdapter, ShardDDLPessimismDDLsKeyAdapter:
 		return 1
 	case UpstreamSubTaskKeyAdapter, StageSubTaskKeyAdapter,
 		ShardDDLPessimismInfoKeyAdapter, ShardDDLPessimismOperationKeyAdapter,
--- a/dm/master/shardddl/pessimist.go
+++ b/dm/master/shardddl/pessimist.go
@@ -147,8 +147,13 @@ func (p *Pessimist) buildLocks(etcdCli *clientv3.Client) (int64, int64, error) {
 	}
 	p.logger.Info("get history shard DDL lock operation", zap.Reflect("operation", opm), zap.Int64("revision", rev2))
 
+	latestDoneDDLsMap, _, err := pessimism.GetAllLatestDoneDDLs(etcdCli)
+	if err != nil {
+		return 0, 0, err
+	}
+
 	// recover the shard DDL lock based on history shard DDL info & lock operation.
-	err = p.recoverLocks(ifm, opm)
+	err = p.recoverLocks(ifm, opm, latestDoneDDLsMap)
 	if err != nil {
 		// only log the error, and don't return it to forbid the startup of the DM-master leader.
 		// then these unexpected locks can be handled by the user.
@@ -394,24 +399,30 @@ func (p *Pessimist) RemoveMetaData(task string) error {
 	for _, info := range infos {
 		p.lk.RemoveLockByInfo(info)
 	}
+	lockIDSet := make(map[string]struct{}, len(ops))
 	for _, op := range ops {
 		p.lk.RemoveLock(op.ID)
+		lockIDSet[op.ID] = struct{}{}
 	}
 
 	// clear meta data in etcd
-	_, err = pessimism.DeleteInfosOperationsByTask(p.cli, task)
+	_, err = pessimism.DeleteInfosOperationsDDLsByTask(p.cli, task, lockIDSet)
 	return err
 }
 
 // recoverLocks recovers shard DDL locks based on shard DDL info and shard DDL lock operation.
-func (p *Pessimist) recoverLocks(ifm map[string]map[string]pessimism.Info, opm map[string]map[string]pessimism.Operation) error {
+func (p *Pessimist) recoverLocks(ifm map[string]map[string]pessimism.Info,
+	opm map[string]map[string]pessimism.Operation, latestDoneDDLsMap map[string][]string) error {
+	// add all last done ddls.
+	p.lk.AddAllLatestDoneDDLs(latestDoneDDLsMap)
+
 	// construct locks based on the shard DDL info.
 	for task, ifs := range ifm {
 		sources := p.taskSources(task)
 		// if no operation exists for the lock, we let the smallest (lexicographical order) source as the owner of the lock.
 		// if any operation exists for the lock, we let the source with `exec=true` as the owner of the lock (the logic is below).
 		for _, info := range pessimismInfoMapToSlice(ifs) {
-			_, _, _, err := p.lk.TrySync(info, sources)
+			_, _, _, err := p.lk.TrySync(p.cli, info, sources)
 			if err != nil {
 				return err
 			}
@@ -471,7 +482,7 @@ func (p *Pessimist) handleInfoPut(ctx context.Context, infoCh <-chan pessimism.I
 			p.logger.Info("receive a shard DDL info", zap.Stringer("info", info))
 
 			p.infoOpMu.Lock()
-			lockID, synced, remain, err := p.lk.TrySync(info, p.taskSources(info.Task))
+			lockID, synced, remain, err := p.lk.TrySync(p.cli, info, p.taskSources(info.Task))
 			if err != nil {
 				p.logger.Error("fail to try sync shard DDL lock", zap.Stringer("info", info), log.ShortError(err))
 				// currently, only DDL mismatch will cause error
@@ -536,6 +547,9 @@ func (p *Pessimist) handleOperationPut(ctx context.Context, opCh <-chan pessimis
 					metrics.ReportDDLError(op.Task, metrics.OpErrRemoveLock)
 				}
 				p.logger.Info("the lock info for the shard DDL lock operation has been cleared", zap.Stringer("operation", op))
+				if err := p.addLatestDoneDDLs(lock.ID, lock.DDLs); err != nil {
+					p.logger.Info("fail to record last done ddls", zap.String("lock", lock.ID), log.ShortError(err))
+				}
 				p.infoOpMu.Unlock()
 				continue
 			}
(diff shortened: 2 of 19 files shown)


lance6716:
could we let DM-worker put `ShardDDLPessimismDDLsKeyAdapter`, to prevent DM-master down between "worker executed DDL" and "receive the done lock through etcd"?
And putting `ShardDDLPessimismDDLsKeyAdapter` should be in the same transaction with marking the lock done.

GMHDBJD:
> could we let DM-worker put `ShardDDLPessimismDDLsKeyAdapter`, to handle DM-master down between "receive the done lock through etcd" and "put `ShardDDLPessimismDDLsKeyAdapter`"?

So you mean we should record last done ddl for every worker? Currently only put last done ddl after lock is resolved(all source done). [Here](https://github.com/pingcap/dm/blob/fc5f8984a3b7b3dbaf0410b3d3b476074d8ad382/dm/master/shardddl/pessimist.go#L597) handle the situation about DM-master down.

> And I don't know if there're some DDL that could be run twice with some DML interleaved and take some effects

I don't think we will support truncate in pessimistic.🤣. We only skip the DDL when conflict error happened. So the case only happened like
```
source1 ddl1
source2 ddl1
done
source1 ddl1 <- skip
source2 ddl2
```
we handle this situation.

dm/common/common.go
```diff
@@ -64,6 +64,9 @@ var (
 	// ShardDDLPessimismOperationKeyAdapter is used to store shard DDL operation in pessimistic model.
 	// k/v: Encode(task-name, source-id) -> shard DDL operation.
 	ShardDDLPessimismOperationKeyAdapter KeyAdapter = keyHexEncoderDecoder("/dm-master/shardddl-pessimism/operation/")
+	// ShardDDLPessimismDDLsKeyAdapter is used to store last done DDLs in pessimistic model.
+	// k/v: Encode(lockID) -> DDLs.
+	ShardDDLPessimismDDLsKeyAdapter KeyAdapter = keyHexEncoderDecoder("/dm-master/shardddl-pessimism/ddls/")
```
lichunzhu, line 69: Just asking, what about `/task/downSchema/downTable`? Then we can delete this by prefix task directly.

pkg/shardddl/pessimism/operation.go
```diff
@@ -35,21 +35,23 @@ type Operation struct {
...
+	Skip   bool     `json:"skip"`   // Whether worker skip this operation
 
 	// only used to report to the caller of the watcher, do not marsh it.
 	// if it's true, it means the Operation has been deleted in etcd.
 	IsDeleted bool `json:"-"`
 }
 
 // NewOperation creates a new Operation instance.
-func NewOperation(id, task, source string, ddls []string, exec, done bool) Operation {
+func NewOperation(id, task, source string, ddls []string, exec, done bool, skip bool) Operation {
```
lichunzhu, line 46:
```suggestion
func NewOperation(id, task, source string, ddls []string, exec, done, skip bool) Operation {
```

pkg/shardddl/pessimism/lock.go
```diff
@@ -64,13 +68,46 @@ func NewLock(id, task, owner string, ddls, sources []string) *Lock {
 // TrySync tries to sync the lock, does decrease on remain, re-entrant.
 // new upstream sources may join when the DDL lock is in syncing,
 // so we need to merge these new sources.
-func (l *Lock) TrySync(caller string, ddls, sources []string) (bool, int, error) {
+func (l *Lock) TrySync(cli *clientv3.Client, caller string, ddls, sources []string, latestDoneDDLs []string) (bool, int, error) {
```
lichunzhu, line 71:
```suggestion
func (l *Lock) TrySync(cli *clientv3.Client, caller string, ddls, sources, latestDoneDDLs []string) (bool, int, error) {
```

pkg/shardddl/pessimism/lock.go
```diff
@@ -64,13 +68,46 @@ func NewLock(id, task, owner string, ddls, sources []string) *Lock {
...
 	// check DDL statement first.
 	if !utils.CompareShardingDDLs(ddls, l.DDLs) {
-		return l.remain <= 0, l.remain, terror.ErrMasterShardingDDLDiff.Generate(l.DDLs, ddls)
+		// handle conflict
+		if !utils.CompareShardingDDLs(latestDoneDDLs, ddls) && !utils.CompareShardingDDLs(latestDoneDDLs, l.DDLs) {
+			return l.remain <= 0, l.remain, terror.ErrMasterShardingDDLDiff.Generate(l.DDLs, ddls)
+		}
+
+		// current ddls idempotent, skip it.
+		if utils.CompareShardingDDLs(latestDoneDDLs, ddls) {
```
lichunzhu, line 83: We can save this function in a bool variable to save one comparison.

dm/master/shardddl/pessimist.go
```diff
@@ -394,24 +399,31 @@ func (p *Pessimist) RemoveMetaData(task string) error {
 	for _, info := range infos {
 		p.lk.RemoveLockByInfo(info)
 	}
+	lockIDSet := make(map[string]struct{}, len(ops))
```
lance6716, line 402: not used 🤔

pkg/shardddl/pessimism/ddls.go
```diff
@@ -0,0 +1,85 @@
...
+	if err != nil {
+		return clientv3.Op{}, err
+	}
+	key := common.ShardDDLPessimismDDLsKeyAdapter.Encode(lockID)
+
+	return clientv3.OpPut(key, string(data)), nil
+}
+
+// PutLatestDoneDDLs puts the last done shard DDL ddls into etcd.
+func PutLatestDoneDDLs(cli *clientv3.Client, lockID string, ddls []string) (int64, error) {
```
lance6716, line 39: may use above function

pkg/shardddl/pessimism/keeper.go
```diff
@@ -49,10 +53,47 @@ func (lk *LockKeeper) TrySync(info Info, sources []string) (string, bool, int, e
...
+// RemoveLatestDoneDDLsByTask remove last done ddls by task.
+func (lk *LockKeeper) RemoveLatestDoneDDLsByTask(task string) []string {
+	lk.mu.Lock()
+	defer lk.mu.Unlock()
+	lockIDs := make([]string, 0, len(lk.latestDoneDDLs))
+	for lockID := range lk.latestDoneDDLs {
+		if t := utils.ExtractTaskFromLockID(lockID); t == task {
+			lockIDs = append(lockIDs, lockID)
+		}
+		delete(lk.latestDoneDDLs, lockID)
```
lance6716, line 83: should we remove the lockID only when it matching the task 🤔

pkg/shardddl/pessimism/operation.go
```diff
@@ -35,21 +35,23 @@ type Operation struct {
 	DDLs   []string `json:"ddls"`   // DDL statements
 	Exec   bool     `json:"exec"`   // execute or skip the DDL statements
 	Done   bool     `json:"done"`   // whether the `Exec` operation has done
+	Skip   bool     `json:"skip"`   // Whether worker skip this operation
```
lance6716, line 38: could we don't add `Skip`, instead, use empty `DDLs`

syncer/syncer.go
```diff
@@ -979,8 +979,15 @@ func (s *Syncer) syncDDL(tctx *tcontext.Context, queueBucket string, db *DBConn,
 				tctx.L().Warn("skip shard DDL handle in pessimistic shard mode", zap.Strings("ddl", sqlJob.ddls))
 			case shardPessimistOp == nil:
 				err = terror.ErrWorkerDDLLockOpNotFound.Generate(shardInfo)
+			case shardPessimistOp.Skip:
```
lance6716, line 982: will this cause a nil pointer dereference 🤔

lance6716 (commented):
will review later

pkg/shardddl/pessimism/keeper.go
```diff
@@ -16,26 +16,30 @@ package pessimism
...
 )
 
 // LockKeeper used to keep and handle DDL lock conveniently.
 // The lock information do not need to be persistent, and can be re-constructed from the shard DDL info.
 type LockKeeper struct {
-	mu    sync.RWMutex
-	locks map[string]*Lock // lockID -> Lock
+	mu             sync.RWMutex
+	locks          map[string]*Lock                          // lockID -> Lock
+	latestDoneDDLs map[string]map[string]map[string][]string // task -> downSchema -> downTable -> ddls
```
lance6716, line 30: seems

```
		lockID           = fmt.Sprintf("%s-`%s`.`%s`", task, downSchema, downTable)
```

and `latestDoneDDLs` should be a part of a `Lock`

lance6716: Oh, i'm reverting https://github.com/pingcap/dm/pull/1626/commits/a770b5d1067488c7fd58dcf12b9fc07e6b3b91f0 😂 

for https://github.com/pingcap/dm/pull/1626#discussion_r619805000, I think if we support "delete them by prefix task directly" by delete their locks by prefix task directly. Not sure if there's such a senario

GMHDBJD: Maybe a task name with '-', I'm afraid of mis-deleting them.

lance6716:
need shfmt

lance6716 (commented):
will review later

pkg/shardddl/pessimism/lock.go
```diff
@@ -64,13 +68,47 @@ func NewLock(id, task, owner string, ddls, sources []string) *Lock {
...
+			return l.remain <= 0, l.remain, terror.ErrMasterShardingDDLDiff.Generate(l.DDLs, ddls)
+		}
+
+		// current ddls idempotent, skip it.
+		if curIdempotent {
+			log.L().Warn("conflict ddls equals last done ddls, skip it", zap.Strings("ddls", ddls), zap.String("source", caller))
+			_, _, err := PutOperations(cli, true, NewOperation(l.ID, l.Task, caller, latestDoneDDLs, false, false, true))
+			return l.remain <= 0, l.remain, err
+		}
+
```
lance6716, line 89: could you give an example of the cases that enter here 😵 ?

pkg/shardddl/pessimism/ops.go
```diff
@@ -60,12 +61,32 @@ func DeleteInfosOperations(cli *clientv3.Client, infos []Info, ops []Operation)
 	return rev, err
 }
 
-// DeleteInfosOperationsByTask deletes the shard DDL infos and operations of a specified task in etcd.
+// DeleteOperationsPutDDLs deletes the shard DDL operations and add latest done DDLs in etcd.
+// This function should often be called by DM-master when calling UnlockDDL.
```
lance6716, line 65:
```suggestion
// This function should often be called by DM-master when calling UnlockDDL and the lock is resolved.
```

pkg/shardddl/pessimism/lock.go
GMHDBJD: curIdempotent: worker1, worker2 ddl1 -> worker 2 ddl2 -> worker1 resync ddl1
otherIdempotent: worker1, worker2 ddl1 -> worker2 resync ddl1 -> worker1 ddl2

pkg/shardddl/pessimism/lock.go
```diff
@@ -64,13 +68,47 @@ func NewLock(id, task, owner string, ddls, sources []string) *Lock {
...
 	if !utils.CompareShardingDDLs(ddls, l.DDLs) {
-		return l.remain <= 0, l.remain, terror.ErrMasterShardingDDLDiff.Generate(l.DDLs, ddls)
+		curIdempotent := utils.CompareShardingDDLs(latestDoneDDLs, ddls)
+		// handle conflict
+		if !curIdempotent && !utils.CompareShardingDDLs(latestDoneDDLs, l.DDLs) {
+			return l.remain <= 0, l.remain, terror.ErrMasterShardingDDLDiff.Generate(l.DDLs, ddls)
+		}
+
+		// current ddls idempotent, skip it.
+		if curIdempotent {
```
lance6716, line 84: I still prefer that we should check `curIdempotent` before `utils.CompareShardingDDLs(ddls, l.DDLs)`

background: the lock is resolved but one non-owner is failed before saving checkpoint and resync the DDL of the resolved lock

my prefer way:
1. the failed DM-worker put the outdated DDL. if we check `curIdempotent` here, we could earlier avoid blocking that DM-worker which may caused by DDL lock

current way:
1. the failed DM-worker (worker1) put the outdated DDL, `utils.CompareShardingDDLs(ddls, l.DDLs)` is true
2. another DM-worker (worker2) put a DDL more likely a new different DDL, enter if-then branch, `curIdempotent` is false, `utils.CompareShardingDDLs(latestDoneDDLs, l.DDLs)` is true. 
    (And i think in most cases `curIdempotent` is false. And when `curIdempotent` is true, worker1 will wait more time to get skip Op)
3. in step2 we enter "other sources' ddls idempotent, skip them" branch, and here we put skip Op for worker1

Above all, worker1 may be blocked for a longer time

lance6716: And I still prefer save a binlog position through same transaction in etcd 😂 When DM-worker starting, this position has higher priority than one in downstream checkpoint

PR #1628, opened 2021-04-25
build: add more arch and os for build

GMHDBJD:
build for arm

add arch and os in github action


lance6716:
an alternative is that we watch nightly build of CI if there're hard problems 😂

lichunzhu (commented):
/lgtm

PR #1638, opened 2021-04-27
*: show an unsynced message for workers which didn't get any shard DDL

lance6716:
some users don't use `show-ddl-locks` to get unsynced tables, they use `query-status` instead, so we also show reasonable message in `query-status`

- re-fill the related field
- also fix a BUG that forget to decrypt password before open connection to upstream which introduced in https://github.com/pingcap/dm/pull/1619

### Check List 

Tests 

 - Integration test

Code changes

 - Has (user) interface methods change

Related changes

### DIFF
--- a/dm/master/server.go
+++ b/dm/master/server.go
@@ -719,6 +719,7 @@ func (s *Server) fillUnsyncedStatus(resps []*pb.QueryStatusResponse) {
 			if syncStatus == nil || len(syncStatus.UnresolvedGroups) != 0 {
 				continue
 			}
+			// TODO: look at s.optimist when `query-status` support show `UnresolvedGroups` in optimistic mode.
 			locks := s.pessimist.ShowLocks(subtaskStatus.Name, []string{resp.SourceStatus.Source})
 			if len(locks) == 0 {
 				continue
--- a/dm/master/server.go
+++ b/dm/master/server.go
@@ -710,40 +710,24 @@ func (s *Server) QueryStatus(ctx context.Context, req *pb.QueryStatusListRequest
 	return resp, nil
 }
 
-// adjust unsynced field in sync status, task-name -> set(target-table)
+// adjust unsynced field in sync status by looking at DDL locks.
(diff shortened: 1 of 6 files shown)


Ehco1996 (approved):

lichunzhu (commented):
/lgtm

dm/master/server.go
```diff
@@ -691,6 +691,45 @@ func (s *Server) QueryStatus(ctx context.Context, req *pb.QueryStatusListRequest
...
+				if _, ok := found[target]; ok {
+					continue
+				}
+				syncStatus.UnresolvedGroups = append(syncStatus.UnresolvedGroups, &pb.ShardingGroup{
+					Target:   target,
+					Unsynced: []string{"this DM-worker doesn't receive any shard DDL of this group"},
+				})
+			}
+		}
+	}
```
lichunzhu, line 731: Weak advice: can we use a separate function for this part and add a unit test for it?

lance6716: https://github.com/pingcap/dm/pull/1638/commits/31d5d1a4629fc1087f0760b6dd599b0e6a6382ca

lance6716:
/hold 

If user only queries one source not whole task, current code is not awared of unsynced workers. So I will change the implementation to use `ShowLocks` of pessimist/optimist

GMHDBJD:
any test for optimistic mode?

lance6716:
> any test for optimistic mode?

seems optimistic mode will discover sources in a shard group lazily, so a totally unsynchronized DM-worker is not counted into shard group. will check later

lance6716:
Oh currently query-status doesn't show `unresolvedGroups` for optimistic tasks. I'll add a TODO

https://github.com/pingcap/dm/blob/423c81d439d3f7ec43f8e72daa8ed21f727b53b1/syncer/status.go#L51

GMHDBJD (approved):
/lgtm

Issue #1636, opened 2021-04-27
DM-master may keeping OOM after upgrading


lance6716:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do? If possible, provide a recipe for reproducing the error.

upgrading to v2.0.2 when **not all the optimistic lock resolved**

2. What did you expect to see?

DM works well

3. What did you see instead?

DM-master become leader and OOM after some minutes or hours (in our case, 100+worker will cause memory going to 2G in around 3 minutes)

lance6716:
closed by https://github.com/pingcap/dm/pull/1635 https://github.com/pingcap/dm/pull/1639

PR #1639, opened 2021-04-28
scheduler: close workers grpc connections timely

lichunzhu:
part 2 of https://github.com/pingcap/dm/issues/1636
dm-master didn't close grpc connections before it close scheduler.

Close `scheduler.workers` grpc connections when scheduler fails to start or trying to Close.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Related changes

### DIFF
--- a/dm/master/scheduler/scheduler.go
+++ b/dm/master/scheduler/scheduler.go
@@ -166,11 +166,16 @@ func NewScheduler(pLogger *log.Logger, securityCfg config.Security) *Scheduler {
 
 // Start starts the scheduler for work.
 // NOTE: for logic errors, it should start without returning errors (but report via metrics or log) so that the user can fix them.
-func (s *Scheduler) Start(pCtx context.Context, etcdCli *clientv3.Client) error {
+func (s *Scheduler) Start(pCtx context.Context, etcdCli *clientv3.Client) (err error) {
 	s.logger.Info("the scheduler is starting")
 
 	s.mu.Lock()
-	defer s.mu.Unlock()
+	defer func() {
+		if err != nil {
+			s.CloseAllWorkers()
+		}
+		s.mu.Unlock()
+	}()
 
 	if s.started {
 		return terror.ErrSchedulerStarted.Generate()
@@ -180,7 +185,7 @@ func (s *Scheduler) Start(pCtx context.Context, etcdCli *clientv3.Client) error 
 	s.reset()           // reset previous status.
 
 	// recover previous status from etcd.
-	err := s.recoverSources(etcdCli)
+	err = s.recoverSources(etcdCli)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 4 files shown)


dm/master/scheduler/scheduler.go
```diff
@@ -239,6 +246,13 @@ func (s *Scheduler) Close() {
 	s.logger.Info("the scheduler has closed")
 }
 
+// CloseAllWorkers closes all the scheduler's workers.
+func (s *Scheduler) CloseAllWorkers() {
+	for _, worker := range s.workers {
```
lance6716, line 251: I think we should remove the worker from `s.workers`. Because in `(*Scheduler).AddWorker`, we skip adding worker if key exists (or reset `s.workers` when `(*Scheduler).Start`)

dm/master/scheduler/scheduler_test.go
```diff
@@ -1224,3 +1226,41 @@ func (t *testScheduler) TestStartStopSource(c *C) {
...
+		names  []string
+	)
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	for i := 1; i < 4; i++ {
+		names = append(names, fmt.Sprintf("worker%d", i))
+	}
+
+	for i, name := range names {
+		info := ha.NewWorkerInfo(name, fmt.Sprintf("127.0.0.1:%d", 50801+i))
```
lance6716, line 1251: How did we successful dial the port if we didn't start worker there 🤔

dm/master/scheduler/scheduler_test.go
```diff
@@ -1224,3 +1226,41 @@ func (t *testScheduler) TestStartStopSource(c *C) {
 	c.Assert(err, IsNil)
 	c.Assert(workers, HasLen, 0)
 }
+
+func checkAllWorkersClosed(c *C, s *Scheduler, closed bool) {
+	for _, worker := range s.workers {
```
lance6716, line 1231: please check `s.workers` is not empty

lichunzhu: add length check in https://github.com/pingcap/dm/pull/1639/commits/91eff928a9d8e986c1a6ce3c81299c9d4a98c238

dm/master/scheduler/scheduler.go
lichunzhu: https://github.com/pingcap/dm/blob/master/dm/master/scheduler/scheduler.go#L1732

lance6716 (approved):

dm/master/scheduler/scheduler_test.go
lance6716: seems `conn.GetState()` is IDLE, so it didn't notice the endpoint is inaccessible 🤔

GMHDBJD (approved):
/lgtm

PR #1640, opened 2021-04-28
*: refine all unit metrics

Ehco1996:
- None

I think all metrics should be treated like in `dm/master/metrics.go`
so, i make some wrapper/setter for unit metrics.

### Check List 

Tests 

- Unit test
- Integration test
- Self Test

Code changes

- [x] `dm/worker/metrics.go`
- [x] `relay/metrics.go`
- [x] `loader/metrics.go`
- [x] `dumpling/metrics.go`
- [x] `syncer/metrics.go`

Side effects

 - **Increased code complexity**
 
Related changes

### DIFF
--- a/dm/worker/metrics.go
+++ b/dm/worker/metrics.go
@@ -25,6 +25,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 
 	"github.com/pingcap/dm/dm/common"
+	"github.com/pingcap/dm/dm/pb"
 	"github.com/pingcap/dm/dumpling"
 	"github.com/pingcap/dm/loader"
 	"github.com/pingcap/dm/pkg/log"
@@ -136,5 +137,13 @@ func InitStatus(lis net.Listener) {
 	err := httpS.Serve(lis)
 	if err != nil && !common.IsErrNetClosing(err) && err != http.ErrServerClosed {
 		log.L().Error("status server returned", log.ShortError(err))
+	}
+}
+
+func updateTaskState(task, sourceID string, stage pb.Stage) {
+	if stage == pb.Stage_Stopped || stage == pb.Stage_Finished {
+		taskState.DeleteAllAboutLabels(prometheus.Labels{"task": task, "source_id": sourceID})
+	} else {
+		taskState.WithLabelValues(task, sourceID).Set(float64(stage))
 	}
 }
--- a/dm/worker/metrics.go
+++ b/dm/worker/metrics.go
@@ -140,7 +140,8 @@ func InitStatus(lis net.Listener) {
 	}
 }
 
-func updateTaskState(task, sourceID string, stage pb.Stage) {
+// UpdateTaskState update worker task stage.
+func UpdateTaskState(task, sourceID string, stage pb.Stage) {
 	if stage == pb.Stage_Stopped || stage == pb.Stage_Finished {
 		taskState.DeleteAllAboutLabels(prometheus.Labels{"task": task, "source_id": sourceID})
 	} else {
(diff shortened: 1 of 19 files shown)


lichunzhu (commented):
What's the benefit of wrapping these vectors?

Ehco1996:
> What's the benefit of wrapping these vectors?

* all metrics in dm-master is wrapped, so I think we should uniform the code？
* Provide the exported func in case other package need use it

dm/worker/metrics.go
```diff
@@ -138,3 +139,17 @@ func InitStatus(lis net.Listener) {
 		log.L().Error("status server returned", log.ShortError(err))
 	}
 }
+
+// UpdatetaskStateGauge update worker task stage.
+func UpdatetaskStateGauge(task, sourceID string, stage pb.Stage) {
```
lance6716, line 144:
```suggestion
func UpdateTaskStateGauge(task, sourceID string, stage pb.Stage) {
```

lance6716: And I'm not sure if we should delete the `Stage_Finished` tasks. Now I prefer don't delete them if they can be seen from `query-status` 😂

dumpling/dumpling.go
```diff
@@ -66,7 +66,7 @@ func (m *Dumpling) Init(ctx context.Context) error {
 
 // Process implements Unit.Process.
 func (m *Dumpling) Process(ctx context.Context, pr chan pb.ProcessResult) {
-	dumplingExitWithErrorCounter.WithLabelValues(m.cfg.Name, m.cfg.SourceID).Add(0)
+	AddDumplingExitWithErrorCounter(float64(0), m.cfg.Name, m.cfg.SourceID)
```
lance6716, line 69:
```suggestion
	AddDumplingExitWithErrorCounter(0, m.cfg.Name, m.cfg.SourceID)
```

lance6716 (commented):
will review later

dm/worker/metrics.go
Ehco1996: i vote for keep stage  `Stage_Finished`😂

pkg/binlog/event/util.go
```diff
@@ -218,7 +218,7 @@ func encodeColumnValue(v interface{}, tp byte, meta uint16) ([]byte, error) {
 	case gmysql.MYSQL_TYPE_DOUBLE:
 		value, ok := v.(float64)
 		if !ok {
-			err = terror.ErrBinlogColumnTypeMisMatch.Generate(v, reflect.TypeOf(v), reflect.TypeOf(float64(0)))
+			err = terror.ErrBinlogColumnTypeMisMatch.Generate(v, reflect.TypeOf(v), reflect.TypeOf(0))
```
lance6716, line 221: should keep float64(0) here, float64(0) and 0 has different type

syncer/metrics.go
```diff
@@ -273,23 +273,156 @@ func InitStatusAndMetrics(addr string) {
...
+func AddJobCountMetric(s *Syncer, isFinished bool, queueBucket string, tp opType, n int64) {
+	m := addedJobsTotalCounter
+	if isFinished {
+		s.count.Add(n)
+		m = finishedJobsTotalCounter
+	}
+
+	switch tp {
+	case insert:
+		m.WithLabelValues("insert", s.cfg.Name, queueBucket, s.cfg.SourceID).Add(float64(n))
```
lance6716, line 412: "insert" could be retrieved by tp.String()

lance6716 (commented):
rest lgtm

syncer/metrics.go
Ehco1996: added

Ehco1996:
close for now

PR #1635, opened 2021-04-26
etcd: fix json unmarshall error and sources not found after upgrade to 2.0.2

GMHDBJD:
### What problem does this PR solve? 

part of https://github.com/pingcap/dm/issues/1636

Info struct changed in v2.0.2, but forgot to update etcd value.

close #1644 
upgrade etcd before scheduler start

### What is changed and how it works?
Try unmarshall and convert old info to new info if unmarshall error happened.
Try upgrade etcd key before scheduler start.

### Check List 

Tests 

 - Unit test
 - upgrade test

PR #1645, opened 2021-04-30
upgrade: upgrade etcd before scheduler shart

### LINKED ISSUE #1644: Source unbound after upgrade
## Bug Report

cluster: 1 master, 2 workers, 2 sources
upgrade from v2.0.x to nightly
upgrade success
3. What did you see instead?
list-member --worker
```
"worker": {
                "msg": "",
                "workers": [
                    {
                        "name": "dm-172.28.0.111-8262",
                        "addr": "172.28.0.111:8262",
                        "stage": "bound",
                        "source": "mysql-replica-02"
                    },
                    {
                        "name": "dm-172.28.0.112-8262",
                        "addr": "172.28.0.112:8262",
                        "stage": "free",
                        "source": ""
                    }
                ]
            }
```
4. Versions of the cluster

        ```
       nightly
        ```

GMHDBJD:
close #1644 

upgrade etcd before scheduler shart

### Check List 

Tests 

 - github action

### DIFF
--- a/dm/master/bootstrap.go
+++ b/dm/master/bootstrap.go
@@ -70,6 +70,20 @@ func (s *Server) bootstrap(ctx context.Context) error {
 	return nil
 }
 
+func (s *Server) bootstrapBeforeSchedulerStart(ctx context.Context) error {
+	log.L().Info("start before scheduler start")
+	// no need for v1.0.x
+	if s.cfg.V1SourcesPath != "" {
+		return nil
+	}
+
+	err := upgrade.TryUpgradeBeforeSchedulerStart(ctx, s.etcdClient)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
 // importFromV10x tries to import/upgrade the cluster from v1.0.x.
 func (s *Server) importFromV10x(ctx context.Context) error {
 	// 1. check whether need to upgrade based on the cluster version.
--- a/dm/master/bootstrap.go
+++ b/dm/master/bootstrap.go
@@ -71,7 +71,7 @@ func (s *Server) bootstrap(ctx context.Context) error {
 }
 
 func (s *Server) bootstrapBeforeSchedulerStart(ctx context.Context) error {
-	log.L().Info("start before scheduler start")
+	log.L().Info("bootstrap before scheduler start")
 	// no need for v1.0.x
 	if s.cfg.V1SourcesPath != "" {
 		return nil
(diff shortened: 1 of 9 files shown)


GMHDBJD:
test upgrade-via-tiup failed because it uses nighlty version, #1635 patch with local build version and succeeds. PTAL

dm/master/bootstrap.go
```diff
@@ -70,6 +70,20 @@ func (s *Server) bootstrap(ctx context.Context) error {
...
 }
 
+func (s *Server) bootstrapBeforeSchedulerStart(ctx context.Context) error {
+	log.L().Info("bootstrap before scheduler start")
+	// no need for v1.0.x
+	if s.cfg.V1SourcesPath != "" {
+		return nil
+	}
+
+	err := upgrade.TryUpgradeBeforeSchedulerStart(ctx, s.etcdClient)
```
lance6716, line 80:
```suggestion
	return upgrade.TryUpgradeBeforeSchedulerStart(ctx, s.etcdClient)
```

lance6716: seems our lint is broken 🤔

pkg/upgrade/upgrade.go
```diff
@@ -204,7 +243,16 @@ func upgradeToVer3(cli *clientv3.Client, uctx Context) error {
 
 	ops := make([]clientv3.Op, 0, len(etcdKeyUpgrades))
 	for _, pair := range etcdKeyUpgrades {
-		resp, err := cli.Get(uctx.Context, pair.old.Path(), clientv3.WithPrefix())
+		resp, err := cli.Get(ctx, pair.new.Path(), clientv3.WithPrefix())
+		if err != nil {
+			return err
+		}
+		if len(resp.Kvs) > 0 {
```
lance6716, line 250: what's the benefits of we changing the upgrading detections? If user removes source configs after scheduler.Start() and DM-master failed to save current version in etcd (may caused by we write a buggy upgradeToVerX in future), the new leader will add source configs again

lance6716 (commented):
rest lgtm

dm/master/bootstrap.go
GMHDBJD: Seems revive can only detect err declare in if
```golang
if err:=...;err!=nil{
    return err
}
return nil
```

GMHDBJD:
merge in #1635

PR #1652, opened 2021-05-06
lint: add shfmt for bash scripts

Ehco1996:

None

add [shfmt](https://github.com/mvdan/sh#shfmt) to format bash scripts in tests

> tips: we can use [shell-format](https://marketplace.visualstudio.com/items?itemName=foxundermoon.shell-format)  to format our code if you use vscode.

### Check List 

 - No code

Code changes

 - None

Side effects

 - None

Related changes

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -122,16 +122,24 @@ unit_test_pkg_binlog: tools_setup
 unit_test_others: tools_setup
 	$(call run_unit_test,$(PACKAGES_OTHERS),unit_test_others)
 
-check: tools_setup fmt lint terror_check tidy_mod
+check: tools_setup lint fmt terror_check tidy_mod
 
-fmt:
-	@echo "gofumports"
-	tools/bin/gofumports -w -d -local $(PACKAGE_NAME) $(PACKAGE_DIRECTORIES) 2>&1 | awk '{print} END{if(NR>0) {exit 1}}'
+fmt: tools_setup
+	@if [[ "${nolint}" != "true" ]]; then\
+		echo "shfmt"; \
+		tools/bin/shfmt -l -w -d "tests/" ; \
+		echo "gofumports"; \
+		tools/bin/gofumports -w -d -local $(PACKAGE_NAME) $(PACKAGE_DIRECTORIES) 2>&1 | awk "{print} END{if(NR>0) {exit 1}}" ;\
+		echo "golangci-lint"; \
+		tools/bin/golangci-lint run --config=$(CURDIR)/.golangci.yml --fix --issues-exit-code=1 $(PACKAGE_DIRECTORIES) ;\
+	fi
 
 lint: tools_setup
 	@if [[ "${nolint}" != "true" ]]; then\
 		echo "golangci-lint"; \
 		tools/bin/golangci-lint run --config=$(CURDIR)/.golangci.yml --issues-exit-code=1 $(PACKAGE_DIRECTORIES); \
+		echo "run shfmt"; \
+		tools/bin/shfmt -d "tests/"; \
 	fi
 
 terror_check:
--- a/tests/online_ddl/run.sh
+++ b/tests/online_ddl/run.sh
@@ -2,98 +2,93 @@
 
 set -eu
 
-cur=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
+cur=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
 source $cur/../_utils/test_prepare
 ONLINE_DDL_ENABLE=${ONLINE_DDL_ENABLE:-true}
 BASE_TEST_NAME=$TEST_NAME
 
 function real_run() {
-    online_ddl_scheme=$1
-    run_sql_file $cur/data/db1.prepare.sql $MYSQL_HOST1 $MYSQL_PORT1 $MYSQL_PASSWORD1
-    check_contains 'Query OK, 2 rows affected'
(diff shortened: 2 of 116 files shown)


Makefile
```diff
@@ -124,9 +124,12 @@ unit_test_others: tools_setup
 
 check: tools_setup fmt lint terror_check tidy_mod
 
-fmt:
+fmt: tools_setup
 	@echo "gofumports"
 	tools/bin/gofumports -w -d -local $(PACKAGE_NAME) $(PACKAGE_DIRECTORIES) 2>&1 | awk '{print} END{if(NR>0) {exit 1}}'
+	@echo "shfmt"
+	tools/bin/shfmt  -l -w tests/
```
lance6716, line 131:
```suggestion
	tools/bin/shfmt -l -w tests/
```

lance6716: should we failed the `fmt` since it's called in `check` and `check` means not silently exit 🤔

maybe `shfmt -d`

dm/master/server.go
```diff
@@ -711,7 +711,7 @@ func (s *Server) QueryStatus(ctx context.Context, req *pb.QueryStatusListRequest
 }
 
 // adjust unsynced field in sync status by looking at DDL locks.
-// because if a DM-worker doesn't receive any shard DDL, it doesn't even know it's unsynced for itself
+// because if a DM-worker doesn't receive any shard DDL, it doesn't even know it's unsynced for itself.
```
lance6716, line 714: why didn't CI discover there's a dot missing 🤔

lance6716 (commented):
rest lgtm

dm/master/server.go
Ehco1996: I am also curious, let me cc

Ehco1996: oh it's because we open [fix: true](https://github.com/pingcap/dm/blob/4c1857c3ff4ee456a622dc0a9a05fe7b3d9e43d4/.golangci.yml#L119) in golangci , as a result golangci wiil auto fix the issue. 

i wil fix this.

Makefile
Ehco1996: i change the order of stage `check` in makefile

i think it's better to run `lint` before `fmt`, bcause `lint` will failed when golangci/shfmt find issue in our code.

Makefile
```diff
@@ -122,16 +122,24 @@ unit_test_pkg_binlog: tools_setup
...
-	tools/bin/gofumports -w -d -local $(PACKAGE_NAME) $(PACKAGE_DIRECTORIES) 2>&1 | awk '{print} END{if(NR>0) {exit 1}}'
+fmt: tools_setup
+	@if [[ "${nolint}" != "true" ]]; then\
+		echo "shfmt"; \
+		tools/bin/shfmt -l -w -d "tests/" ; \
+		echo "gofumports"; \
+		tools/bin/gofumports -w -d -local $(PACKAGE_NAME) $(PACKAGE_DIRECTORIES) 2>&1 | awk "{print} END{if(NR>0) {exit 1}}" ;\
+		echo "golangci-lint"; \
+		tools/bin/golangci-lint run --config=$(CURDIR)/.golangci.yml --fix --issues-exit-code=1 $(PACKAGE_DIRECTORIES) ;\
+	fi
```
GMHDBJD, line 135: Why fmt contains lint? Now if we execute make build, golangci-lint and shfmt will run multiple times

dm/master/server.go
GMHDBJD: Can we use set it true as default and false for [check-and-build](https://github.com/pingcap/dm/blob/master/.github/workflows/check-and-build.yml) in CI? So we can fix it automatically by `make build`, and discover it by GitHub action

Makefile
Ehco1996: golangci-lint with arg `--fix`  means fix unlint code, so this is fmt not lint any more.

after this change, `make check` will run in this order:
```
...
lint (both golangci-lint and shfmt will just check code and not fix)
fmt ( golangci-lint and shfmt will fix the wrong style code)
...
```

GMHDBJD (approved):
/lgtm

lance6716:
please resolve conflicts then we could merge this PR

lance6716:
there's a failing CI @Ehco1996 https://internal.pingcap.net/idc-jenkins/blue/organizations/jenkins/dm_ghpr_test/detail/dm_ghpr_test/9221/pipeline

tests/online_ddl/run.sh
```diff
@@ -2,98 +2,93 @@
...
-    check_sync_diff $WORK_DIR $cur/conf/diff_config.toml
-
-    run_sql_file_online_ddl $cur/data/db1.increment.sql $MYSQL_HOST1 $MYSQL_PORT1 $MYSQL_PASSWORD1 online_ddl $online_ddl_scheme
-    run_sql_file_online_ddl $cur/data/db2.increment.sql $MYSQL_HOST2 $MYSQL_PORT2 $MYSQL_PASSWORD2 online_ddl $online_ddl_scheme
-
-    # manually create index to pass check_sync_diff
-    run_sql_tidb "show create table online_ddl.t_target"
-    check_not_contains "KEY \`name\`"
-    run_sql_tidb "alter table online_ddl.t_target add key name (name)"
-
```
lance6716, line 52: these lines should not be deleted

Ehco1996: may something wrong when i rebase master ...

Ehco1996:
@lance6716  all ci passed , PTAL

PR #1657, opened 2021-05-07
loader: revert #1544 to reduce risks

lichunzhu:
We need to release v2.0.3 recently. Some improvement PRs will be reverted to reduce the risk of bugs.

revert https://github.com/pingcap/dm/pull/1544

### Check List 

Tests 

 - Unit test
 - Integration test

Side effects


lance6716:
and we should revert this "revert PR" once v2.0.3 is released?

(will review later)

lichunzhu:
> and we should revert this "revert PR" once v2.0.3 is released?
> 
> (will review later)

Yes

Issue #1656, opened 2021-05-07
binlog event filters may has conflict with online schema change tools


lance6716:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do? If possible, provide a recipe for reproducing the error.

https://asktug.com/t/topic/69966

2. What did you expect to see?

no error occured and filters take effects, so index didn't get created at downstream

3. What did you see instead?

error: `online ddl metadata for pt temporary table test._dim_warehouse1_new not found`

PR #1659, opened 2021-05-07
syncer: save empty SQL when online DDL is filtered

lance6716:
close https://github.com/pingcap/dm/issues/1656

- save an empty SQL when online DDL is filtered, and remove it when read it from online DDL component
- use `?` not fmt.Sprintf to build a query

### Check List 

Tests 

 - Integration test

Code changes

 - Has persistent data change (not likely, because if there's quote in source-id, an error already occurred)

Side effects

 - Increased code complexity

Related changes

### DIFF
--- a/syncer/syncer.go
+++ b/syncer/syncer.go
@@ -1752,11 +1752,17 @@ func (s *Syncer) handleQueryEvent(ev *replication.QueryEvent, ec eventContext, o
 		skipBinlogDurationHistogram.WithLabelValues("query", s.cfg.Name, s.cfg.SourceID).Observe(time.Since(ec.startTime).Seconds())
 		ec.tctx.L().Warn("skip event", zap.String("event", "query"), zap.String("statement", originSQL), zap.String("schema", usedSchema))
 		*ec.lastLocation = *ec.currentLocation // before record skip location, update lastLocation
+
 		// we try to insert an empty SQL to s.onlineDDL, because ignoring it will cause a "not found" error
+		if s.onlineDDL == nil {
+			return s.recordSkipSQLsLocation(*ec.lastLocation)
+		}
+
 		stmts, err2 := parserpkg.Parse(parser2, originSQL, "", "")
 		if err2 != nil {
 			ec.tctx.L().Info("failed to parse a filtered SQL for online DDL", zap.String("SQL", originSQL))
 		}
+		// if err2 != nil, stmts should be nil so below for-loop is skipped
 		for _, stmt := range stmts {
 			if _, ok := stmt.(ast.DDLNode); ok {
 				tableNames, err3 := parserpkg.FetchDDLTableNames(usedSchema, stmt)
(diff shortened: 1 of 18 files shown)


tests/online_ddl/conf/dm-task.yaml
```diff
@@ -68,6 +70,14 @@ column-mappings:
     target-column: "id"
     arguments: ["2", "", "t"]
 
+filters:
+  filter-rule-index:
+    schema-pattern: "*"
+    table-pattern: "*"
+    events: ["create index","drop index"]
+    sql-pattern: ["ALTER\\s.*TABLE[\\s\\S]*ADD\\s+KEY", "ALTER\\s.*TABLE[\\s\\S]*DROP\\s+KEY"]
```
GMHDBJD, line 78: How about filter index statement for special name, so we will have positive and negative test.

syncer/syncer.go
```diff
@@ -1751,6 +1752,21 @@ func (s *Syncer) handleQueryEvent(ev *replication.QueryEvent, ec eventContext, o
...
+			ec.tctx.L().Info("failed to parse a filtered SQL for online DDL", zap.String("SQL", originSQL))
+		}
+		for _, stmt := range stmts {
+			if _, ok := stmt.(ast.DDLNode); ok {
+				tableNames, err3 := parserpkg.FetchDDLTableNames(usedSchema, stmt)
+				if err3 != nil {
+					continue
+				}
+				// nolint:errcheck
+				s.onlineDDL.Apply(ec.tctx, tableNames, "", stmt)
```
GMHDBJD, line 1767: Should we check onlineDDL is nil?

syncer/syncer.go
```diff
@@ -1751,6 +1752,21 @@ func (s *Syncer) handleQueryEvent(ev *replication.QueryEvent, ec eventContext, o
 		skipBinlogDurationHistogram.WithLabelValues("query", s.cfg.Name, s.cfg.SourceID).Observe(time.Since(ec.startTime).Seconds())
 		ec.tctx.L().Warn("skip event", zap.String("event", "query"), zap.String("statement", originSQL), zap.String("schema", usedSchema))
 		*ec.lastLocation = *ec.currentLocation // before record skip location, update lastLocation
+		// we try to insert an empty SQL to s.onlineDDL, because ignoring it will cause a "not found" error
+		stmts, err2 := parserpkg.Parse(parser2, originSQL, "", "")
+		if err2 != nil {
+			ec.tctx.L().Info("failed to parse a filtered SQL for online DDL", zap.String("SQL", originSQL))
+		}
+		for _, stmt := range stmts {
```
GMHDBJD, line 1760: if else?

GMHDBJD (commented):
rest LGTM. make build to fix fmt.

tests/online_ddl/conf/dm-task.yaml
lance6716: what's the "negative", do you mean filter a SQL that parser can't parse, and test `handle-error skip` it?

lichunzhu (commented):
/lgtm

PR #1670, opened 2021-05-11
Support migration data from source with different time_zone by default

glorv:
Automatically adapt source and target DB time_zone settings so users don't need to explicitly set `timezone` in dm config. 

close https://github.com/pingcap/dm/issues/1691

- Deprecated the `timezone` setting in dm config. (maybe we should directly remove this config field)
- Always set `timezone='+00:00'` for all db connections in dumpling/loader/syncer.

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

Side effects

Related changes


lance6716:
> maybe we should directly remove this config field

we can't remove it because I remember there's a checking about undecoded config items, to fail the user of mistyping. (And I think this also disables smoothly downgrade from a version with more config items)

glorv:
@lance6716 @lichunzhu @GMHDBJD  PTAL

dm/master/scheduler/scheduler.go
```diff
@@ -380,7 +380,7 @@ func (s *Scheduler) GetSourceCfgByID(source string) *config.SourceConfig {
 	if !ok {
 		return nil
 	}
-	clone := cfg
+	clone := *cfg
```
lance6716, line 383: suddenly found that, because `SourceConfig` contains some reference type member, we can't clone it by this way 😂 

Not checked what changes does the invoker of this function make to `SourceConfig`

https://play.golang.org/p/pwI414VqK0O

lance6716 (commented):
will review later

dm/master/scheduler/scheduler.go
lance6716: yes, original code (and seems no effects). 

Not sure we should change it to deepclone or don't clone, in this PR I think left a TODO is OK

lance6716: or let other guy check it 😂

lichunzhu (commented):
/lgtm

lance6716:
we should log a warning if overwrite user's `timezone` in session variables. rest lgtm

glorv:
> we should log a warning if overwrite user's `timezone` in session variables. rest lgtm

done

lance6716 (approved):

PR #1679, opened 2021-05-12
test: refine action trigger

GMHDBJD:
refine action trigger to speed up github action

- binlog-9999/upstream-switch/arm build workflow will not trigger by PR
- tmate session reduce to 30 minutes


.github/workflows/upgrade-via-tiup.yml
```diff
@@ -82,7 +82,7 @@ jobs:
         path: go/src/github.com/pingcap/dm
 
     - name: Build
-      if: ${{ github.ref != 'refs/heads/main' }}
+      if: ${{ github.ref != 'refs/heads/master' }}
```
lance6716, line 85: originally, because DM uses master not main as branch name, all these steps will not be filterd. 🤔 Do you want to skip this action?

lance6716 (commented):
also should add v2.0.3 in previous_v2

.github/workflows/upgrade-via-tiup.yml
GMHDBJD: Yes. If it's master branch, we use nightly version, don't need to build local version

lance6716 (approved):

lance6716:
@Ehco1996 ptal

lichunzhu (commented):
/lgtm

Issue #1667, opened 2021-05-10
cover all possible scenario in upgrading tests 


lance6716:
## Feature Request

**Is your feature request related to a problem? Please describe:**

In order to avoid my bug like https://github.com/pingcap/dm/issues/1644 happening again in future, we should cover every possible scenarios before upgrading, and check main commands works well after upgrading.

**Describe the feature you'd like:**

we should cover

- pessimistic lock not resolved
- optimistic lock not resolved
- online DDL not finished
- start-task could continue from checkpoint at sync stage
- start-task from checkpoint at load stage (not sure)
- pause-task will not resume automaticlly, and resume-task could work
- relay enabled 

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

PR #1661, opened 2021-05-08
tests: add mariaDB to version upgrading CI

lance6716:
### What problem does this PR solve? 
part of https://github.com/pingcap/dm/issues/1280

### What is changed and how it works?
as title

### Check List 

Tests 

 - No code

Related changes

 - Need to cherry-pick to the release branch

PR #1681, opened 2021-05-12
tests: cover all possible scenarios in upgrading tests

lichunzhu:
Try to fix part of https://github.com/pingcap/dm/issues/1667

Add uncovered cases in https://github.com/pingcap/dm/issues/1667.

### Check List 

Tests 

 - Unit test
 - Integration test

Related changes


lance6716:
mariaDB is partly added in https://github.com/pingcap/dm/pull/1661, could replace it

tests/tiup/lib.sh
```diff
@@ -115,6 +149,23 @@ function exec_incremental_stage2() {
...
+	# pessimistic shard ddls
+	exec_sql mysql1 3306 "ALTER TABLE $DB5.$TBL2 ADD COLUMN c3 INT;"
+	exec_sql mariadb2 3306 "ALTER TABLE $DB6.$TBL2 ADD COLUMN c3 INT;"
+
+	# prepare pessimistic incremental data
+	exec_sql mysql1 3306 "INSERT INTO $DB5.$TBL1 (c1, c2, c3) VALUES (203, '203', 203);"
+	exec_sql mysql1 3306 "INSERT INTO $DB5.$TBL2 (c1, c2, c3) VALUES (204, '204', 204);"
+	exec_sql mariadb2 3306 "INSERT INTO $DB6.$TBL2 (c1, c2, c3) VALUES (213, '213', 213);"
+	exec_sql mariadb2
+  3306 "INSERT INTO $DB6.$TBL3 (c1, c2, c3) VALUES (214, '214', 214);"
```
lichunzhu, line 168:
```suggestion
	exec_sql mariadb2 3306 "INSERT INTO $DB6.$TBL3 (c1, c2, c3) VALUES (214, '214', 214);"
```

tests/tiup/upgrade-from-v2.sh
```diff
@@ -53,6 +58,12 @@ function upgrade_to_current_v2() {
 }
 
 function migrate_in_v2 {
+	run_dmctl_with_retry $CUR_VER "query-status" "Running" 2 "Paused" 1
```
lance6716, line 61: I also want to check `show-ddl-locks` and `query-status` here 😂

lance6716 (commented):
> [ ] start-task could continue from checkpoint at sync stage

It's better we `stop-task` before upgrading, and check `start-task` after upgrading works fine.

> [ ] relay enabled

I think this is not covered, so unmarked it in original issue

lichunzhu:
> It's better we `stop-task` before upgrading, and check `start-task` after upgrading works fine.
Could you please give me a reason? I think `pause-task` can test both `pause-task will not resume automaticlly, and resume-task could work` and `recover from checkpoint in sync stage`
> I think this is not covered, so unmarked it in original issue
One source has relay enabled.

lance6716:
>  I think pause-task can test both pause-task will not resume automaticlly, and resume-task could work and recover from checkpoint in sync stage

good idea

> One source has relay enabled 

not found `start-relay`

lance6716 (approved):

PR #1683, opened 2021-05-12
scheduler: check invalid source before apply last bound

lichunzhu:
When dm-master degrade from higher versions and then upgrade to higher versions, there might be some bounds without source configs.

Check source configs before we apply last bounds.

### Check List 

Tests 

 - Unit test
 - Integration test

Related changes


lance6716 (commented):
I prefer remove the bound/unbound relations containing unfound sources in `recoverWorkersBounds`. 🤔

Ehco1996 (approved):

dm/master/scheduler/scheduler.go
```diff
@@ -1548,11 +1556,14 @@ func (s *Scheduler) tryBoundForWorker(w *Worker) (bounded bool, err error) {
 
 	// randomly pick one from unbounds
 	if source == "" {
-		for source = range s.unbounds {
-			s.logger.Info("found unbound source when worker bound",
-				zap.String("worker", w.BaseInfo().Name),
-				zap.String("source", source))
-			break // got a source.
+		for unboundedSource := range s.unbounds {
```
lance6716, line 1559: after `recoverWorkersBounds`, not existing sources are not put in `s.unbounds`, so I think this logic could be removed

dm/master/scheduler/scheduler.go
```diff
@@ -1584,6 +1595,13 @@ func (s *Scheduler) tryBoundForWorker(w *Worker) (bounded bool, err error) {
 // called should update the s.unbounds.
```
lance6716, line 1595:
```suggestion
// caller should update the s.unbounds.
```

dm/master/scheduler/scheduler.go
```diff
@@ -1584,6 +1595,13 @@ func (s *Scheduler) tryBoundForWorker(w *Worker) (bounded bool, err error) {
...
 func (s *Scheduler) tryBoundForSource(source string) (bool, error) {
 	var worker *Worker
+	// 0. check whether this source has source configuration. If not, we shouldn't bound this but just abandon it
+	// This usually happens when dm didn't handle the source compatibility correctly. For example, downgrade from a higher dm version
+	// We can just mark this source bounded to abandon this source. After we fix the compatibility problem dm-master will restart the scheduler or manually add this source
+	if _, ok := s.sourceCfgs[source]; !ok {
+		s.logger.Warn("source configuration not found for the source to bound, add this source to unbounded", zap.String("source", source))
+		return true, nil
```
lance6716, line 1603: maybe 
a) we should `return false, nil`, to inform caller add this source to unbound. or
b) we maintain the invariant: "after scheduler is started, all sources in `bound`, `unbound` must have their source config, sources in `lastBound` may not have source configs". So if `tryBoundForSource` is not called for `lastBound`, this check could be deleted.

I prefer b) @GMHDBJD

lance6716 (commented):
PTAL again @GMHDBJD

dm/master/scheduler/scheduler.go
GMHDBJD: b) LGTM

dm/master/scheduler/scheduler_test.go
```diff
@@ -1267,3 +1303,70 @@ func (t *testScheduler) TestCloseAllWorkers(c *C) {
...
+	s.etcdCli = etcdTestCli
+	// found source configs before bound
+	s.sourceCfgs[sourceID1] = &config.SourceConfig{}
+	s.sourceCfgs[sourceID2] = &config.SourceConfig{}
+	s.unbounds[sourceID1] = struct{}{}
+	s.unbounds[sourceID2] = struct{}{}
+	c.Assert(s.AddWorker(workerName1, workerAddr1), IsNil)
+	c.Assert(s.AddWorker(workerName2, workerAddr2), IsNil)
+
+	wg.Add(2)
```
lance6716, line 1336: wg could be remove 🤔

lance6716 (commented):
rest lgtm

dm/master/scheduler/scheduler_test.go
lichunzhu: add `wg.Wait` in the end

GMHDBJD (approved):
/lgtm

PR #1698, opened 2021-05-17
*: update go version to 1.16

lance6716:
close https://github.com/pingcap/dm/issues/1686

simply search and replace 1.13 to 1.16

### Check List 

Tests 

 - No code

Related changes


lance6716:
seems we should update go version in CI 🤔 
https://ci.pingcap.net/blue/organizations/jenkins/dm_ghpr_new_test/detail/dm_ghpr_new_test/192/pipeline

Issue #1697, opened 2021-05-16
CREATE TABLE IF NOT EXISTS may causes error like `Column count doesn't match value count: x (columns) vs y (values)`


lance6716:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do? If possible, provide a recipe for reproducing the error.
https://asktug.com/t/topic/92863

   1. at upstream: create table test.t (c int primary key); 
   2. start a "all" mode task
   3. at upstream: create table if not exists test.t (c1 int, c2 int); insert into test.t values (1);
   4. query-status

2. What did you expect to see?
no error

3. What did you see instead?
gen insert sqls failed, schema: test, table: t: Column count doesn't match value count: 2 (columns) vs 1 (values)"

Ehco1996:
`create table test.t if not exists t (c1 int, c2 int); insert into test.t values (1);` -> mysql syntas err

should update to ` create table if not exists test.t (c1 int, c2 int); insert into test.t values (1);`

PR #1699, opened 2021-05-17
syncer: load table structure from dump files

lance6716:
close https://github.com/pingcap/dm/issues/1697

- as title
- originally we use MySQL 5.7 & 8.0 as upstreams, they has different default CHARSET. Because DM ignores "table already exists" error and fetches table from downstream, only one table with one CHARSET will be created at downstream and DM loads it for two sharding tables. Now, DM will see table strcuture of diferent CHARSET, so set default in advance.

### Check List 

Tests 

 - Unit test
 - Integration test

Side effects

 - Breaking backward compatibility (but fix a bug)

Related changes


lichunzhu (commented):
/lgtm

_utils/terror_gen/errors_release.txt
```diff
@@ -277,10 +277,10 @@ ErrSyncerUnitInjectDDLOnly,[code=36023:class=sync-unit:scope=internal:level=low]
 ErrSyncerUnitInjectDDLWithoutSchema,[code=36024:class=sync-unit:scope=internal:level=low], "Message: injected DDL %s without schema name not valid"
 ErrSyncerUnitNotSupportedOperate,[code=36025:class=sync-unit:scope=internal:level=medium], "Message: op %s not supported"
 ErrSyncerUnitNilOperatorReq,[code=36026:class=sync-unit:scope=internal:level=medium], "Message: nil request not valid"
-ErrSyncerUnitDMLColumnNotMatch,[code=36027:class=sync-unit:scope=internal:level=high], "Message: Column count doesn't match value count: %d (columns) vs %d (values)"
+ErrSyncerUnitDMLColumnNotMatch,[code=36027:class=sync-unit:scope=internal:level=high], "Message: Column count doesn't match value count: %d (columns) vs %d (values), Workaround: You could use `operate-schema` to get and set the table structure, and please check the log files to see if a related DDL has been skipped."
```
GMHDBJD, line 280: I think `Please check the log files to see if a related DDL has been skipped, and you could use operate-schema to get and set the table structure` will be better.😂

syncer/syncer.go
```diff
@@ -2258,6 +2265,13 @@ func (s *Syncer) trackDDL(usedSchema string, sql string, tableNames [][]*filter.
 		}
 	}
 
+	if tryFetchDownstreamTable {
```
GMHDBJD, line 2268: Can we fetch the upstream table instead of downstream?

GMHDBJD (commented):
rest LGTM

syncer/syncer.go
lance6716: DM may require an "old" table structure, but could only get latest one.

I think downstream table is OK, because if downstream table strcuture is wrong DML will fail as well. The issue which this PR solves is that table is created at full phase, so schema tracker didn't track it.

GMHDBJD: I'm afraid for optimistic mode, downstream table may changed by other source.

GMHDBJD: But seems no good way to get the upstream table at certain pos.

GMHDBJD (approved):
/lgtm

syncer/syncer.go
lance6716: bad design of MySQL binlog 😭

GMHDBJD:
Will downschema table different with upstream, such as user manually create table in downstream

lance6716:
> Will downstream table be different with upstream? such as user manually create table in downstream.

DM in fact converted every CREATE TABLE to CREATE TABLE IF NOT EXISTS. Do you think we should disable that?

If so, when user manually create table in downstream, he should execute CREATE TABLE and schema-tracker use this table structure (but we can't rely on user) 

/hold

lance6716:
Because schema-tracker is intended to track the missing table structure in binlog, I think the following table structure is ground truth (should drop conflictive table structure first):

1. create table sql file in dump phase
2. CREATE TABLE in query event

and we should attach IF NOT EXISTS just before executed to downstream to avoid interference

We also fetch downstream table, but marked it as "not from upstream", so when meet CREATE TABLE IF NOT EXISTS in query event we could correctly decide doing it @GMHDBJD

GMHDBJD:
Maybe we have another solution, create table via dump file when begin to sync? Though it cannot fix incremental mode. But incremental mode usually have table info in checkpoint, otherwise, if it's a new incremental task, more manually operation for user is acceptable?

lance6716:
i agree on "create table via dump file when begin to sync"

and I did find out a case need further logic, may leave for future

lance6716:
/hold

fetching from downstream is too slow, should read dump files

lance6716:
originally we use MySQL 5.7 & 8.0 as upstreams, they has different default CHARSET. Because DM ignores "table already exists" error and fetches table from downstream, only one table with one CHARSET will be created at downstream and DM loads it for two sharding tables. Now, DM will see table strcuture of diferent CHARSET, so set default in advance.

syncer/syncer.go
```diff
@@ -1129,23 +1130,78 @@ func (s *Syncer) Run(ctx context.Context) (err error) {
...
+				}
+			}
+			for _, dbAndFile := range tableFiles {
+				db, file := dbAndFile[0], dbAndFile[1]
+				filepath := path.Join(s.cfg.Dir, file)
+				content, err3 := common2.GetFileContent(filepath)
+				if err3 != nil {
+					hasErr = true
+					tctx.L().Warn("fail to read file for creating table in schema tracker",
+						zap.String("db", db),
```
lichunzhu, line 1171: Can we put these things in one function? They look so strange here.

lichunzhu (commented):
How do we deal with incremental mode if we meet a `create table if not exists` DDL binlog?

lance6716:
Now we can't correctly handle `create table if not exists` for incremental tasks, which because DM will not fetch downstream tables first. As discussed in https://github.com/pingcap/dm/pull/1699#issuecomment-846374842

lichunzhu (approved):
LGTM

lance6716:
PTAL @GMHDBJD

Ehco1996 (approved):

Issue #1586, opened 2021-04-12
can't smoothly continue sync when switch GTID from on to off


lance6716:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do? If possible, provide a recipe for reproducing the error.

https://asktug.com/t/topic/69319/36?u=lance6716

2. What did you expect to see?

continue sync from previous location

3. What did you see instead?

sync from start

lance6716:
closed by https://github.com/pingcap/dm/pull/1723

relay's issue has been seperated to https://github.com/pingcap/dm/issues/1460

PR #1723, opened 2021-05-25
syncer: fix syncer gtid auto switch from off to on

lichunzhu:
https://github.com/pingcap/dm/issues/1586 part 1

Add auto adjust gtid before we start a task.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

 - Increased code complexity

Related changes


dm/worker/server.go
```diff
@@ -254,8 +254,10 @@ func (s *Server) doStartKeepAlive() {
 }
 
 func (s *Server) stopKeepAlive() {
-	s.kaCancel()
-	s.kaWg.Wait()
+	if s.kaCancel != nil {
```
lance6716, line 257: in which case `stopKeepAlive` is called before or concurrent with `startKeepAlive`? we may need a lock if so 😂

lance6716 (commented):
will review later

Is the test case needed to be added to other_integrations.txt?

dm/worker/server.go
lichunzhu: 1. dm-worker/main.go `s.Start()`
2. Start() return an error before `s.startKeepAlive()`. e.g.: worker port already in use
3. dm-worker/main.go `s.Close()`, report a nil pointer info in stdout

pkg/binlog/position.go
```diff
@@ -316,6 +316,39 @@ func CompareLocation(location1, location2 Location, cmpGTID bool) int {
...
+		if canCmp {
+			if cmp != 0 {
+				return cmp
+			}
+			if cmp = compareIndex(location1.Suffix, location2.Suffix); cmp != 0 {
+				return cmp
+			}
+			if location1.gtidSet != nil && location1.gtidSet.String() != "" {
+				return cmp
+			}
```
GMHDBJD, line 336: What if both gtid sets are empty, but suffix are not? e.g. location1:(pos1,"",1), location2:(pos2,"",0)

tests/adjust_gtid/run.sh
```diff
@@ -0,0 +1,149 @@
...
+		gtid2=$(grep "GTID:" $WORK_DIR/worker1/dumped_data.$TASK_NAME/metadata | awk -F: '{print $2,":",$3}' | tr -d ' ')
+		name1=$(grep "Log: " $WORK_DIR/worker2/dumped_data.$TASK_NAME/metadata | awk -F: '{print $2}' | tr -d ' ')
+		pos1=$(grep "Pos: " $WORK_DIR/worker2/dumped_data.$TASK_NAME/metadata | awk -F: '{print $2}' | tr -d ' ')
+		gtid1=$(grep "GTID:" $WORK_DIR/worker2/dumped_data.$TASK_NAME/metadata | awk -F: '{print $2,":",$3}' | tr -d ' ')
+	fi
+
+	check_checkpoint $SOURCE_ID1 $name1 $pos1 $gtid1
+	check_checkpoint $SOURCE_ID2 $name2 $pos2 $gtid2
+	dmctl_stop_task_with_retry $TASK_NAME $MASTER_PORT
+	clean_gtid
```
GMHDBJD, line 112: Maybe start a source without gtid and than restart with gtid closer to the user scenario? Moreover, adjust gtid in relay seems not be implemented and tested in this test

tests/adjust_gtid/run.sh
```diff
@@ -0,0 +1,149 @@
...
+
+	export GO_FAILPOINTS='github.com/pingcap/dm/syncer/AdjustGTIDExit=return(true)'
```
GMHDBJD, line 114: dumplicate export

pkg/binlog/position.go
lichunzhu: I think `Suffix` is only used in `handle-error`?

tests/adjust_gtid/run.sh
lichunzhu: 1. I think cleaning gtid data is the same. Operation source again is a little complicated so I use this way to simplify the test.
2. adjust GTID in relay will be implemented in the next PR

cmd/dm-worker/main_test.go
```diff
@@ -56,6 +56,7 @@ func TestRunMain(_ *testing.T) {
 	defer func() { utils.OsExit = oldOsExit }()
 	utils.OsExit = func(code int) {
 		log.L().Info("os exits", zap.Int("exit code", code))
+		_ = utils.KillMySelf()
```
lance6716, line 59: I can't remember why we need this helper function now. maybe the process won't exit after this test function returns?

if so, I think the reason may be that lthe test function should be named `TestMain` not `TestRunMain`, please check if this works.

pkg/binlog/reader/reader.go
```diff
@@ -42,3 +42,9 @@ type Reader interface {
 	// Status returns the status of the reader.
 	Status() interface{}
 }
+
+// Streamer provides the ability to get binlog event from remote server or local file.
+type Streamer interface {
```
lance6716, line 47: can't we use this interface from streamer folder? so we could write

streamer.GetGTIDsForPos and reader.GetGTIDsForPos

pkg/binlog/position.go
```diff
@@ -316,6 +316,39 @@ func CompareLocation(location1, location2 Location, cmpGTID bool) int {
...
+			// empty GTIDSet, then compare by position
+			log.L().Warn("both gtidSets are empty, will compare by position", zap.Stringer("location1", location1), zap.Stringer("location2", location2))
+		} else {
+			// if can't compare by GTIDSet, then compare by position
+			log.L().Warn("gtidSet can't be compared, will compare by position", zap.Stringer("location1", location1), zap.Stringer("location2", location2))
+		}
+	}
+
+	cmp := ComparePosition(location1.Position, location2.Position)
```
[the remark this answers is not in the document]
lance6716: Does this function always called with location2 as empty location? If so, we could not be bothered to add an unit test, just simplifying the logic and tighten function name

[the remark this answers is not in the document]
lichunzhu: Don't understand. What is `canCmp` used for?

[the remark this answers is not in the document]
lance6716: such as location1 and location2 come from different upstream, and the two upstream didn't enable GTID.

So if they set binlog with different filename, we could detect the error and reject sync.

pkg/binlog/reader/reader.go
lichunzhu: Do you mean declare this in `streamer` folder? This will cause a circular reference.

pkg/binlog/position.go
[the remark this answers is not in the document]
lichunzhu: I don't think we can get `canCmp` from comparing binlog filenames. It can't explain anything.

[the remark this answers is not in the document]
lichunzhu: `binlog filename` is a setting for mysql. It doesn't contain uuid info.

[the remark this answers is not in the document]
lance6716: I misunderstand the scenario, I thought this function compares any two upstream binlog location, but by design it should only compare one upstream with gtid turned on/off.

pkg/binlog/reader/util.go
```diff
@@ -30,26 +30,14 @@ import (
...
-		Pos:  0,
-	}
-	err := r.StartSyncByPos(startPos)
-	if err != nil {
-		return nil, err
-	}
-	defer r.Close()
-
+// GetGTIDsForPosStreamer tries to get GTID sets for the specified binlog position (for the corresponding txn) from a Streamer.
+func GetGTIDsForPosStreamer(ctx context.Context, r Streamer, endPos gmysql.Position) (gtid.Set, error) {
```
lance6716, line 34:
```suggestion
func GetGTIDsForPosFromStreamer(ctx context.Context, r Streamer, endPos gmysql.Position) (gtid.Set, error) {
```

syncer/checkpoint.go
```diff
@@ -781,7 +781,8 @@ func (cp *RemoteCheckPoint) Load(tctx *tcontext.Context) error {
 			gset,
 		)
 		if isGlobal {
-			if binlog.CompareLocation(location, binlog.NewLocation(cp.cfg.Flavor), cp.cfg.EnableGTID) > 0 {
+			// Use CompareLocationAsPossible here to make sure checkpoint can be updated if gset is empty
+			if binlog.CompareLocationAsPossible(location, binlog.NewLocation(cp.cfg.Flavor), cp.cfg.EnableGTID) > 0 {
```
lance6716, line 785: Does this function always called with location2 as empty location? If so, we could not be bothered to add an unit test, just simplifying the logic and tighten function name

syncer/syncer.go
```diff
@@ -1128,23 +1130,30 @@ func (s *Syncer) Run(ctx context.Context) (err error) {
...
-				tctx.L().Info("try to remove loaded files")
-				metadataFile := path.Join(s.cfg.Dir, "metadata")
-				if err = os.Remove(metadataFile); err != nil {
-					tctx.L().Warn("error when remove loaded dump file", zap.String("data file", metadataFile), zap.Error(err))
-				}
-				if err = os.Remove(s.cfg.Dir); err != nil {
-					tctx.L().Warn("error when remove loaded dump folder", zap.String("data folder", s.cfg.Dir), zap.Error(err))
-				}
+	}
+	needFlushCheckpoint, err := s.adjustGlobalPointGTID(tctx)
```
lance6716, line 1134: we could move it to Syncer.Init to avoid resuming triggers it, after `s.checkpoint.Load(tctx)`. And flush checkpoint inside `adjustGlobalPointGTID`

tests/adjust_gtid/run.sh
```diff
@@ -0,0 +1,155 @@
...
+	run_sql "select binlog_name,binlog_pos,binlog_gtid from dm_meta.${TASK_NAME}_syncer_checkpoint where id=\"$source_id\" and is_global=1" $TIDB_PORT $TIDB_PASSWORD
+	check_contains $expected_name
+	check_contains $expected_pos
+	if [[ -n $expected_gtid ]]; then
+		check_contains $expected_gtid
+	fi
+}
+
+function run() {
+	run_sql_both_source "SET @@GLOBAL.SQL_MODE='ANSI_QUOTES,NO_AUTO_VALUE_ON_ZERO'"
```
lance6716, line 47: recover these changes

tests/adjust_gtid/run.sh
```diff
@@ -0,0 +1,155 @@
...
+	sed -i "s/binlog-pos-placeholder-2/4/g" $WORK_DIR/dm-task.yaml
+	dmctl_start_task "$WORK_DIR/dm-task.yaml" "--remove-meta"
+	# check task has started
+	check_metric $WORKER1_PORT "dm_worker_task_state{source_id=\"mysql-replica-01\",task=\"$TASK_NAME\"}" 3 1 3
+	check_metric $WORKER2_PORT "dm_worker_task_state{source_id=\"mysql-replica-02\",task=\"$TASK_NAME\"}" 3 1 3
+
+	# use sync_diff_inspector to check full dump loader
+	check_sync_diff $WORK_DIR $cur/conf/diff_config.toml
+
+	# $worker1_run_source_1 > 0 means source1 is operated to worker1
```
lance6716, line 95: now source1 is always bound to worker1

syncer/syncer.go
lance6716: ( but failpoint should be left here)

lance6716 (commented):
rest lgtm

syncer/syncer.go
lichunzhu: I don't think we should move this to `Init` because we need to adjust the `Meta.Pos` and it's loaded here. Besides, moving to `Init` may make `syncer.Init` exceed init time.

GMHDBJD (commented):
/lgtm

lance6716 (approved):

lance6716:
add require-LGT1 for GMHDBJD

GMHDBJD (approved):
LGTM

PR #1743, opened 2021-06-03
ci: try fix chaos mesh

Ehco1996:
截屏2021-06-03 上午8 27 28

something wrong with the latest chaos-mesh， try use v1.0.0

### Check List 

Tests 

 - No code

Code changes

Side effects

Related changes


GMHDBJD:
I think we better figure out why `scheduler` removed in the new version of chaos-mesh and update the config file rather than downgrade it.

Ehco1996:
> I think we better figure out why `scheduler` removed in the new version of chaos-mesh and update the config file rather than downgrade it.

i found this: https://github.com/chaos-mesh/chaos-mesh/discussions/1520, seems `scheduler` became a new `crd` in leatest version, if we want to use this, we need update our chaos tesat yaml.

given there is lot's of PR pending by this `chaos-mesh` job, lock version can be  a quick fix?

lance6716 (approved):

GMHDBJD (approved):
LGTM

PR #1782, opened 2021-06-17
syncer: support filter row by SQL expression

lance6716:

support filter row changes by SQL expression

add an item in configuration files, example.

```
expression-filter:
  expr-1:
    schema: "expr_filter"
    table: "t1"
    insert-value-expr: "c_null IS NOT NULL"
```

note that `schema` and `table` don't support pattern matching.

parse the expression into SQL AST, using TiDB's AST rewriter to generate executor AST, evaluate each row changes to decide passing it.

### Check List 

Tests 

 - [x] white-box: check can handle data types that go-mysql returns to DM
 - [x] every MySQL data type can be handled
 - [x] can filter by INSERT, old and new value of UDPATE, DELETE
 - [x] works during DDL, dropping the filtering column makes the filter disabled, adding the column back enables it again
 - [x] works on generated column
 - [x] **the timestamp value in expression should explicitly specify timezone** `c_timestamp = '2021-01-01 12:34:56.5678+08:00'`
 - [x] works on SQL functions `ts > NOW()`
 - [x] works on multiple column `a*a + b*b = c*c`
 - [x] benchmark
    sysbench using DM master branch https://github.com/pingcap/dm/commit/c2a7514f49a3be2e0499322adbb8a15a45641b0f and TiDB v5.1.0: **CPS 28k ~ 30k**, left mountain in picture 1
    Using DM this PR https://github.com/pingcap/dm/pull/1782/commits/a591b169b4c58bad306c7a2eb2ee6028e99e3f3f, not enable expression filter, **CPS  28k ~ 29k**, middle mountain in picture 1
    Using DM this PR https://github.com/pingcap/dm/pull/1782/commits/a591b169b4c58bad306c7a2eb2ee6028e99e3f3f, filter 1% INSERT of 1 of 4 tables, **CPS 27k ~ 29k**, right mountain in picture 1
    Using DM this PR https://github.com/pingcap/dm/pull/1782/commits/a591b169b4c58bad306c7a2eb2ee6028e99e3f3f, filter 99% INSERT of 1 of 4 tables, **CPS 27k ~ 30k**, picture 2
    
image

image

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes


syncer/dml.go
```diff
@@ -130,6 +147,26 @@ func genUpdateSQLs(param *genDMLParam) ([]string, [][]string, [][]interface{}, e
 			oriChangedValues = extractValueFromData(oriChangedData, ti.Columns)
 		}
 
+		for _, expr := range oldValueFilters {
```
glorv, line 150: Does user expected `old-value-filter` and `new-value-filter` to be `and` relation instead of `or` relation?  Or what if user want `and` but we only support `or`

lance6716: How about we add a field in config, to let user specify "AND" "OR"

glorv: I think we can change the `UpdateFilter` to include both old and new filter, if user only want to change one of them then just leave the other empty, thus the semantic  is always `AND`. For the `OR` case, user can just add two separated filter

lance6716: OK.

We might also make (insert-value-expr), (update-old-value-expr, update-new-value-expr), (delete-value-expr) these three groups exclusive, to indicate AND semantic.

lance6716:
/run-all-tests

failed test recorded

glorv (approved):

pkg/utils/common.go
```diff
@@ -209,3 +217,79 @@ func NonRepeatStringsEqual(a, b []string) bool {
...
+		case uint32:
+			d = uint64(v)
+		case uint:
+			d = uint64(v)
+		case decimal.Decimal:
+			d = v.String()
+		}
+		datum := types.NewDatum(d)
+
+		// TODO: should we use timezone of upstream?
```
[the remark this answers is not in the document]
lance6716: User may intuitively use upstream or local timezone. Anyway, we could tell user always explicitly set timezone like `insert-value-expr: "c_timestamp = '2021-01-01 12:34:56.5678+08:00'"`

pkg/utils/common.go
```diff
@@ -209,3 +217,79 @@ func NonRepeatStringsEqual(a, b []string) bool {
...
+		case uint8:
+			d = uint64(v)
+		case uint16:
+			d = uint64(v)
+		case uint32:
+			d = uint64(v)
+		case uint:
+			d = uint64(v)
+		case decimal.Decimal:
+			d = v.String()
```
qw4990, line 283: Better to add a default branch.

lance6716: what to do in default branch? `d` is directly used other than above types.

syncer/expr_filter_group.go
```diff
@@ -0,0 +1,200 @@
...
+	key := dbutil.TableName(db, table)
+	delete(g.insertExprs, key)
+	delete(g.updateOldExprs, key)
+	delete(g.updateNewExprs, key)
+	delete(g.deleteExprs, key)
+}
+
+// SkipDMLByExpression returns true when given row matches the expr, which means this row should be skipped.
+func SkipDMLByExpression(row []interface{}, expr expression.Expression, upstreamCols []*model.ColumnInfo) (bool, error) {
+	log.L().Debug("will evaluate the expression", zap.Stringer("expression", expr), zap.Any("raw row", row))
```
GMHDBJD, line 169: We may add some metrics in the future.

tests/expression_filter/conf/diff_config.toml
```diff
@@ -0,0 +1,56 @@
+# diff Configuration.
```
GMHDBJD, line 1: Seems no use? How about delete this file.

GMHDBJD (commented):
rest LGTM

_utils/terror_gen/errors_release.txt
```diff
@@ -167,6 +167,10 @@ ErrConfigEtcdParse,[code=20035:class=config:scope=internal:level=high], "Message
 ErrConfigMissingForBound,[code=20036:class=config:scope=internal:level=high], "Message: source bound %s doesn't have related source config in etcd"
 ErrConfigBinlogEventFilter,[code=20037:class=config:scope=internal:level=high], "Message: generate binlog event filter, Workaround: Please check the `filters` config in source and task configuration files."
 ErrConfigGlobalConfigsUnused,[code=20038:class=config:scope=internal:level=high], "Message: The configurations as following %v are set in global configuration but instances don't use them, Workaround: Please check the configuration files."
+ErrConfigExprFilterManyExpr,[code=20039:class=config:scope=internal:level=high], "Message: expression filter can only specify one of (insert, update, delete) expressions, but %s has specified %v, Workaround: If you want to filter by A or B, please write two filters."
+ErrConfigExprFilterNotFound,[code=20040:class=config:scope=internal:level=high], "Message: mysql-instance(%d)'s expression-filters %s not exist in expression-filter, Workaround: Please check the `expression-filters` config in task configuration file."
```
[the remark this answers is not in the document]
lance6716: support filter UPDATE that old value matches expression A and new value matches expression B, like 

```
  update_old_and_new:
    schema: "expr_filter"
    table: "t5"
    update-old-value-expr: "c = 1"
    update-new-value-expr: "c = 2"
```

syncer/syncer.go
```diff
@@ -2451,6 +2462,7 @@ func (s *Syncer) trackDDL(usedSchema string, sql string, tableNames [][]*filter.
 			ec.tctx.L().Error("cannot track DDL", zap.String("schema", usedSchema), zap.String("statement", sql), log.WrapStringerField("location", ec.currentLocation), log.ShortError(err))
 			return terror.ErrSchemaTrackerCannotExecDDL.Delegate(err, sql)
 		}
+		s.exprFilterGroup.ResetExprs(srcTable.Schema, srcTable.Name)
```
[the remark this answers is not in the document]
lance6716: IMO `exprFilterGroup` is reset because the table structure which came from schema tracker has been changed. `Syncer.reset()` doesn't reset schema tracker, so I think it's no need to reset `exprFilterGroup`.

GMHDBJD (approved):
LGTM

PR #1796, opened 2021-06-21
grafana: change some dashboard type to graph

Ehco1996:

as title

1. Manual editing of dashboards in grafana v6.1.6
2. export the dashboard json
3. Replace the datasources field in the json with `"${DS_TEST-CLUSTER}",`.

### Check List 

Tests 
 - No code

Code changes
 - No code

Side effects

Related changes


lichunzhu (approved):
LGTM

dm/dm-ansible/scripts/dm.json
```diff
@@ -1,40 +1,4 @@
 {
-  "__inputs": [
```
lance6716, line 2: what's the effects of these missing lines 🤔

dm/dm-ansible/scripts/dm.json
```diff
@@ -2590,2868 +2687,2634 @@
...
-            {
-              "expr": "dm_syncer_replication_lag{task=~\"$task\", instance=~\"$instance\"}",
-              "format": "time_series",
-              "intervalFactor": 2,
-              "refId": "A"
-            }
-          ],
-          "thresholds": "",
-          "title": "replicate lag",
-          "type": "singlestat",
```
lance6716, line 3471: why can't we only change this line

Ehco1996: I've tried changing just this area before, but it messes up the layout of the dashboard afterwards

dm/dm-ansible/scripts/dm.json
Ehco1996: found [this](https://community.grafana.com/t/set---inputs-values-when-pushing-a-dashboard-json-config-to-grafana-via-the-api/2836)

It should have no effect on us

lance6716 (changes requested):
https://github.com/pingcap/dm/blob/804ff138a2e9b5f190ba16272f7e645654b8e296/monitoring/dashboards/dashboard.go#L79-L83

dm/dm-ansible/scripts/dm.json
```diff
@@ -6056,20 +5993,10 @@
...
-      "24h",
-      "2d",
-      "7d",
-      "30d"
-    ]
+    "time_options": ["5m", "15m", "1h", "6h", "12h", "24h", "2d", "7d", "30d"]
   },
   "timezone": "",
-  "title": "Test-Cluster-DM-task",
+  "title": "dm-test-DM-task",
```
lichunzhu, line 5999: This line is not replaced.

Ehco1996: fixed

dm/dm-ansible/scripts/dm.json
lichunzhu: suggest keeping these

lichunzhu (commented):
Please help fix the test

dm/dm-ansible/scripts/dm.json
lance6716: keep same structure with other repo

https://github.com/pingcap/tidb/blob/master/metrics/grafana/tidb.json

Ehco1996: fixed and i know why `__inputs` is gone


```
In Grafana v5.3.4+ the export modal has new checkbox for sharing for external use (other instances). If the checkbox is not checked then the __inputs section will not be included in the exported JSON file.
```

tests/ha_master/run.sh
```diff
@@ -248,6 +248,7 @@ function test_list_member() {
 }
 
 function run() {
+	run_sql_both_source "SET @@GLOBAL.SQL_MODE=''"
```
lance6716, line 251: why add this line

Ehco1996: [`ERROR 1064 (42000) at line 8: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '"$.id")' at line 1`](https://ci.pingcap.net/blue/organizations/jenkins/dm_ghpr_new_test/detail/dm_ghpr_new_test/737/pipeline/149)

this failed caused by sqlmode not right ?

lance6716: yes, but in case "ha_master" we didn't change sql_mode, and the cases are run independently in my mind. we should investigate why the sql_mode is changed to a bad value (maybe 'ANSI_QUOTES')

Ehco1996: this is beacuse we reuse test pod in ci, and other test such as `all_mode` will set sqlmode to `ANSI_QUOTES,NO_AUTO_VALUE_ON_ZERO`

lance6716: we should reset the sql_mode after finishing `all_mode`, or reset before any case starting. Only resetting here is not acceptable.

And I'm not sure if we have reuse `golang` pod, or pods can be accessedrun concurrently 😵 cc @GMHDBJD @glorv

lance6716:
maybe the base branch should be master, or we should port it to master in another PR

tests/ha_master/run.sh
Ehco1996: ok  i will fix this

lance6716 (commented):
I just forget what's the root cause of this issue 😂 we should remove `singlestat`?

Ehco1996:
> I just forget what's the root cause of this issue 😂 we should remove `singlestat`?

root cause is we need chang some dashboard singlestat -> graph

lance6716:
here's a diff after `jq '.panels |= sort_by(.id)' dm.json`, comparing this PR and v2.0.3, for your reference

https://www.diffchecker.com/VbjsCnfS

lance6716:
> > I just forget what's the root cause of this issue 😂 we should remove `singlestat`?
> 
> root cause is we need chang some dashboard singlestat -> graph

should we change all `singlestat`? or why this PR only change part of them

Ehco1996:
> > > I just forget what's the root cause of this issue 😂 we should remove `singlestat`?
> > 
> > 
> > root cause is we need chang some dashboard singlestat -> graph
> 
> should we change all `singlestat`? or why this PR only change part of them

only change two (lag and remainning time) because  when metrics is 0 `singlestat` dashboard  show `NaN`

Some clients feel this is misleading

lance6716 (approved):

lichunzhu (approved):
LGTM

PR #1809, opened 2021-06-24
syncer: fix syncer not being triggered by ticker

### LINKED ISSUE #1798: secondsBehindMaster didn't return 0 after workload is gone
## Bug Report

https://asktug.com/t/topic/93842

3. What did you see instead?

Ehco1996:
close #1798 

as title

### Check List 

Tests 

 - Integration test

Code changes

Side effects
 - Increased code complexity

Related changes

### DIFF
--- a/syncer/syncer.go
+++ b/syncer/syncer.go
@@ -1193,6 +1193,7 @@ func (s *Syncer) syncDML(tctx *tcontext.Context, queueBucket string, db *DBConn,
 	var err error
 	var affect int
 	ticker := time.NewTicker(waitTime)
+	defer ticker.Stop()
 	for {
 		select {
 		case sqlJob, ok := <-jobChan:
--- a/syncer/syncer.go
+++ b/syncer/syncer.go
@@ -1192,6 +1192,7 @@ func (s *Syncer) syncDML(tctx *tcontext.Context, queueBucket string, db *DBConn,
 
 	var err error
 	var affect int
+	ticker := time.NewTicker(waitTime)
 	for {
 		select {
 		case sqlJob, ok := <-jobChan:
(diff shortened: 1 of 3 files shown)


syncer/syncer.go
```diff
@@ -1192,6 +1192,7 @@ func (s *Syncer) syncDML(tctx *tcontext.Context, queueBucket string, db *DBConn,
 
 	var err error
 	var affect int
+	ticker := time.NewTicker(waitTime)
```
lance6716, line 1195: ticker.Stop()

lance6716 (commented):
rest lgtm

lance6716 (approved):

GMHDBJD (approved):
LGTM

lichunzhu:
Is this bug caused by `successF` isn't triggered every time?
If so, I don't think this PR solved this bug.
@Ehco1996 @lance6716

GMHDBJD:
> Is this bug caused by `successF` isn't triggered every time?
> If so, I don't think this PR solved this bug.
> @Ehco1996 @lance6716

You are right. 🤣

Ehco1996:
> Is this bug caused by `successF` isn't triggered every time?
> If so, I don't think this PR solved this bug.
> @Ehco1996 @lance6716

@lichunzhu  i don't understand this, `successF` only called after `executeSQLs` success . and 

the root cause is when no job in queue, lag will never be updated because ticker not work

lichunzhu:
@Ehco1996 
After this PR, lag update logic will still not be used. Change time after to ticker doesn't work. Because job length is still zero.

Ehco1996:
> @Ehco1996
> After this PR, lag update logic will still not be used. Change time after to ticker doesn't work. Because job length is still zero.

oh...i know,i will provide another PR to fix this

lance6716:
this PR is wrong because ticker will be generating tick at background, however we require wait 100ms after entering `select`. 

please help us fix it in https://github.com/pingcap/dm/pull/1855 @Ehco1996

lance6716:
maybe we should call `Ticker.Reset` @Ehco1996

glorv:
> maybe we should call `Ticker.Reset` @Ehco1996

In this way, we can simply call `time.After` instead

lance6716:
> > maybe we should call `Ticker.Reset` @Ehco1996
> 
> In this way, we can simply call `time.After` instead

OK. I'm a little bit wory about performance when concurrency is high 
image

glorv:
ok. Then maybe we should call `timer.Reset` at each loop start. 😂️

Issue #1717, opened 2021-05-24
DM's schema tracker doesn't track DDL when we skip some DDLs which will lead to column count unmatch


lichunzhu:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do? If possible, provide a recipe for reproducing the error.
Configure `binlog-event-filter` to filter some DDLs to downstream database.

2. What did you expect to see?
The `binlog-event-filter` takes effect and the replication can continue automatically.

3. What did you see instead?
DM task will pause with error 'column count unmatch'.

Root Cause:
1. DM will not track the DDLs to skip in the schema tracker. Then DM will face count unmatch errors when DM tries to parse DML events from the upstream binlogs.
2. If DM tracks these DDLs in the schema tracker, the generated DMLs will contain skipped columns now. Since the DDL is not executed in the downstream, the DML will also fail in the downstream database.

4. Versions of the cluster

    - DM version (run `dmctl -V` or `dm-worker -V` or `dm-master -V`):
       DM v2.0.3

PR #1775, opened 2021-06-15
syncer: check block-allow list before online DDL components

lance6716:
### What problem does this PR solve? 

close https://github.com/pingcap/dm/issues/1776

### What is changed and how it works?

move block-allow list check and binlog filter check before online DDL components

### Check List 

Tests 

 - Integration test

Code changes

 - Has interface methods change

Side effects

Related changes

 - Need to cherry-pick to the release branch
 - Need to be included in the release note

PR #1824, opened 2021-06-30
syncer: split handleQueryEvent to small functions

Ehco1996:

Before we can solve #1717, we need to clean up the logic for `handleQueryEvent`.

as title

### Check List 

Tests 

 - Integration test

Code changes

 - None

Side effects

- None

Related changes

### DIFF
--- a/syncer/ddl.go
+++ b/syncer/ddl.go
@@ -14,15 +14,20 @@
 package syncer
 
 import (
+	"reflect"
 	"time"
 
 	"github.com/go-mysql-org/go-mysql/replication"
+	"github.com/pingcap/failpoint"
 	"github.com/pingcap/parser"
 	"github.com/pingcap/parser/ast"
 	"github.com/pingcap/tidb-tools/pkg/filter"
 	"go.uber.org/zap"
 
+	"github.com/pingcap/dm/dm/config"
+	"github.com/pingcap/dm/pkg/binlog/event"
 	tcontext "github.com/pingcap/dm/pkg/context"
+	"github.com/pingcap/dm/pkg/log"
 	parserpkg "github.com/pingcap/dm/pkg/parser"
 	"github.com/pingcap/dm/pkg/terror"
 	"github.com/pingcap/dm/pkg/utils"
@@ -34,6 +39,12 @@ type parseDDLResult struct {
 	stmt   ast.StmtNode
 	ignore bool
 	isDDL  bool
+}
+
+type shardingDDLInfo struct {
+	name       string
+	tableNames [][]*filter.Table
+	stmt       ast.StmtNode
 }
 
 func (s *Syncer) parseDDLSQL(sql string, p *parser.Parser, schema string) (result parseDDLResult, err error) {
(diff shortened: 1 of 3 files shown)


tests/online_ddl/run.sh
```diff
@@ -49,12 +48,11 @@ function real_run() {
 	# manually create index to pass check_sync_diff
 	run_sql_tidb "show create table online_ddl.t_target"
 	check_not_contains "KEY \`name\`"
-	run_sql_tidb "alter table online_ddl.t_target add key name (name)"
+	# TODO why need this?
+	# run_sql_tidb "alter table online_ddl.t_target add key name (name)"
 
-	echo "use sync_diff_inspector to check increment data"
```
lance6716, line 54: `echo`s are helpful when read CI log to locate where the failure happened

tests/online_ddl/run.sh
```diff
@@ -49,12 +48,11 @@ function real_run() {
 	# manually create index to pass check_sync_diff
 	run_sql_tidb "show create table online_ddl.t_target"
 	check_not_contains "KEY \`name\`"
-	run_sql_tidb "alter table online_ddl.t_target add key name (name)"
+	# TODO why need this?
```
lance6716, line 51: sync-diff will report "different table structure"

Ehco1996: root cause located: binlog filter don't filter sql

input sql is:
```
ALTER TABLE `online_ddl`.`t1` ADD INDEX `name`(`name`)`
```
* paser return this sql type is `AlterTableStmt`, but we set filter event type is `create index`
* filter regular expressions:`"ALTER\\s.*TABLE[\\s\\S]*ADD\\s+KEY\\s+NAME"` also does not match this sql.

lance6716: I remember in query events the sql is written as-is, so it should be `alter table t2 add key name (name);` rather than your sql.

lance6716: maybe caused by you delete this filter logic in https://github.com/pingcap/dm/pull/1824/files#diff-d3b44c58774ef21a03b59014f103c9ff35671c9d131ce9dc3d6b480e9ce0004dL40

don't change the logic if you don't totally understand them.

Ehco1996: * the input raw is `ALTER TABLE `online_ddl`.`_t2_new` add key name (name)`, I just found it by print sql  after`case *replication.QueryEvent:`

* I deliberately deleted the repetitive `skipQuery` because i am refarctoring hande ddl logic and i don't want to call `skipQuery` twice

* The main refactoring plan is [here](https://github.com/Ehco1996/dm/blob/0d1888456680c44b6aaa7bc908597fd80d307445/syncer/syncer.go#L1919) I deliberately deleted the repetitive logic


thie PR is still **WIP**,  I will ask for your help to review it when it is finished

lance6716: the reason why we filter SQL in two place is: if the SQL can't be parsed, it should support filtering by regular expression in a very early stage. But if it can be parsed, some transformation is needed to decide if it's should be filtered.

Also it's better to find a way to unify them.

You could verify the reasons of some code with us before made changes, to reduce unnecesary work.

tests/online_ddl/conf/dm-task.yaml
```diff
@@ -73,8 +73,12 @@ filters:
   filter-rule-index:
     schema-pattern: "*"
     table-pattern: "*"
-    events: ["create index","drop index"]
-    sql-pattern: ["ALTER\\s.*TABLE[\\s\\S]*ADD\\s+KEY\\s+NAME", "ALTER\\s.*TABLE[\\s\\S]*DROP\\s+KEY"]
+    events: ["create index", "drop index"]
+    sql-pattern:
+      [
+        "alter table(.*?)add (key|index)(.*?)name",
```
lance6716, line 79: should not change original test case if no strong reason. User executed `ADD KEY` so he will write an `ADD KEY` filter

syncer/syncer.go
```diff
@@ -1934,184 +1933,54 @@ func (s *Syncer) handleRowsEvent(ev *replication.RowsEvent, ec eventContext) err
 }
 
 func (s *Syncer) handleQueryEvent(ev *replication.QueryEvent, ec eventContext, originSQL string) error {
-	if originSQL == "BEGIN" {
```
lance6716, line 1937: removing this line will cause performance regression, because BEGIN is a hot path

syncer/syncer.go
```diff
@@ -1934,184 +1933,54 @@ func (s *Syncer) handleRowsEvent(ev *replication.RowsEvent, ec eventContext) err
...
-	var (
-		sqls                []string
-		onlineDDLTableNames map[string]*filter.Table
-	)
-
-	// for DDL, we don't apply operator until we try to execute it. so can handle sharding cases
-	// We use default parser because inside function where need parser, sqls are came from parserpkg.SplitDDL, which is StringSingleQuotes, KeyWordUppercase and NameBackQuotes
-	sqls, onlineDDLTableNames, err = s.resolveDDLSQL(ec.tctx, parser.New(), parseResult.stmt, usedSchema)
+	sqls, onlineDDLTableNames, err := s.filterAndResolveDDLSQL(ev, ec, originSQL)
+	usedSchema := string(ev.Schema)
```
lance6716, line 1961: check `err` right after it was returned

tests/online_ddl/conf/dm-task.yaml
Ehco1996: user raw sql is  `alter table t2 add key name (name);`
this sql will be rewrite to 
```
ALTER TABLE `online_ddl`.`_t1_new` add key name (name)
```
in pt, and this sql cannot be filtered by the binlog filter

The reason we were able to pass the test before is that this sql was ignored for `ignore, err := s.skipQuery(nil, nil, sql)` (table name:`_t1_new` not match with block-allow-list)

---
but in new logic, we don't `skipQuery` twice, so we need update this binlog filter rule

lance6716: as I said in https://github.com/pingcap/dm/pull/1824#discussion_r663740810 if you can't find a way to properly keep the old behaviour and no strong reasons, don't delete that skip

we should change the code instead of requiring users to change their usages.

tests/online_ddl/run.sh
lance6716: DON'T mark my comments as resolved, I don't think it's resolved

syncer/ddl.go
```diff
@@ -33,25 +40,22 @@ type parseDDLResult struct {
...
-		return parseDDLResult{
-			stmt:   nil,
-			ignore: false,
-			isDDL:  false,
-		}, err
-	} else if ignore {
+// filterDDLSQL filter two type of sql: 1. tidb not support 2. not ddl sql such as begin/flush.
+func (s *Syncer) filterDDLSQL(sql string, p *parser.Parser) (result parseDDLResult, err error) {
+	sql = utils.TrimCtrlChars(sql)
+	if ignore := utils.IsBuildInSkipDDL(sql); ignore || sql == "BEGIN" {
```
lance6716, line 52: `sql == "BEGIN"` here is duplicated with the first line of `handleQueryEvent`, we could
- keep that line, remove this check. Or
- remove the check in `handleQueryEvent`, but a benchmark is needed.

syncer/ddl.go
```diff
@@ -33,25 +40,22 @@ type parseDDLResult struct {
...
-	// check skip before parse (used to skip some un-supported DDLs)
-	ignore, err := s.skipQuery(nil, nil, sql)
-	if err != nil {
-		return parseDDLResult{
-			stmt:   nil,
-			ignore: false,
-			isDDL:  false,
-		}, err
-	} else if ignore {
+// filterDDLSQL filter two type of sql: 1. tidb not support 2. not ddl sql such as begin/flush.
```
lance6716, line 49: this function wants to filter sql of "tidb not support", and we provided `IsBuildInSkipDDL` to achieve that. But we might miss some pattern and need user to specify a regular expression to filter it.

so `ignore, err := s.skipQuery(nil, nil, originSQL)` should be called before any parsing.

We might add a test in CI, mariaDB has lots of unsupported SQL, we could execute one of them and check the binlog filter could work.

syncer/ddl.go
```diff
@@ -295,8 +276,298 @@ func (s *Syncer) clearOnlineDDL(tctx *tcontext.Context, targetSchema, targetTabl
...
+			ec.tctx.L().Error("cannot track DDL", zap.String("schema", usedSchema), zap.String("statement", sql), log.WrapStringerField("location", ec.currentLocation), log.ShortError(err))
+			return terror.ErrSchemaTrackerCannotExecDDL.Delegate(err, sql)
+		}
+		s.exprFilterGroup.ResetExprs(srcTable.Schema, srcTable.Name)
+	}
+	return nil
+}
+
+// filterAndResolveDDLSQL 1. parses the raw sql from upstream and returns the sqls that need to be handled. 2.handle onlineddl.
+func (s *Syncer) filterAndResolveDDLSQL(ev *replication.QueryEvent, ec eventContext, originSQL string) (sqls []string, tables map[string]*filter.Table, err error) {
```
GMHDBJD, line 379: I think put this function in syncer.go will be better since it call other function like `skipBinlogDurationHistogram`

Ehco1996:
/hold

will rebase after https://github.com/pingcap/dm/pull/1775 mergeed

lance6716:
> /hold
> 
> will rebase after #1775 mergeed

merged now

Ehco1996:
> > /hold
> > will rebase after #1775 mergeed
> 
> merged now

@lance6716  after your commit, i don't need to change skip logic 👍👍👍

so i just mv some func to another file PTAL

PR #1855, opened 2021-07-10
syncer: update lag metric calculation logic

### LINKED ISSUE #1798: secondsBehindMaster didn't return 0 after workload is gone

Ehco1996:
close #1798 

* update lag metric even if no job in queue
* resets the time interval for each loop to prevent a certain amount of time being spent on the previous ticker execution to `executeSQLs` resulting in the next tikcer not waiting for the full waitTime.
* use different worker lag key in each `syncDML` goroutine

### Check List 

Tests 

- IT test

Code changes

Side effects
- None

Related changes


lance6716:
the value of parameter `queueBucket` is shared by multiple syncDML, because the value is generated by i % n
we'd better not using a shared value as map key

syncer/syncer.go
```diff
@@ -1248,7 +1246,7 @@ func (s *Syncer) syncDML(tctx *tcontext.Context, queueBucket string, db *DBConn,
 				successF()
 				clearF()
 			}
-		case <-ticker.C:
+		case <-time.After(waitTime):
```
lance6716, line 1249: maybe try ticker.Reset before `select`.

`time.After` will generate a new NewTimer every time, which is not good to performance in fact.

lance6716 (commented):
And we could use failpoint to change the `waitTime`, to check log the interval is no less than it (to prevent former bug).

Ehco1996:
> And we could use failpoint to change the `waitTime`, to check log the interval is no less than it (to prevent former bug).

 Checking the logging interval between syncs is difficult because there are several threads doing this, any suggestions? @lance6716

lance6716:
> > And we could use failpoint to change the `waitTime`, to check log the interval is no less than it (to prevent former bug).
> 
> Checking the logging interval between syncs is difficult because there are several threads doing this, any suggestions? @lance6716

maybe reduce worker count, or log the worker id (queue name without modulo) to distinguish

tests/_utils/check_ticker_interval.py
```diff
@@ -0,0 +1,29 @@
+import sys
+
+
```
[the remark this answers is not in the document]
Ehco1996: it's already formated by [black](https://github.com/psf/black)

syncer/syncer.go
```diff
@@ -1260,6 +1272,12 @@ func (s *Syncer) syncDML(tctx *tcontext.Context, queueBucket string, db *DBConn,
 				}
 				successF()
 				clearF()
+			} else if currentJobCnt == 0 && !s.workerIniting.Load() {
+				// update lag metric even if there is no job in the queue
+				// metric will only be updated when all wokers have been initialised to avoid data races.
+				tctx.L().Debug("no job in queue, update lag to zero",
+					zap.String("queueName", queueBucket), zap.Int("current ts", int(time.Now().Unix())))
+				s.updateReplicationLag(nil, queueBucket)
```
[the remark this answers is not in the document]
Ehco1996: add a new func `dmlWorkerLagKey` to calc woker lag key, and removed var `workerIniting`

syncer/syncer.go
```diff
@@ -202,7 +202,7 @@ type Syncer struct {
 
 	tsOffset            atomic.Int64             // time offset between upstream and syncer, DM's timestamp - MySQL's timestamp
 	secondsBehindMaster atomic.Int64             // current task delay second behind upstream
-	workerLagMap        map[string]*atomic.Int64 // worker's sync lag key:queueBucketName val: lag
+	workerLagMap        map[string]*atomic.Int64 // worker's sync lag key:WorkerLagKey val: lag
```
glorv, line 205: Shall we just use `[]*atomic.Int64` here? the workerId should be the slice index?

Ehco1996: I think the string key is better for reading as there are skipped jobs and ddl jobs that also need to be calculated separately.

glorv: ok, lgtm. I'm not sure if we we pack worker log related logic into a separated struct then since ddl and skip jog are special logic, we can move them out of the map and convert to map into a slice.

tests/metrics/run.sh
```diff
@@ -0,0 +1,121 @@
...
+	cnt=$(echo "${query_msg}" | jq -r --arg min_val $min_val '.sources[].subTaskStatus[].sync | select((.secondsBehindMaster|tonumber)>($min_val | tonumber)).secondsBehindMaster' | wc -l)
+	if [ $cnt != $need_cnt ]; then
+		echo "check secondsBehindMaster faild, cnt: $cnt need_cnt: $need_cnt"
+		exit 1
+	fi
+}
+
+function test_syncer_metrics() {
+	# add changeTickerInterval to keep metric from updating to zero too quickly when there is no work in the queue.
+	export GO_FAILPOINTS="github.com/pingcap/dm/syncer/BlockSyncerUpdateLag=return(\"ddl,1\");github.com/pingcap/dm/syncer/changeTickerInterval=return(10)"
```
lance6716, line 23: maybe remove the `changeTickerInterval` inject here and increase the batch.

tests/metrics/run.sh
```diff
@@ -0,0 +1,121 @@
...
+
+	# check the time interval between dm-worker updates when the queue is empty, the desired time interval is 1s
+	kill_dm_worker
+	export GO_FAILPOINTS="github.com/pingcap/dm/syncer/changeTickerInterval=return(1)"
+	run_dm_worker $WORK_DIR/worker1 $WORKER1_PORT $cur/conf/dm-worker1.toml
+	check_rpc_alive $cur/../bin/check_worker_online 127.0.0.1:$WORKER1_PORT
+	run_dm_worker $WORK_DIR/worker2 $WORKER2_PORT $cur/conf/dm-worker2.toml
+	check_rpc_alive $cur/../bin/check_worker_online 127.0.0.1:$WORKER2_PORT
+	sleep 3 # we need sleep 3 because current ticker interval is 1 and we need at least 2 ticker's log
+	python2 $cur/../_utils/check_ticker_interval.py $WORK_DIR/worker1/log/dm-worker.log 1
```
lance6716, line 101: other python scripts use shebang, not a explicit `python` call here

syncer/syncer.go
```diff
@@ -239,7 +239,6 @@ func NewSyncer(cfg *config.SubTaskConfig, etcdClient *clientv3.Client) *Syncer {
 	}
 	syncer.recordedActiveRelayLog = false
 	syncer.workerLagMap = make(map[string]*atomic.Int64)
```
lance6716, line 241: may reserve a size for `workerLagMap`

syncer/syncer.go
```diff
@@ -1260,6 +1270,12 @@ func (s *Syncer) syncDML(tctx *tcontext.Context, queueBucket string, db *DBConn,
 				}
 				successF()
 				clearF()
+			} else {
+				// update lag metric even if there is no job in the queue
+				// metric will only be updated when all wokers have been initialised to avoid data races.
+				tctx.L().Debug("no job in queue, update lag to zero",
+					zap.String("workerLagKey", workerLagKey), zap.Int("current ts", int(time.Now().Unix())))
```
lance6716, line 1277:
```suggestion
					zap.String("workerLagKey", workerLagKey), zap.Int64("current ts", time.Now().Unix()))
```

syncer/syncer.go
```diff
@@ -1388,11 +1404,11 @@ func (s *Syncer) Run(ctx context.Context) (err error) {
 		s.wg.Add(1)
 		name := queueBucketName(i)
 		s.queueBucketMapping = append(s.queueBucketMapping, name)
-		s.workerLagMap[name] = atomic.NewInt64(0)
+		s.workerLagMap[dmlWorkerLagKey(i)] = atomic.NewInt64(0)
 		go func(i int, name string) {
 			ctx2, cancel := context.WithCancel(ctx)
 			ctctx := s.tctx.WithContext(ctx2)
-			s.syncDML(ctctx, name, s.toDBConns[i], s.jobs[i])
+			s.syncDML(ctctx, name, s.toDBConns[i], s.jobs[i], i)
```
lance6716, line 1411: maybe pass the value of `dmlWorkerLagKey(i)` to avoid call it again inside `syncDML`

lance6716 (commented):
rest lgtm

syncer/syncer.go
Ehco1996: ok, maybe anothr PR i will do this
BTW could you tell me why we should use slice rather than map ?

syncer/syncer.go
lance6716: now **avoid call it again** outside syncDML

tests/_utils/check_ticker_interval.py
```diff
@@ -0,0 +1,29 @@
...
+def check_worker_ticker_interval(log_path, min_interval):
+    """
+    check the time interval between dm-worker updates when the queue is empty
+
+    log_path: dm-worker-log path
+    min_interval: min interval
+    """
+    logs = []
+    with open(log_path, "r") as f:
```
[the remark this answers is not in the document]
Ehco1996: use python here because i can't use bash to get last two logs i need, 
i don't know why `head` cmd seems not working  in test scripts but work in my local shell
I spent 4 hours debugging this problem and was frustrated that it didn't work out in the end
so why not use a high-level language to deal with this?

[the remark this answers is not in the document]
lance6716: it's OK to use python since we have `check_point`.

maybe that's caused by `head` has many distributions, like BSD/GNU

Makefile
```diff
@@ -81,6 +81,7 @@ dm-portal-frontend:
 	tools/bin/statik -src=./dm/portal/frontend/build -dest=./dm/portal/
 
 tools_setup:
+	@chmod +x tests/_utils/*.py
```
lance6716, line 84: do this then `git add` again 😅

syncer/syncer.go
```diff
@@ -203,6 +203,7 @@ type Syncer struct {
 	tsOffset            atomic.Int64             // time offset between upstream and syncer, DM's timestamp - MySQL's timestamp
 	secondsBehindMaster atomic.Int64             // current task delay second behind upstream
 	workerLagMap        map[string]*atomic.Int64 // worker's sync lag key:WorkerLagKey val: lag
+	workerLagMapIniting atomic.Bool              // whether workerLagMap is in initing
```
lance6716, line 206: I still don't know why we need this

syncer/syncer.go
```diff
@@ -1403,14 +1407,16 @@ func (s *Syncer) Run(ctx context.Context) (err error) {
...
 		name := queueBucketName(i)
 		s.queueBucketMapping = append(s.queueBucketMapping, name)
-		s.workerLagMap[dmlWorkerLagKey(i)] = atomic.NewInt64(0)
+		workerLagKey := dmlWorkerLagKey(i)
+		s.workerLagMap[workerLagKey] = atomic.NewInt64(0)
 		go func(i int, name string) {
 			ctx2, cancel := context.WithCancel(ctx)
 			ctctx := s.tctx.WithContext(ctx2)
-			s.syncDML(ctctx, name, s.toDBConns[i], s.jobs[i], dmlWorkerLagKey(i))
+			s.syncDML(ctctx, name, s.toDBConns[i], s.jobs[i], workerLagKey)
```
lance6716, line 1414: seems a wrong usage of closure, pass `workerLagKey` like `i, name`

syncer/syncer.go
lance6716: https://www.google.com/search?q=use+slice+rather+than+map&oq=use+slice+rather+than+map&aqs=chrome..69i57j69i64.3091j0j7&sourceid=chrome&ie=UTF-8

syncer/syncer.go
Ehco1996: got it...... we do not need workerLagMapIniting

Ehco1996:
/hold 
need fix test data race

tests/metrics/data/db2.prepare.sql
```diff
@@ -0,0 +1,19 @@
...
+    ts timestamp,
+    PRIMARY KEY (id)
+);
+
+insert into
+    t2 (name, ts)
+values
+    ('Arya', now()),
+    ('Bran', '2021-05-11 10:01:05'),
+    ('Sansa', NULL);
```
lichunzhu, line 19: Just asking.. Why do you write sql in this format 🤔

tests/metrics/run.sh
```diff
@@ -0,0 +1,116 @@
...
+
+	# check the time interval between dm-worker updates when the queue is empty, the desired time interval is 1s
+	kill_dm_worker
+	export GO_FAILPOINTS="github.com/pingcap/dm/syncer/changeTickerInterval=return(1)"
+	run_dm_worker $WORK_DIR/worker1 $WORKER1_PORT $cur/conf/dm-worker1.toml
+	check_rpc_alive $cur/../bin/check_worker_online 127.0.0.1:$WORKER1_PORT
+	run_dm_worker $WORK_DIR/worker2 $WORKER2_PORT $cur/conf/dm-worker2.toml
+	check_rpc_alive $cur/../bin/check_worker_online 127.0.0.1:$WORKER2_PORT
+	sleep 3 # we need sleep 3 because current ticker interval is 1 and we need at least 2 ticker's log
+	$cur/../_utils/check_ticker_interval.py $WORK_DIR/worker1/log/dm-worker.log 1
```
lichunzhu, line 100: Why do we check interval here?

lichunzhu (commented):
Rest LGTM

tests/metrics/run.sh
Ehco1996: to make sure ticker Reset is working properly
---
resets the time interval for each loop to prevent a certain amount of time being spent on the previous ticker execution to `executeSQLs` resulting in the next tikcer not waiting for the full waitTime.

tests/metrics/data/db2.prepare.sql
Ehco1996: oh, it's formated by sql formatting tools

tests/metrics/run.sh
Ehco1996: when not set changeTickerInterval, lag will update to zero even if current ddl lag is large than zero 😅

so i need lag metrics not updated so quick

tests/metrics/run.sh
lichunzhu: But I don't think this test can make sure "the next ticker not waiting for the full waitTime". If you set `changeTickerInterval` to 1 this value will definitely be larger than 1. What's more, there is no new job except `flushJob` from `jobChan`, so this also can't satisfy your expectations.

lance6716: > But I don't think this test can make sure "the next ticker not waiting for the full waitTime". If you set `changeTickerInterval` to 1 this value will definitely be larger than 1. What's more, there is no new job except `flushJob` from `jobChan`, so this also can't satisfy your expectations.

originally there's a bug 😂 so not "...definitely be larger than 1"

lichunzhu: I know that bug... I mean this test can't make us confirm that bug is fixed.

lichunzhu (approved):
LGTM

tests/metrics/run.sh
```diff
@@ -0,0 +1,125 @@
...
+		"\"secondsBehindMaster\": \"0\"" 2
+	echo "check zero job done!"
+
+	kill_dm_worker
+	export GO_FAILPOINTS="github.com/pingcap/dm/syncer/changeTickerInterval=return(5)"
+	# First set the ticker interval to 5s -> expect the execSQL interval to be greater than 5s
+	# At 5s, the first no job log will appear in the log
+	# At 6s, the ticker has already waited 1s and the ticker goes to 1/5th of the way
+	# At 6s, a dml job is added to jobchan and the ticker is reset
+	# At 11s the ticker  wirte the log of the second nojob
```
lance6716, line 98: fix typo

tests/metrics/run.sh
```diff
@@ -0,0 +1,125 @@
...
+	run_dm_worker $WORK_DIR/worker1 $WORKER1_PORT $cur/conf/dm-worker1.toml
+	check_rpc_alive $cur/../bin/check_worker_online 127.0.0.1:$WORKER1_PORT
+	run_dm_worker $WORK_DIR/worker2 $WORKER2_PORT $cur/conf/dm-worker2.toml
+	check_rpc_alive $cur/../bin/check_worker_online 127.0.0.1:$WORKER2_PORT
+	echo "sleep 6s"
+	sleep 6
+	echo "make a dml job"
+	run_sql_source1 "use metrics;insert into t1 (id, name, ts) values (1004, 'zmj4', '2022-05-11 12:01:05')"
+	echo "sleep 5s"
+	sleep 5
```
lance6716, line 109: maybe sleep slightly longer than 5 seconds, or add some retry, to avoid a unstable test.

lance6716 (commented):
rest lgtm

tests/metrics/run.sh
Ehco1996: current test metrics task Less than 1 minutes，and  `5s` interval basically guarantees that the test will pass

* 1~3 second is  unstable
* 4 second It looks unlucky 😬

lance6716: I'm afraid the worker runs slower so it takes **more than** 5 seconds to log, so we might sleep 6 seconds for example

lance6716 (approved):

PR #1856, opened 2021-07-11
relay: not watch whole dir to reduce cpu usage

### LINKED ISSUE #1827: Improve performance when there's lots of relay log files.
## Feature Request

during a debug, we found `lstat` system call may get slow.

image

there're 5000 relay log files of that worker, maybe we should test if this is the root cause.

the version of DM is v2.0.1

the disk is AWS EBS io2 type.

`time ls > /dev/null` costs 0.03s

We may check if the last relay log file has benn changing first. If not, we turn to list directory then.

note upgrading go version to v1.16 (which has been done in DM v2.0.4) may improve the performance.

Ehco1996:

close #1827 

> relay spend too much cpu beacuse we used to watching whole relay log dir
> system call `os.Lstat` have poor performance when there are too many files in a directory

image

This PR solves the above problem in the following way

* ensure relay meta file updated immediately after reveive `rotate` event
* check the size of the file first, if the size is the same as the latest file, we assume there is no new wirte
* chenk relay meta file to see if the new relay log is created

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes
 -  pkg/reader

Side effects

 - Increased code complexity

Related changes


lance6716:
you has too many open PRs, we'd better finish previous PRs 😂

Ehco1996:
maybe  we can only watch the meta file

Ehco1996:
> maybe we can only watch the meta file

we can not watch meta file, because meta file flushed every 30 seconds

and i thinks now this pr is ready for review  @lance6716  @lichunzhu PTAL
When a single binlog file is of normal size, the cpu consumption is already small
Only when rotate occurs will os.list be used to scan the current directory

lance6716:
> because meta file flushed every 30 seconds

we can change the logic to flushing meta immediately when binlog rotates

Ehco1996:
> > because meta file flushed every 30 seconds
> 
> we can change the logic to flushing meta immediately when binlog rotates

We need to know not only that the name of the binlogfile has changed but also that the file size of the binlog file has changed

I don't think it's a good idea to flush the meta every time we get an event

lance6716:
> We need to know not only that the name of the binlogfile has changed but also that the file size of the binlog file has changed

if the relay log is not rotated, we can watch the relay log file size

> I don't think it's a good idea to flush the meta every time we get an event

I didn't say every event, only rotate event

pkg/streamer/file.go
```diff
@@ -194,128 +195,61 @@ func fileSizeUpdated(path string, latestSize int64) (int, error) {
 	}
 }
 
-// relaySubDirUpdated checks whether the relay sub directory updated
-// including file changed, created, removed, etc.
-func relaySubDirUpdated(ctx context.Context, watcherInterval time.Duration, dir string,
+// relayLogUpdatedOrNewCreated checks whether current relay log file is updated or new relay log is created.
+// we check the size of the file first, if the size is the same as the latest file, we assume there is no new wirte
```
lance6716, line 199:
```suggestion
// we check the size of the file first, if the size is the same as the latest file, we assume there is no new write
```

you may need a plugin to detect typo, so many so far. you can try find rest of typos in this PR on your own before our next review 😛

pkg/streamer/file.go
```diff
@@ -194,128 +195,61 @@ func fileSizeUpdated(path string, latestSize int64) (int, error) {
...
-				result <- watchResult{
-					updatePath: "",
-					err:        newCtx.Err(),
-				}
+	ticker := time.NewTicker(watcherInterval)
+	defer ticker.Stop()
+	for {
+		select {
+		case <-ctx.Done():
+			if err := ctx.Err(); err != nil {
```
lance6716, line 208: If Done is closed, Err returns a non-nil error explaining why

pkg/streamer/file.go
```diff
@@ -194,128 +195,61 @@ func fileSizeUpdated(path string, latestSize int64) (int, error) {
...
-					continue
-				} else if !event.HasOps(watcher.Modify, watcher.Create) {
-					log.L().Debug("skip uninterested event", zap.Stringer("operation", event.Op), zap.String("path", event.Path))
-					continue
-				}
-				baseName := filepath.Base(event.Path)
-				if !binlog.VerifyFilename(baseName) {
-					log.L().Debug("skip watcher event for invalid relay log file", zap.Reflect("event", event))
-					continue // not valid binlog created, updated
+				defer metaFile.Close()
```
lance6716, line 235: I think lint should report this "defer in for-loop" 🤔 We should check why lint doesn't work first

pkg/streamer/file.go
```diff
@@ -194,128 +195,61 @@ func fileSizeUpdated(path string, latestSize int64) (int, error) {
...
 				}
-				result <- watchResult{
-					updatePath: event.Path,
-					err:        nil,
+				if meta.BinLogName != latestFile {
+					nextFilePath := filepath.Join(dir, meta.BinLogName)
+					log.L().Info("newer relay log file is already generated, start parse from it",
+						zap.String("now file", latestFilePath),
+						zap.String("new file", nextFilePath))
+					updatePathCh <- nextFilePath
```
lance6716, line 247: check previous file doesn't have new content

pkg/streamer/file.go
Ehco1996: `Pragmatically speaking golangci-lint is a meta-linter and does not implement any linting or static analysis checks of its own. It rather relies on checks that it gets from other linters (such as govet, staticcheck, ineffassign, etc). As a result your "new check" suggestion would be more actionable if made to another GitHub project. The nature of the suggested check makes it, in my opinion, a prime candidate for go-critic`

coppied from  https://github.com/golangci/golangci-lint/issues/627

pkg/streamer/file.go
Ehco1996: Can you please tell me the reason why I need to do this check

We check the current file size in each for loop first.
and only check the contents of the meta file when the file size is constant

The previous file in the current meta file is not necessarily the current file.
So it might not make sense to check if the size of the previous file has changed?

pkg/streamer/file.go
Ehco1996: Already checked it out, thanks for your patience

pkg/streamer/file.go
Ehco1996: My bad, the code was generated via Copilot, I didn't think it through 😬

pkg/streamer/file.go
lance6716: so we should turn on which linter?

pkg/streamer/file.go
lance6716: after the check, new content may be written and then new relay log file generated

pkg/streamer/file.go
```diff
@@ -194,128 +195,59 @@ func fileSizeUpdated(path string, latestSize int64) (int, error) {
...
-			case <-newCtx.Done():
-				result <- watchResult{
-					updatePath: "",
-					err:        newCtx.Err(),
-				}
+	ticker := time.NewTicker(watcherInterval)
+	defer ticker.Stop()
+	for {
+		select {
+		case <-ctx.Done():
```
lance6716, line 207: please add a comment for this function to express that the function will be block until send something to the two channel. And send the error here

pkg/streamer/file.go
Ehco1996: maybe we can add a new linter ? https://github.com/go-critic/go-critic

pkg/streamer/file.go
```diff
@@ -194,128 +195,59 @@ func fileSizeUpdated(path string, latestSize int64) (int, error) {
...
-				if !ok {
-					result <- watchResult{
-						updatePath: "",
-						err:        terror.ErrWatcherChanClosed.Generate("events", dir),
-					}
+			default:
+				// current watched file size have no change means that no new writes have been made
+				// our relay meta file will be updated immediately after receive the rotate event
+				// although we can't ensure the binlog filename in meta is the next one we expected
+				// but if we return a different filename with latestFile,the outer logic (parseDirAsPossible)
```
lance6716, line 226:
```suggestion
				// but if we return a different filename with latestFile, the outer logic (parseDirAsPossible)
```

pkg/streamer/file.go
```diff
@@ -194,128 +195,59 @@ func fileSizeUpdated(path string, latestSize int64) (int, error) {
...
-			case event, ok := <-watcher2.Events:
-				if !ok {
-					result <- watchResult{
-						updatePath: "",
-						err:        terror.ErrWatcherChanClosed.Generate("events", dir),
-					}
+			default:
+				// current watched file size have no change means that no new writes have been made
+				// our relay meta file will be updated immediately after receive the rotate event
+				// although we can't ensure the binlog filename in meta is the next one we expected
```
lance6716, line 225: what's "we expected"?

pkg/streamer/file.go
```diff
@@ -194,128 +195,59 @@ func fileSizeUpdated(path string, latestSize int64) (int, error) {
...
+				if err != nil {
+					errCh <- terror.Annotate(err, "decode meta toml faild")
+					return
 				}
-				result <- watchResult{
-					updatePath: event.Path,
-					err:        nil,
+				if meta.BinLogName != latestFile {
+					nextFilePath := filepath.Join(dir, meta.BinLogName)
+					log.L().Info("newer relay log file is already generated, start parse from it",
```
lance6716, line 242: in above comment you say outer logic will find the right file to parse

relay/relay.go
```diff
@@ -556,6 +556,19 @@ func (r *Relay) handleEvents(ctx context.Context, reader2 reader.Reader, transfo
 				return terror.Annotatef(err, "save position %s, GTID sets %v into meta", lastPos, lastGTID)
 			}
 		}
+		if _, ok := e.Event.(*replication.RotateEvent); ok {
```
lance6716, line 559: you have used `tResult.NextLogName` below, so no need to check the type here

pkg/streamer/file.go
```diff
@@ -199,128 +199,72 @@ func fileSizeUpdated(path string, latestSize int64) (int, error) {
...
-				} else if !event.HasOps(watcher.Modify, watcher.Create) {
-					log.L().Debug("skip uninterested event", zap.Stringer("operation", event.Op), zap.String("path", event.Path))
-					continue
-				}
-				baseName := filepath.Base(event.Path)
-				if !binlog.VerifyFilename(baseName) {
-					log.L().Debug("skip watcher event for invalid relay log file", zap.Reflect("event", event))
-					continue // not valid binlog created, updated
+				meta := &Meta{}
+				_, err = toml.DecodeReader(metaFile, meta)
```
lance6716, line 239: toml.DecodeFile
learn the API of package before you use it

pkg/streamer/file.go
```diff
@@ -199,128 +199,72 @@ func fileSizeUpdated(path string, latestSize int64) (int, error) {
...
+					return
 				}
-				result <- watchResult{
-					updatePath: event.Path,
-					err:        nil,
+				if meta.BinLogName != latestFile {
+					// we need recheck file size again, as the file may have been changed during our metafile check.
+					cmp, err := fileSizeUpdated(latestFilePath, latestFileSize)
+					if err != nil {
+						errCh <- terror.ErrRelayLogFileSizeSmaller.Generate(latestFilePath)
```
lance6716, line 249: use proper error

pkg/streamer/file_test.go
```diff
@@ -295,14 +301,14 @@ func (t *testFileSuite) TestFileSizeUpdated(c *C) {
...
-func (t *testUtilSuite) TestRelaySubDirUpdated(c *C) {
+func (t *testFileSuite) TestrelayLogUpdatedOrNewCreated(c *C) {
 	var (
 		relayFiles = []string{
 			"mysql-bin.000001",
 			"mysql-bin.000002",
-			"mysql-bin.000003",
-			"mysql-bin.000004",
 		}
+		binlogPos       = uint32(1)
```
lance6716, line 310:
```suggestion
		binlogPos       = uint32(4)
```

pkg/streamer/file_test.go
```diff
@@ -367,96 +381,37 @@ func (t *testUtilSuite) TestRelaySubDirUpdated(c *C) {
...
+
+		relayLogUpdatedOrNewCreated(ctx, watcherInterval, subDir, relayPaths[0], relayFiles[0], 0, updatePathCh, errCh)
 		c.Assert(len(errCh), Equals, 0)
 		c.Assert(len(updatePathCh), Equals, 1)
 		up := <-updatePathCh
 		c.Assert(up, Equals, relayPaths[1])
 	}()
 	wg.Wait()
-
-	// 4. file updated
```
lance6716, line 392: why do you delete tests again?

pkg/streamer/reader_test.go
```diff
@@ -1284,3 +1285,13 @@ func (t *testReaderSuite) writeUUIDs(c *C, relayDir string, uuids []string) []by
 	c.Assert(err, IsNil)
 	return buf.Bytes()
 }
+
+func (t *testReaderSuite) createMetaFile(c *C, binlogFileName, gtidSet, uuidDirPath string, pos uint32) {
+	println("create file in ", path.Join(uuidDirPath, utils.MetaFilename))
```
lance6716, line 1290: remove debug message

pkg/streamer/reader_test.go
```diff
@@ -408,6 +411,7 @@ func (t *testReaderSuite) TestParseFileRelayNeedSwitchSubDir(c *C) {
 	// has relay log file in next sub directory, need to switch
 	ctx2, cancel2 := context.WithTimeout(context.Background(), parseFileTimeout)
 	defer cancel2()
+	t.createMetaFile(c, filename, "", relayDir, uint32(offset))
```
lance6716, line 414: please check if `createMetaFile` always use `offset` and `filename`

pkg/streamer/reader_test.go
```diff
@@ -553,6 +558,8 @@ func (t *testReaderSuite) TestStartSyncByPos(c *C) {
 			filename := filepath.Join(baseDir, UUIDs[i], filenamePrefix+strconv.Itoa(j))
 			err = ioutil.WriteFile(filename, eventsBuf.Bytes(), 0o600)
 			c.Assert(err, IsNil)
+			// create meta file
+			t.createMetaFile(c, filenamePrefix+strconv.Itoa(j), "", path.Join(baseDir, UUIDs[i]), startPos.Pos)
```
lance6716, line 562: why inside the loop? it will overwrite previous file anyway

pkg/streamer/reader_test.go
```diff
@@ -593,6 +599,8 @@ func (t *testReaderSuite) TestStartSyncByPos(c *C) {
 		c.Assert(err, IsNil)
 	}
 
+	t.createMetaFile(c, filenamePrefix, "", path.Join(baseDir, UUIDs[2]), lastPos)
```
lance6716, line 602: wrong filename

pkg/streamer/reader_test.go
```diff
@@ -1284,3 +1285,13 @@ func (t *testReaderSuite) writeUUIDs(c *C, relayDir string, uuids []string) []by
 	c.Assert(err, IsNil)
 	return buf.Bytes()
 }
+
+func (t *testReaderSuite) createMetaFile(c *C, binlogFileName, gtidSet, uuidDirPath string, pos uint32) {
```
lance6716, line 1289: reorder parameters to c, binlog location informations, path

relay/relay.go
```diff
@@ -556,6 +556,17 @@ func (r *Relay) handleEvents(ctx context.Context, reader2 reader.Reader, transfo
...
 		}
+		if tResult.NextLogName != "" && !utils.IsFakeRotateEvent(e.Header) {
+			// if the binlog is rotated, we need to save and flush the next binlog filename to meta
+			lastPos.Name = tResult.NextLogName
+			err := r.SaveMeta(lastPos, lastGTID.Clone())
+			if err != nil {
+				return terror.Annotatef(err, "save position %s, GTID sets %v into meta", lastPos, lastGTID)
+			}
+		}
+		if err2 := r.FlushMeta(); err2 != nil {
```
lance6716, line 567: here will flush for every events?

relay/relay_test.go
```diff
@@ -443,6 +443,13 @@ func (t *testRelaySuite) TestHandleEvent(c *C) {
 	c.Assert(pos, DeepEquals, binlogPos)
 	c.Assert(gs.String(), Equals, "") // no GTID sets in event yet
 
+	// after handle rotate envent, we save and flush the meta immediately
+	newctx, newcancel := context.WithTimeout(context.Background(), 10*time.Millisecond)
```
lance6716, line 447: newCtx

pkg/streamer/file.go
Ehco1996: could you please tell me what does this phrase mean? `send the error here`

pkg/streamer/file_test.go
Ehco1996: added new test case in https://github.com/pingcap/dm/pull/1856/commits/ae3415f7e887f076bc45c9255d7ebd317a9d5a5a

relay/relay_test.go
```diff
@@ -443,6 +443,13 @@ func (t *testRelaySuite) TestHandleEvent(c *C) {
 	c.Assert(pos, DeepEquals, binlogPos)
 	c.Assert(gs.String(), Equals, "") // no GTID sets in event yet
 
+	// after handle rotate envent, we save and flush the meta immediately
+	newCtx, newcancel := context.WithTimeout(context.Background(), 10*time.Millisecond)
```
lance6716, line 447: newCancel. please install a plugin to find naming problems

pkg/streamer/file_test.go
```diff
@@ -367,96 +378,83 @@ func (t *testUtilSuite) TestRelaySubDirUpdated(c *C) {
...
-		c.Assert(up, Equals, relayPaths[2])
-	}()
-
-	// wait watcher started, create a new file
-	time.Sleep(2 * watcherInterval)
-	err = ioutil.WriteFile(relayPaths[2], data, 0o600)
-	c.Assert(err, IsNil)
 	wg.Wait()
 
-	// 6.
```
lance6716, line 426: why remove these tests?

Ehco1996: Because the logic of this function has been changed

The test section for file created on line 409 is already above (line 389)

The logic for listening for directory-related events on line 426 is also gone

Ehco1996 (commented):
self review for typo

relay/relay_test.go
Ehco1996: Feels much better with this plugin: `Code Spell Checker`

pkg/streamer/file.go
```diff
@@ -186,141 +187,90 @@ func fileSizeUpdated(path string, latestSize int64) (int, error) {
...
 	}
 }
 
-// relaySubDirUpdated checks whether the relay sub directory updated
-// including file changed, created, removed, etc.
-func relaySubDirUpdated(ctx context.Context, watcherInterval time.Duration, dir string,
+// relayLogUpdatedOrNewCreated checks whether current relay log file is updated or new relay log is created.
+// we check the size of the file first, if the size is the same as the latest file, we assume there is no new write
+// so we need to check relay meta file to see if the new relay log is created.
+// this func will be blocked until current filesize changed or meta file updated or context cancelled.
```
GMHDBJD, line 208: Should we watch the `server-uuid.index` file? For master-slave swith

Ehco1996: I think the function `needSwitchSubDir` is used to handle master-slave switching? there is another goroutine in `parseFile` that already does this

relay/relay_test.go
```diff
@@ -443,6 +443,13 @@ func (t *testRelaySuite) TestHandleEvent(c *C) {
 	c.Assert(pos, DeepEquals, binlogPos)
 	c.Assert(gs.String(), Equals, "") // no GTID sets in event yet
 
+	// after handle rotate event, we save and flush the meta immediately
```
lance6716, line 446: if you want to test "after handle rotate event, we save and flush the meta immediately", don't start another handleEvents because previous one handled rotate event

lance6716:
rest lgtm

pkg/streamer/file.go
```diff
@@ -186,141 +187,90 @@ func fileSizeUpdated(path string, latestSize int64) (int, error) {
...
-	defer cancel()
-	go func() {
-		for {
-			select {
-			case <-newCtx.Done():
-				result <- watchResult{
-					updatePath: "",
-					err:        newCtx.Err(),
-				}
+	ticker := time.NewTicker(watcherInterval)
```
lichunzhu, line 211: This ticker is used only once? Maybe `time.After` is enough?

pkg/streamer/file.go
```diff
@@ -186,141 +187,90 @@ func fileSizeUpdated(path string, latestSize int64) (int, error) {
...
-						updatePath: "",
-						err:        terror.ErrWatcherChanRecvError.Delegate(err2, dir),
-					}
-				}
+			}
+			failpoint.Inject("CMPAlwaysReturn0", func(_ failpoint.Value) {
+				cmp = 0
+			})
+			switch {
+			case cmp < 0:
```
lichunzhu, line 229: What if we clean some relay logs?

pkg/streamer/file.go
Ehco1996: when the binlog file size does not change at first and the relay meta does not move forward
will rely on the ticker to cycle check the binlogfile file size

pkg/streamer/file.go
Ehco1996: `fileSizeUpdated` will return a error and task will pause

pkg/streamer/file.go
lichunzhu: Sorry. I take this look mistakenly.

pkg/streamer/file.go
lichunzhu: What about `relayPurger`? Will this cause the task to pause?

Ehco1996: relayPurger does not affect the task 

because it does not clear the protected relaylog， see more in `setActiveRelayLog`

pkg/streamer/file_test.go
```diff
@@ -295,14 +297,15 @@ func (t *testFileSuite) TestFileSizeUpdated(c *C) {
...
+func (t *testFileSuite) TestrelayLogUpdatedOrNewCreated(c *C) {
 	var (
 		relayFiles = []string{
 			"mysql-bin.000001",
 			"mysql-bin.000002",
 			"mysql-bin.000003",
-			"mysql-bin.000004",
 		}
+		binlogPos       = uint32(4)
+		binlogGTID      = "ba8f633f-1f15-11eb-b1c7-0242ac110002.000001"
```
lichunzhu, line 308: This doesn't look like a GTID?
```suggestion
		binlogGTID      = "ba8f633f-1f15-11eb-b1c7-0242ac110002-1"
```

lichunzhu (commented):
Rest LGTM

lichunzhu (approved):
LGTM

pkg/streamer/file_test.go
```diff
@@ -367,96 +378,83 @@ func (t *testUtilSuite) TestRelaySubDirUpdated(c *C) {
...
-	// err = os.Rename(relayPaths[0], relayPaths[3])
-	// c.Assert(err, IsNil)
-
-	// wait again, update the file
-	time.Sleep(2 * watcherInterval)
-	err = ioutil.WriteFile(relayPaths[2], data, 0o600)
-	c.Assert(err, IsNil)
-	wg.Wait()
+	// 5. context timeout (no new write)
+	// revert meta file to old file
```
lance6716, line 442: why bother to revert to old file? does not do this causing a problem?

lance6716 (commented):
rest lgtm

pkg/streamer/file_test.go
Ehco1996: Because the latest binlog file is currently binlog.002
But test `#4` above deliberately wrote the error binlog.003 into the meta 

The current test is to check that the error was returned as expected when the ctx was cancelled, so need to change the binlog file name in the wrong meta back to binlog.0002

pkg/streamer/file.go
```diff
@@ -186,141 +187,86 @@ func fileSizeUpdated(path string, latestSize int64) (int, error) {
...
-				}
-				result <- watchResult{
-					updatePath: event.Path,
-					err:        nil,
+				if meta.BinLogName != latestFile {
+					// we need check file size again, as the file may have been changed during our metafile check
+					cmp, err := fileSizeUpdated(latestFilePath, latestFileSize)
+					if err != nil {
+						errCh <- terror.Annotatef(err, "latestFilePath=%s latestFileSize=%d",
+							latestFilePath, latestFileSize)
```
lance6716, line 251: return here. And i'm requesting unit tests that "if relayLogUpdatedOrNewCreated returns, len(updatePathCh) + len(errCh) should be equal to 1"

Ehco1996: add in https://github.com/pingcap/dm/pull/1856/commits/1e4d300963b62604a66ecfc4e0d6aa9c5be2ef72

lance6716: RETURN HERE

Ehco1996: mybad... you are so attentive

lance6716 (approved):

Issue #1735, opened 2021-05-31
SQL mode is still not automatic enough


lance6716:
## Feature Request

**Is your feature request related to a problem? Please describe:**

Currently DM will query global SQL Mode of upstream to sync. When user set a session level one, the inconsistent of SQL mode may cause an error

**Describe the feature you'd like:**

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

together-wang:
sql_mode automatic matching error

NO_ZERO_DATE exists in upstream sql_mode, and downstream sql_mode is empty
When the old table with zero_date in the upstream is synchronized through DM, the error Invalid default value is reported, and it is not automatically matched to the downstream sql_mode

![image](https://user-images.githubusercontent.com/44355590/120608811-2dea7780-c484-11eb-97b6-a01f04e9a817.png)
![image](https://user-images.githubusercontent.com/44355590/120608854-38a50c80-c484-11eb-9417-c21f78f53d03.png)
![image](https://user-images.githubusercontent.com/44355590/120609013-5ffbd980-c484-11eb-84ef-4571f1d20778.png)
![image](https://user-images.githubusercontent.com/44355590/120609322-a4877500-c484-11eb-9ae5-d1e68be2074c.png)
![image](https://user-images.githubusercontent.com/44355590/120609486-caad1500-c484-11eb-9396-1c912af976d6.png)

Issue #1818, opened 2021-06-25
make task config loader.SQL_MODE configurable


lichunzhu:
## Feature Request

**Is your feature request related to a problem? Please describe:**

If MySQL has some data dumped in session sql_mode, using current global sql_mode to adjust loader's SQL_MODE might be ineffective.

**Describe the feature you'd like:**

make task config loader.SQL_MODE configurable

**Describe alternatives you've considered:**

Change downstream TiDB's global sql_mode.

**Teachability, Documentation, Adoption, Migration Strategy:**

lance6716:
In what case there's "MySQL has some data dumped in session sql_mode", does dumpling has this configuration 🤔

lichunzhu:
> In what case there's "MySQL has some data dumped in session sql_mode", does dumpling have this configuration 🤔

For example, I load some data without "NO_ZERO_IN_DATE" in sql_mode. But the global sql_mode might be "NO_ZERO_IN_DATE".
Loader will fail because of this.

lance6716:
related to https://github.com/pingcap/dm/issues/1735

I think we could always mask some SQL mode, to provide better compatibility.

lichunzhu:
> related to #1735
> 
> I think we could always mask some SQL mode, to provide better compatibility.

SGTM

kolbe:
@lance6716 what do you mean by "always mask some SQL mode"?

lance6716:
> @lance6716 what do you mean by "always mask some SQL mode"?

A rough thought, we should remove some SQL mode such as NO_ZERO_IN_DATE when synchronize to downstream, and keep rest of them to handle changes of content made by ANSI_QUOTES, for example.

kolbe:
I disagree. I think the behavior of having the DM worker match the upstream SQL mode is 100% correct, but there should be an option to override that if the data requires it.

lance6716:
> I disagree. I think the behavior of having the DM worker match the upstream SQL mode is 100% correct, but there should be an option to override that if the data requires it.

ok we shall left an option to deal with automatic logic going wrong. in the meantime, it's better to improve automatic logic. I guess if user meets some errors he/she still has no idea how to correctly set a SQL mode.

kolbe:
The error number returned by TiDB Server can be interpreted by DM to give better information about the resolution.

Another improvement to DM could be to simply display error output in a more friendly manner. Instead of showing raw JSON with embedded newlines, maybe we can show it in a more friendly, human-readable way?

[...]

PR #1869, opened 2021-07-14
loader, syncer: Update sql_mode to more compatible

okJiang:

close https://github.com/pingcap/dm/issues/1735 and https://github.com/pingcap/dm/issues/1818

By add/delete some upstream's global.sql_mode to make downstream's sql_mode more compatible.

### Check List 

Tests 

 - Integration test

Code changes

Side effects

Related changes


loader/loader.go
```diff
@@ -529,7 +529,44 @@ func (l *Loader) Init(ctx context.Context) (err error) {
...
+		}
+		waitEnable := sqlModes
+		for _, enable := range needEnable {
+			for _, sqlMode := range sqlModes {
+				if enable != sqlMode {
+					waitEnable = append(waitEnable, enable)
+					break
+				}
+			}
+		}
```
GMHDBJD, line 567: How about extracting the same logic to a function. We can put it in pkg/utils

syncer/syncer.go
```diff
@@ -2742,8 +2742,43 @@ func (s *Syncer) createDBs(ctx context.Context) error {
...
+				"NO_ZERO_DATE",
+				"ERROR_FOR_DIVISION_BY_ZERO",
+				"NO_AUTO_CREATE_USER",
+				"STRICT_TRANS_TABLES",
+				"STRICT_ALL_TABLES",
+			}
+			for _, disable := range needDisable {
+				for i, sqlMode := range sqlModes {
+					if disable == sqlMode {
+						sqlModes = append(sqlModes[:i], sqlModes[i+1:]...)
```
lance6716, line 2762: maybe parser has an elegant way to check and modify SQL modes

tests/sql_mode/run.sh
```diff
@@ -0,0 +1,50 @@
+#!/bin/bash
```
GMHDBJD, line 1: New test case should be added in other_intergration.txt

lance6716 (commented):
also notice there're two usage of SQL mode: one to execute SQLs and one to parse upstream DDLs

syncer/syncer.go
okJiang: I have thought about this problem, but I did not think of a better way. How about replacing and adding string directly?

```go
for _, disable := range needDisable {
	strings.RelpaceAll(l.cfg.LoaderConfig.SQLMode, disable, "")
}
for _, enable := range needEnable {
	l.cfg.LoaderConfig.SQLMode += "," + enable
}
```

loader/loader.go
okJiang: No problem. Do we need to do this whenever there are two duplicate logics?

tests/sql_mode/run.sh
okJiang: done

loader/loader.go
```diff
@@ -529,7 +529,44 @@ func (l *Loader) Init(ctx context.Context) (err error) {
 		}
 	}
 	if !hasSQLMode {
-		lcfg.To.Session["sql_mode"] = l.cfg.LoaderConfig.SQLMode
+		// When upstream's datatime is 2020-00-00, 2020-00-01, 2020-06-00
```
glorv, line 532: Do we still need to support manually set sql_mode here?

syncer/syncer.go
lance6716: maybe https://github.com/pingcap/dm/blob/8089f1a3ba31f4a73e64e3346571e4ac46c5e6d8/pkg/utils/db.go#L429

okJiang: I have thought about this problem but did not think of a better way. How about replace and add string directly like this:
```go
for _, disable := range needDisable {
	l.cfg.LoaderConfig.SQLMode = strings.ReplaceAll(l.cfg.LoaderConfig.SQLMode, disable, "")
}
for _, enable := range needEnable {
	l.cfg.LoaderConfig.SQLMode += "," + enable
}
```

tests/sql_mode/data/db2.increment.sql
```diff
@@ -0,0 +1,19 @@
+set @@session.SQL_MODE='PIPES_AS_CONCAT,IGNORE_SPACE,ONLY_FULL_GROUP_BY,NO_UNSIGNED_SUBTRACTION,NO_DIR_IN_CREATE,NO_AUTO_VALUE_ON_ZERO,NO_BACKSLASH_ESCAPES,STRICT_TRANS_TABLES,STRICT_ALL_TABLES,NO_ZERO_IN_DATE,NO_ZERO_DATE,ALLOW_INVALID_DATES,ERROR_FOR_DIVISION_BY_ZERO,HIGH_NOT_PRECEDENCE,NO_ENGINE_SUBSTITUTION,REAL_AS_FLOAT';
```
glorv, line 1: I prefer to set target sql_mode before each insert statement, this way is more like whay user do.

tests/sql_mode/data/db2.prepare.sql
```diff
@@ -0,0 +1,10 @@
+drop database if exists `sql_mode`;
```
glorv, line 1: We should also prepare some sql_mode specified data so we can test the batch load phase don't impact by different sql_mode

loader/loader.go
okJiang: What do you mean? I do not understand.

syncer/syncer.go
okJiang: I just find Str2SQLMode https://github.com/pingcap/parser/blob/master/mysql/const.go#L469. Should I add SQLMode2Str func?

lance6716: OK, I think parser will reject adding this function, we can add SQLMode2Str in DM

okJiang: So, should I open a new pr to add this function. I think I also should add a unit test for it?

pkg/utils/db.go
```diff
@@ -524,3 +524,85 @@ func AddGSetWithPurged(ctx context.Context, gset gtid.Set, conn *sql.Conn) (gtid
...
+	if err3 != nil {
+		return "", err3
+	}
+	mode = (mode &^ disableMode) | enableMode
+
+	return GetSQLModeByStr(mode), nil
+}
+
+// SQLMode2Str is the sql_mode to string represent of sql_mode map.
+var SQLMode2Str = map[tmysql.SQLMode]string{
```
lance6716, line 564: could use `init()` and iterate through parser.Str2SQLMode to build a reverse map

glorv:
@okJiang  Seems after we remove `STRICT_TRANS_TABLES` from sql_mode, this will cause insert with invalid value to succeed instead of failed, which is expected. So  https://github.com/pingcap/dm/blob/master/tests/handle_error/run.sh#L20-L38 this test will failed.

But In the original test case, the insert fails because upstream change column type but downstream (tidb) don't support this, so after skip ddl, the insert should fail bacuase data in not fit in downstream anymore.

@GMHDBJD @lance6716  @kennytm PTAL at this case, I think we should still keep the `STRICT_TRANS_TABLES` and all user to manually set sql_mode e.g. to just remove `STRICT_TRANS_TABLES` if needed.

pkg/utils/db.go
```diff
@@ -524,3 +524,86 @@ func AddGSetWithPurged(ctx context.Context, gset gtid.Set, conn *sql.Conn) (gtid
...
+	}
+	disableMode, err2 := tmysql.GetSQLMode(disable)
+	if err2 != nil {
+		return "", err2
+	}
+	enableMode, err3 := tmysql.GetSQLMode(enable)
+	if err3 != nil {
+		return "", err3
+	}
+	mode = (mode &^ disableMode) | enableMode
```
Ehco1996, line 558: could you please explain the result after this bit manipulation?

tests/sql_mode/run.sh
```diff
@@ -0,0 +1,50 @@
...
+
+set -eu
+
+cur=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
+source $cur/../_utils/test_prepare
+WORK_DIR=$TEST_DIR/$TEST_NAME
+
+function run() {
+	run_sql_source1 "SET @@GLOBAL.SQL_MODE='PIPES_AS_CONCAT,IGNORE_SPACE,ONLY_FULL_GROUP_BY,NO_UNSIGNED_SUBTRACTION,NO_DIR_IN_CREATE,NO_AUTO_VALUE_ON_ZERO,NO_BACKSLASH_ESCAPES,STRICT_TRANS_TABLES,STRICT_ALL_TABLES,NO_ZERO_IN_DATE,NO_ZERO_DATE,ALLOW_INVALID_DATES,ERROR_FOR_DIVISION_BY_ZERO,NO_AUTO_CREATE_USER,HIGH_NOT_PRECEDENCE,NO_ENGINE_SUBSTITUTION,REAL_AS_FLOAT'"
+	# run_sql_source1 "SET @@GLOBAL.SQL_MODE='ANSI_QUOTES'"
```
Ehco1996, line 11: maybe remove this line?

pkg/utils/db.go
okJiang: Here, disableMode contains all sqlMode bits that need to be deleted, and enableMode contains all sqlMode bits that need to be added. Then use the upstream's mode NAND disableMode, and then OR with enableMode to get the result we want. For example
```txt
 (0110 &^ 0011) | 1000 
= 0100 | 1000
= 1100
```

Ehco1996: maybe add a comment to code

glorv:
> Integration test `handle_error` failed by:
> 
> ```
> [2021-07-14T10:22:13.455Z] got=0 expected=2
> 
> [2021-07-14T10:22:13.456Z] command: query-status test Error .*: Incorrect int value count: 0 != expected: 2, failed the 9-th time, will retry again
> ```
> 
> @okJiang Seems after we remove `STRICT_TRANS_TABLES` from sql_mode, this will cause insert with invalid value to succeed instead of failed, which is expected. So https://github.com/pingcap/dm/blob/master/tests/handle_error/run.sh#L20-L38 this test will failed.
> 
> But In the original test case, the insert fails because upstream change column type but downstream (tidb) don't support this, so after skip ddl, the insert should fail bacuase data in not fit in downstream anymore.
> 
> @GMHDBJD @lance6716 @kennytm PTAL at this case, I think we should still keep the `STRICT_TRANS_TABLES` and all user to manually set sql_mode e.g. to just remove `STRICT_TRANS_TABLES` if needed.

I rethought about this, If data have been successfully inserted in the upstream, it must be valid even if the origin value may not be. So I think we should always enable `STRICT_TRANS_TABLES` to avoid ignoring this kind of bad data.

GMHDBJD:
> I rethought about this, If data have been successfully inserted in the upstream, it must be valid even if the origin value may not be. So I think we should always enable `STRICT_TRANS_TABLES` to avoid ignoring this kind of bad data.

@glorv I think we should keep `STRICT_TRANS_TABLES` consistent with upstream, and allow users to specify it. 
If the user removes `STRICT_TRANS_TABLES` in upstream but keeps it downstream, some data may be invalid when replicating to downstream. https://asktug.com/t/topic/63881

okJiang:
> > Integration test `handle_error` failed by:
> > ```
> > [2021-07-14T10:22:13.455Z] got=0 expected=2
> > 
> > [2021-07-14T10:22:13.456Z] command: query-status test Error .*: Incorrect int value count: 0 != expected: 2, failed the 9-th time, will retry again
> > ```
> >   
> > @okJiang Seems after we remove `STRICT_TRANS_TABLES` from sql_mode, this will cause insert with invalid value to succeed instead of failed, which is expected. So [`master`/tests/handle_error/run.sh#L20-L38](https://github.com/pingcap/dm/blob/master/tests/handle_error/run.sh#L20-L38) this test will failed.
> > But In the original test case, the insert fails because upstream change column type but downstream (tidb) don't support this, so after skip ddl, the insert should fail bacuase data in not fit in downstream anymore.
> > @GMHDBJD @lance6716 @kennytm PTAL at this case, I think we should still keep the `STRICT_TRANS_TABLES` and all user to manually set sql_mode e.g. to just remove `STRICT_TRANS_TABLES` if needed.
> 
> I rethought about this, If data have been successfully inserted in the upstream, it must be valid even if the origin value may not be. So I think we should always enable `STRICT_TRANS_TABLES` to avoid ignoring this kind of bad data.

But once always enable `STRICT_TRANS_TABLES`, most of the work we do is meaningless. Because 

> Strict mode affects handling of division by zero, zero dates, and zeros in dates
> https://dev.mysql.com/doc/refman/5.7/en/sql-mode.html#sql-mode-strict

And at the beginning we wanted to solve the problem that user insert invalid datetime https://github.com/pingcap/dm/issues/1735.

lance6716:
not catch up with your discussion, I just want to remind the data maybe inserted in a session SQL mode that is different with global SQL mode

tests/sql_mode/data/db2.prepare.sql
okJiang: In this pr, db1 is the global sql_mode enable and the session sql_mode disable; db2 is the opposite. And the selected test sql_modes are almost all samples that may cause errors. 

> https://pingcap.feishu.cn/docs/doccn61Fi9k5P8x22M5YnL1eNPh

PTAL, and what sql_mode specified data should I prepare?

loader/loader.go
```diff
@@ -528,10 +528,22 @@ func (l *Loader) Init(ctx context.Context) (err error) {
 			break
 		}
 	}
+
 	if !hasSQLMode {
-		lcfg.To.Session["sql_mode"] = l.cfg.LoaderConfig.SQLMode
+		// When upstream's datatime is 2020-00-00, 2020-00-01, 2020-06-00
+		// and so on, downstream will be 2019-11-30, 2019-12-01, 2020-05-31,
```
Ehco1996, line 534: maybe comment L534-L537 should be add in func `AdjustSQLModeCompatible`

Ehco1996: rest LGTM

okJiang:
@glorv @lance6716 @GMHDBJD  PTAL again

loader/loader.go
```diff
@@ -528,10 +528,17 @@ func (l *Loader) Init(ctx context.Context) (err error) {
...
 		}
 	}
+
 	if !hasSQLMode {
-		lcfg.To.Session["sql_mode"] = l.cfg.LoaderConfig.SQLMode
+		sqlModes, err3 := utils.AdjustSQLModeCompatible(l.cfg.LoaderConfig.SQLMode)
+		if err3 != nil {
+			l.logger.Warn("cannot adjust sql_mode compatible", log.ShortError(err3))
+		}
+		lcfg.To.Session["sql_mode"] = sqlModes
```
lichunzhu, line 537: If `err3 != nil`, `lcfg.To.Session["sql_mode"]` will be set to an empty string "". Is that expected?
If so, pls put a comment here.

tests/sql_mode/data/db2.increment.sql
```diff
@@ -0,0 +1,24 @@
+set @@session.sql_mode='ONLY_FULL_GROUP_BY,NO_UNSIGNED_SUBTRACTION,NO_DIR_IN_CREATE,STRICT_TRANS_TABLES,STRICT_ALL_TABLES,NO_ZERO_IN_DATE,NO_ZERO_DATE,ALLOW_INVALID_DATES,ERROR_FOR_DIVISION_BY_ZERO,HIGH_NOT_PRECEDENCE,NO_ENGINE_SUBSTITUTION,REAL_AS_FLOAT';
+-- NO_AUTO_CREATE_USER set failed in mysql8.0
+
+use sql_mode;
+
+-- test sql_mode PIPES_AS_CONCAT
+set @@session.sql_mode=concat(@@session.sql_mode, ',PIPES_AS_CONCAT');
```
glorv, line 7: What is the final effect of the sql_mode since it contains items that are conflict with each other 🤔？ And I think maybe we needn't do this concat

okJiang: The concated sql_mode dose not conflict with each other. The follow is also. So it does not matter whether it is concated ​​or directly assigned.

glorv: ok

loader/loader.go
okJiang: I think it is. Added in [`c50551a`](https://github.com/pingcap/dm/pull/1869/commits/c50551a94c60658f17abd0d18a5082a01d26bd38)

lichunzhu (approved):
LGTM

okJiang:
PTAL @lance6716 @GMHDBJD

loader/loader.go
```diff
@@ -528,10 +528,17 @@ func (l *Loader) Init(ctx context.Context) (err error) {
...
 	if !hasSQLMode {
-		lcfg.To.Session["sql_mode"] = l.cfg.LoaderConfig.SQLMode
+		sqlModes, err3 := utils.AdjustSQLModeCompatible(l.cfg.LoaderConfig.SQLMode)
+		if err3 != nil {
+			l.logger.Warn("cannot adjust sql_mode compatible, the sql_mode will be assigned \"\"", log.ShortError(err3))
+		}
+		lcfg.To.Session["sql_mode"] = sqlModes
 	}
 
+	l.logger.Info("loader's sql_mode is ", zap.String("sqlmode", lcfg.To.Session["sql_mode"]))
```
lance6716, line 540:
```suggestion
	l.logger.Info("loader's sql_mode is", zap.String("sqlmode", lcfg.To.Session["sql_mode"]))
```

pkg/utils/db.go
```diff
@@ -524,3 +524,56 @@ func AddGSetWithPurged(ctx context.Context, gset gtid.Set, conn *sql.Conn) (gtid
...
 	return ret, nil
 }
+
+// AdjustSQLModeCompatible adjust downstream sql mode to compatible.
+// When upstream's datatime is 2020-00-00, 2020-00-01, 2020-06-00
+// and so on, downstream will be 2019-11-30, 2019-12-01, 2020-05-31,
+// as if set the 'NO_ZERO_IN_DATE', 'NO_ZERO_DATE'.
+// This is because the implementation of go-mysql, that you can see
+// https://github.com/go-mysql-org/go-mysql/blob/master/replication/row_event.go#L1063-L1087
+func AdjustSQLModeCompatible(sqlModes string) (string, error) {
```
lance6716, line 534: we might return original string when has error, so the caller could simply use original value if it falied to adjust.

Or let the caller choose original value or empty value to use.

pkg/utils/db.go
```diff
@@ -524,3 +524,56 @@ func AddGSetWithPurged(ctx context.Context, gset gtid.Set, conn *sql.Conn) (gtid
 	_ = ret.Set(newGset)
 	return ret, nil
 }
+
+// AdjustSQLModeCompatible adjust downstream sql mode to compatible.
+// When upstream's datatime is 2020-00-00, 2020-00-01, 2020-06-00
```
lance6716, line 529: may left a TODO to remind us to fix this bug in go-mysql?

syncer/syncer.go
```diff
@@ -2860,10 +2860,13 @@ func (s *Syncer) createDBs(ctx context.Context) error {
 	if !hasSQLMode {
 		sqlMode, err2 := utils.GetGlobalVariable(ctx, s.fromDB.BaseDB.DB, "sql_mode")
 		if err2 != nil {
-			s.tctx.L().Warn("cannot get sql_mode from upstream database", log.ShortError(err2))
-		} else {
-			s.cfg.To.Session["sql_mode"] = sqlMode
+			s.tctx.L().Warn("cannot get sql_mode from upstream database, the sql_mode will be assigned \"\"", log.ShortError(err2))
```
lance6716, line 2863: if here we failed `GetGlobalVariable`, the `sqlMode` will be empty and after `AdjustSQLModeCompatible`, the active sql mode is "IGNORE_SPACE, NO_AUTO_VALUE_ON_ZERO, ALLOW_INVALID_DATES" not "" as printed in log.

tests/sql_mode/data/db1.prepare.sql
```diff
@@ -0,0 +1,42 @@
...
+CREATE TABLE `t_1` (
+  `id` int NOT NULL AUTO_INCREMENT,
+  `name` varchar(60),
+  `num` int,
+  `dt` datetime,
+  PRIMARY KEY (id)
+);
+
+-- test sql_mode PIPES_AS_CONCAT
+insert into t_1(num) values('pipes'||'as'||'concat');
```
lance6716, line 15: why this can be inserted into an `int` field 😂

tests/sql_mode/data/db1.prepare.sql
```diff
@@ -0,0 +1,42 @@
...
+-- test sql_mode ANSI_QUOTES
+insert into t_1(name) values("a");
+
+-- test sql_mode IGNORE_SPACE
+create table count (id int not null, primary key(id));
+
+-- test sql_mode NO_AUTO_VALUE_ON_ZERO
+insert into t_1(id, name) values (10, 'a');
+insert into t_1(id, name) values (0, 'b');
+insert into t_1(id, name) values (0, 'c');
```
lance6716, line 26: since we didn't set sql mode in this SQL file, above two insert are converted to auto increment value, so the binlog is writtern with the auto increment value not zero, and this didn't test DM will add NO_AUTO_VALUE_ON_ZERO with `AdjustSQLModeCompatible`

since db2 (source2) has done this test, maybe we could remove above lines.

tests/sql_mode/data/db1.increment.sql
```diff
@@ -0,0 +1,32 @@
...
+
+-- test sql_mode ANSI_QUOTES
+insert into t_1(name) values("a");
+
+-- test sql_mode IGNORE_SPACE
+create table sum (id int not null, primary key(id));
+
+-- test sql_mode NO_AUTO_VALUE_ON_ZERO
+insert into t_1(id, name) values (30, 'a');
+insert into t_1(id, name) values (0, 'b');
```
lance6716, line 15: ditto

tests/sql_mode/data/db2.increment.sql
```diff
@@ -0,0 +1,24 @@
...
+insert into t_2(name) values("a");
+
+-- test sql_mode IGNORE_SPACE
+set @@session.sql_mode=concat(@@session.sql_mode, ',IGNORE_SPACE');
+insert into t_2(name) values(concat ('ignore', 'space'));
+
+-- test sql_mode NO_AUTO_VALUE_ON_ZERO
+set @@session.sql_mode=concat(@@session.sql_mode, ',NO_AUTO_VALUE_ON_ZERO');
+insert into t_2(id, name) values (20, 'a');
+replace into t_2(id, name) values (0, 'b');
```
lance6716, line 20: maybe we could 

```
replace into t_2(id, name) values (0, 'c');
```

to test DM can correctly handle this "update"

lance6716 (commented):
rest lgtm

tests/sql_mode/data/db1.prepare.sql
okJiang: Because it exists here as a bit operation OR.

tests/sql_mode/data/db1.prepare.sql
okJiang: Yes, I think it just tested what you said above.

tests/sql_mode/data/db2.increment.sql
okJiang: All above is fixed in https://github.com/pingcap/dm/pull/1869/commits/d49933a5c59864e2dbcc310213f25df1aaf21028

PTAL

PR #1746, opened 2021-06-03
dmctl: fix handle err commands

lichunzhu:
### What problem does this PR solve? 
https://github.com/pingcap/dm/issues/1692 part 1

### What is changed and how it works?
Add `binlog`, `shard-ddl-locks`, `binlog-schema`, `config` commands and update relative integration tests.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

 - Has exported function/method change

Related changes

 - Need to cherry-pick to the release branch
 - Need to update the documentation

PR #1877, opened 2021-07-16
downgrade: add import/export-configs command

GMHDBJD:
add import-config/export-config command for downgrade
- export config from etcd
- import config with start-task and operate-source

### Check List 

Tests 

 - Unit test
 - Integration test


dm/ctl/master/import_export_config.go
```diff
@@ -0,0 +1,383 @@
...
+		RunE:  exportCfgsFunc,
+	}
+	cmd.Flags().StringP("dir", "d", "configs", "specify the destinary directory, default is `./configs`")
+	return cmd
+}
+
+// NewImportCfgsCmd creates a importCfg command.
+func NewImportCfgsCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "import-configs [--dir directory]",
```
lichunzhu, line 58: https://github.com/pingcap/dm/pull/1746
How about `config export` and `config import`? This command looks not compatible with the newer design.

dm/config/source_config.go
```diff
@@ -353,3 +353,85 @@ func (c *SourceConfig) check(metaData *toml.MetaData, err error) error {
...
+
+	// omit default values, so we can ignore them for later marshal
+	s.omitDefaultVals()
+
+	return s.Yaml()
+}
+
+// SourceConfigForDowngrade is the base configuration for source in v2.0.
+// This config is used for downgrade(export-configs) from a higher dmctl version.
+// When we add any new config item into SourceConfig, we should update it also.
```
lance6716, line 376: maybe an unit test using reflect to check the number of field?

dm/config/source_config.go
```diff
@@ -353,3 +353,85 @@ func (c *SourceConfig) check(metaData *toml.MetaData, err error) error {
...
+		Tracer:          sourceCfg.Tracer,
+		CaseSensitive:   sourceCfg.CaseSensitive,
+		Filters:         sourceCfg.Filters,
+	}
+}
+
+// omitDefaultVals change default value to empty value for new config item.
+// If any default value for new config item is not empty(0 or false or nil),
+// we should change it to empty.
+func (c *SourceConfigForDowngrade) omitDefaultVals() {
```
lance6716, line 426: should this function be removed?

dm/ctl/master/import_export_config.go
```diff
@@ -0,0 +1,383 @@
...
+	defer cancel()
+	if err := createSources(ctx, sourceCfgs); err != nil {
+		return err
+	}
+	if err := createTasks(ctx, taskCfgs); err != nil {
+		return err
+	}
+	if len(relayWorkers) > 0 {
+		common.PrintLinesf("The original relay workers have been exported to `%s`.", path.Join(dir, relayWorkersFilename))
+		common.PrintLinesf("Currently unsupport recover relay workers. You may need to execute `transfer-source` and `start-relay` command manually.")
```
lance6716, line 124:
```suggestion
		common.PrintLinesf("Currently DM doesn't support recovering relay workers. You may need to execute `transfer-source` and `start-relay` command manually.")
```

maybe

lance6716 (commented):
will review later

dm/config/source_config.go
GMHDBJD: I think we should keep it, for later use.

dm/config/source_config.go
```diff
@@ -353,3 +353,85 @@ func (c *SourceConfig) check(metaData *toml.MetaData, err error) error {
...
+	s := NewSourceConfigForDowngrade(c)
+
+	// encrypt password
+	cipher, err := utils.Encrypt(utils.DecryptOrPlaintext(c.From.Password))
+	if err != nil {
+		return "", err
+	}
+	s.From.Password = cipher
+
+	// omit default values, so we can ignore them for later marshal
```
lance6716, line 368: https://pkg.go.dev/gopkg.in/yaml.v2

```
omitempty    Only include the field if it's not set to the zero
             value for the type or to empty slices or maps.
             Zero valued structs will be omitted if all their public
             fields are zero, unless they implement an IsZero
             method (see the IsZeroer interface type), in which
             case the field will be excluded if IsZero returns true.
```

dm/config/source_config.go
```diff
@@ -353,3 +353,85 @@ func (c *SourceConfig) check(metaData *toml.MetaData, err error) error {
...
+	Tracer          map[string]interface{} `yaml:"tracer"`
+	// any new config item, we mark it omitempty
+	CaseSensitive bool                  `yaml:"case-sensitive,omitempty"`
+	Filters       []*bf.BinlogEventRule `yaml:"filters,omitempty"`
+}
+
+// NewSourceConfigForDowngrade creates a new base config for downgrade.
+func NewSourceConfigForDowngrade(sourceCfg *SourceConfig) *SourceConfigForDowngrade {
+	return &SourceConfigForDowngrade{
+		EnableGTID:      sourceCfg.EnableGTID,
```
lance6716, line 402: maybe we can use reflect to copy these fields. I'm afraid we forget to copy newly added field here

dm/config/task_test.go
```diff
@@ -1011,3 +1012,11 @@ func (t *testConfig) TestExclusiveAndWrongExprFilterFields(c *C) {
 	err = cfg.adjust()
 	c.Assert(terror.ErrConfigExprFilterWrongGrammar.Equal(err), IsTrue)
 }
+
+func (t *testConfig) TestTaskConfigForDowngrade(c *C) {
```
lance6716, line 1016: and ensure other config 🤔

dm/ctl/master/config.go
```diff
@@ -0,0 +1,395 @@
...
+}
+
+// newExportCfgsCmd creates a exportCfg command.
+func newExportCfgsCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "export [--dir directory]",
+		Short: "Export the configurations of sources and tasks.",
+		RunE:  exportCfgsFunc,
+	}
+	cmd.Flags().StringP("dir", "d", "configs", "specify the destinary directory, default is `./configs`")
```
lance6716, line 62:
```suggestion
	cmd.Flags().StringP("dir", "d", "configs", "specify the output directory, default is `./configs`")
```

or destination

dm/ctl/master/config.go
```diff
@@ -0,0 +1,395 @@
...
+	relayWorkers, _, err := ha.GetAllRelayConfig(cli)
+	if err != nil {
+		common.PrintLinesf("can not get relay workers from etcd")
+		return nil, nil, nil, err
+	}
+	return sourceCfgsMap, subTaskCfgsMap, relayWorkers, nil
+}
+
+func createDirectory(dir string) (string, string, error) {
+	if err := os.MkdirAll(dir, 0o755); err != nil {
```
lance6716, line 198: seems no need to create parent directories because we use MkdirAll

dm/ctl/master/config.go
```diff
@@ -0,0 +1,395 @@
...
+			common.PrintLinesf("fail to write source config to file `%s`", sourceFile)
+			return err
+		}
+	}
+	return nil
+}
+
+func writeTaskCfgs(taskDir string, subTaskCfgsMap map[string]map[string]config.SubTaskConfig) error {
+	subTaskCfgsListMap := make(map[string][]*config.SubTaskConfig, len(subTaskCfgsMap))
+	// from source => task => subtask to task => source => subtask
```
lance6716, line 235:
```suggestion
	// from source => task => subtask to task => subtasks
```

dm/ctl/master/config.go
```diff
@@ -0,0 +1,395 @@
...
+	for _, subTaskCfgs := range subTaskCfgsMap {
+		for task, subTaskCfg := range subTaskCfgs {
+			clone := subTaskCfg
+			if subTaskCfgList, ok := subTaskCfgsListMap[task]; ok {
```
lance6716, line 239: maybe append to `nil` slice is tolerated by golang

dm/ctl/master/config.go
```diff
@@ -0,0 +1,395 @@
...
+				subTaskCfgsListMap[task] = append(subTaskCfgList, &clone)
+			} else {
+				subTaskCfgsListMap[task] = []*config.SubTaskConfig{&clone}
+			}
+		}
+	}
+	// from task => source => subtask to task => taskCfg
```
lance6716, line 246:
```suggestion
	// from task => subtasks to task => taskCfg
```

tests/new_relay/configs/sources/mysql-replica-01.yaml
```diff
@@ -0,0 +1,32 @@
...
+relay-binlog-name: ""
+relay-binlog-gtid: ""
+source-id: mysql-replica-01
+from:
+  host: 127.0.0.1
+  port: 3306
+  user: root
+  max-allowed-packet: null
+  session:
+    time_zone: "+00:00"
```
lance6716, line 17: not sure we should treat this line as default config

tests/tiup/lib.sh
```diff
@@ -172,11 +172,52 @@ function exec_incremental_stage2() {
...
+	exec_sql mariadb2 3306 "INSERT INTO $DB6.$TBL2 (c1, c2, c3) VALUES (413, '413', 413);"
+	exec_sql mariadb2 3306 "INSERT INTO $DB6.$TBL3 (c1, c2, c3) VALUES (414, '414', 414);"
+}
+
 function patch_nightly_with_tiup_mirror() {
 	# clone packages for upgrade.
 	# FIXME: use nightly version of grafana and prometheus after https://github.com/pingcap/tiup/issues/1334 fixed.
 	tiup mirror clone tidb-dm-nightly-linux-amd64 --os=linux --arch=amd64 \
 		--alertmanager=v0.17.0 --grafana=v5.0.1 --prometheus=v5.0.1 \
+		--dm-master=v2.0.1 --dm-worker=v2.0.1 \
```
lance6716, line 220: v2.0.1 is $1?

lance6716 (commented):
rest lgtm

dm/config/source_config.go
GMHDBJD: I add a clone func in 6fac04b , but I don't have confidence, so use it in unit_test.

Ehco1996: i think reflect is too magic and not easy for read and debug.. maybe we shold keep convert those struct manually

dm/config/task_test.go
```diff
@@ -1015,8 +1016,75 @@ func (t *testConfig) TestExclusiveAndWrongExprFilterFields(c *C) {
...
+			destFieldType := destField.Type()
+			srcFieldType := srcField.Type()
+			if destFieldType.Kind() == reflect.Ptr {
+				destFieldType = destFieldType.Elem()
+			}
+			if srcFieldType.Kind() == reflect.Ptr {
+				srcFieldType = srcFieldType.Elem()
+			}
+			if destFieldType != srcFieldType {
+				cloneValues(destField, srcField)
```
lance6716, line 1084: what can we do if the type is not matched? I think we can only panic here

dm/config/task_test.go
```diff
@@ -1015,8 +1016,75 @@ func (t *testConfig) TestExclusiveAndWrongExprFilterFields(c *C) {
...
+	destType := dest.Type()
+	srcType := src.Type()
+	if destType.Kind() == reflect.Ptr {
+		destType = destType.Elem()
+	}
+	if srcType.Kind() == reflect.Ptr {
+		srcType = srcType.Elem()
+	}
+
+	if destType.Kind() == reflect.Slice {
```
lance6716, line 1052: this `if` seems not used 🤔

lance6716 (approved):

dm/config/task_test.go
GMHDBJD: For `MySQLInstance` and `MySQLInstanceForDowngrade`

dm/config/task_test.go
GMHDBJD: For `[]*MySQLInstance` and `[]*MySQLInstanceForDowngrade` 😭

Ehco1996:
@GMHDBJD  TestTaskConfigForDowngrade reflect test faild please check it,https://ci.pingcap.net/blue/organizations/jenkins/dm_ghpr_new_test/detail/dm_ghpr_new_test/1250/pipeline

GMHDBJD:
PTAL again

PR #1897, opened 2021-07-21
chore: remove no job in queue log

Ehco1996:

as title

* only print log when enable failpoint
* do not truncate log file when run test case
* use go scripts to send feishu notify when test failed

### Check List 

Tests 

 - Integration test

Code changes

Side effects

Related changes

### DIFF
--- a/tests/_utils/test_prepare
+++ b/tests/_utils/test_prepare
@@ -144,9 +144,9 @@ function run_case() {
 
     eval ${init_table_cmd}
 
-    # truncate -s 0 $WORK_DIR/master/log/dm-master.log
-    # truncate -s 0 $WORK_DIR/worker1/log/dm-worker.log
-    # truncate -s 0 $WORK_DIR/worker2/log/dm-worker.log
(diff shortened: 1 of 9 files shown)


tests/_utils/test_prepare
```diff
@@ -144,9 +144,9 @@ function run_case() {
 
     eval ${init_table_cmd}
 
-    truncate -s 0 $WORK_DIR/master/log/dm-master.log
-    truncate -s 0 $WORK_DIR/worker1/log/dm-worker.log
-    truncate -s 0 $WORK_DIR/worker2/log/dm-worker.log
+    # truncate -s 0 $WORK_DIR/master/log/dm-master.log
```
GMHDBJD, line 147: This is not a good way. Some tests rely on logs(check_log_contains). In shardddl tests, we run different cases in a single cluster, if we keep previous logs, check_log_contains is imprecise.

lichunzhu (approved):
LGTM

GMHDBJD (approved):
LGTM

PR #1900, opened 2021-07-21
*: manual pause a task which has been pause by error

### LINKED ISSUE #1729: Task auto resume after manually pause an error task.
## Bug Report

upstream report an error after start/resume task immediately. (e.g. failed to create table)
task auto-resume
fail to execute `operate-schema` because task not pause
manually pause task
exec `operate-schema set`
3. What did you see instead?
pause task fails because error immediately occur after auto resume, we can't pause a paused task.
operate-schema fails because task auto-resume since it's paused by error
4. Versions of the cluster

        ```
        master
        ```

GMHDBJD:
close #1729 

TaskChecker will not resume a task that is manually paused(isCancel: true)
When users manual pauses a task which has been pause by error(task stage is paused while isCancel is false), mark the isCancel to true.

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

### DIFF
--- a/dm/worker/subtask.go
+++ b/dm/worker/subtask.go
@@ -473,7 +473,7 @@ func (st *SubTask) Close() {
 	updateTaskState(st.cfg.Name, st.cfg.SourceID, pb.Stage_Stopped)
 }
 
-// Pause pauses the running sub task.
+// Pause pauses a running sub task or a sub task paused by error.
 func (st *SubTask) Pause() error {
 	if st.markResultCanceled() {
 		return nil
--- a/dm/worker/subtask.go
+++ b/dm/worker/subtask.go
@@ -435,12 +435,20 @@ func (st *SubTask) setResult(result *pb.ProcessResult) {
 	st.result = result
 }
 
-func (st *SubTask) markResultCanceled() {
+// markResultCanceled mark result as canceled if stage is Paused.
+// This func is used to pause a task which has been paused by error,
+// so the task will not auto resume by task checker.
+func (st *SubTask) markResultCanceled() bool {
 	st.Lock()
 	defer st.Unlock()
-	if st.result != nil {
-		st.result.IsCanceled = true
+	if st.stage == pb.Stage_Paused {
+		if st.result != nil && !st.result.IsCanceled {
+			st.l.Info("manually pause task which has been paused by errors")
+			st.result.IsCanceled = true
+			return true
+		}
 	}
+	return false
 }
 
 // Result returns the result of the sub task.
@@ -467,13 +475,8 @@ func (st *SubTask) Close() {
 
 // Pause pauses the running sub task.
 func (st *SubTask) Pause() error {
-	if st.Stage() == pb.Stage_Paused {
-		result := st.Result()
-		if result != nil && !result.IsCanceled {
-			st.l.Info("manually pause task which has been paused by errors")
-			st.markResultCanceled()
-			return nil
-		}
+	if st.markResultCanceled() {
+		return nil
 	}
 
 	if !st.stageCAS(pb.Stage_Running, pb.Stage_Pausing) {
--- a/dm/worker/subtask.go
+++ b/dm/worker/subtask.go
@@ -435,6 +435,14 @@ func (st *SubTask) setResult(result *pb.ProcessResult) {
 	st.result = result
 }
 
+func (st *SubTask) markResultCanceled() {
+	st.Lock()
+	defer st.Unlock()
+	if st.result != nil {
+		st.result.IsCanceled = true
+	}
+}
+
 // Result returns the result of the sub task.
 func (st *SubTask) Result() *pb.ProcessResult {
 	st.RLock()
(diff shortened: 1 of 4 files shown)


dm/worker/subtask.go
```diff
@@ -459,6 +475,10 @@ func (st *SubTask) Close() {
 
 // Pause pauses the running sub task.
```
Ehco1996, line 476: maybe update the comment

syncer/streamer_controller.go
```diff
@@ -253,6 +256,12 @@ func (c *StreamerController) GetEvent(tctx *tcontext.Context) (event *replicatio
 	failpoint.Inject("GetEventError", func() {
 		err = errors.New("go-mysql returned an error")
 	})
+	failpoint.Inject("GetEventErrorOnce", func() {
+		if !mockGetEventErr {
```
Ehco1996, line 260: why we need check `mockGetEventErr` ?

GMHDBJD: Otherwise, the same error occurs when we manually resume-task.

Ehco1996: oh i know, you don't want to remove failpoint and restart worker

dm/worker/subtask.go
```diff
@@ -435,6 +435,22 @@ func (st *SubTask) setResult(result *pb.ProcessResult) {
 	st.result = result
 }
 
+// markResultCanceled mark result as canceled if stage is Paused.
+// This func is used to pause a task which has been paused by error,
```
lance6716, line 439: we should also support "pause a task which has no error, just the user want to pause it" 🤔

dm/worker/subtask.go
```diff
@@ -457,8 +473,12 @@ func (st *SubTask) Close() {
 	updateTaskState(st.cfg.Name, st.cfg.SourceID, pb.Stage_Stopped)
 }
 
-// Pause pauses the running sub task.
+// Pause pauses a running sub task or a sub task paused by error.
 func (st *SubTask) Pause() error {
+	if st.markResultCanceled() {
```
lance6716, line 478: if `markResultCanceled` is called before chaning stage, user need to pause twice to pin the paused status. I think his expectation is pausing once

dm/worker/subtask.go
GMHDBJD: I don't understand what you mean. We currently support pause-task manually for a normal task.

syncer/streamer_controller.go
lichunzhu: Can we do this by `"github.com/pingcap/dm/syncer/GetEventError=1*return"`?
Reference: https://github.com/pingcap/failpoint#key-concepts
Example: [dumpling failpoint](https://github.com/pingcap/dumpling/blob/843aff7/tests/basic/run.sh#L71)

tests/drop_column_with_index/run.sh
```diff
@@ -33,9 +36,30 @@ function run() {
...
+		"pause-task test" \
+		"\"result\": true" 1 \
+		"get event returned an error" 1
+	run_dm_ctl_with_retry $WORK_DIR "127.0.0.1:$MASTER_PORT" \
+		"query-status test" \
+		"get event returned an error" 1 \
+		"\"stage\": \"Paused\"" 1 \
+		"\"isCanceled\": true" 1
+
+	sleep 5
```
lichunzhu, line 56: why sleep 5 here?

dm/worker/subtask.go
lichunzhu: +1

tests/drop_column_with_index/run.sh
GMHDBJD: auto resume interval 5 seconds

dm/worker/subtask.go
GMHDBJD: If user pause a running task, DM will [callCurrentCancel](https://github.com/pingcap/dm/blob/068c30eb344c2556f13b1dbf7a6e55feed7ca1f3/dm/worker/subtask.go#L466) and [fetchResult](https://github.com/pingcap/dm/blob/068c30eb344c2556f13b1dbf7a6e55feed7ca1f3/dm/worker/subtask.go#L252) will mark the task paused and set isCanceled to true.
If task paused by error, isCanceled is false, so simply replace the value, and [task checker](https://github.com/pingcap/dm/blob/068c30eb344c2556f13b1dbf7a6e55feed7ca1f3/dm/worker/task_checker.go#L260) will not resume it.

lance6716: so the discription of https://github.com/pingcap/dm/issues/1729 is wrong? task will keep paused when user manually paused it, but just before manually pause, some error would occur and get a paused status, so we wrongly take it as manually pause didn't take effect?

GMHDBJD: Yes！

glorv (approved):

tests/drop_column_with_index/run.sh
lance6716: we may reduce check interval and sleep twice as it

GMHDBJD: If check-interval is too small, the task may auto resume before manually pause, and test will be unstable.

lance6716 (approved):

PR #1903, opened 2021-07-22
syncer/dbconn: add more retry error

### LINKED ISSUE #1825: v2.0.4-Transaction-level retry for errors such as TxnLockNotFound
RT, 
txnlocknotfound may appear frequently in tidb-server in some scenarios
When tidb acts as the downstream of dm and undertakes online real-time query services, the performance regression and synchronization delay caused by safe-mode have a greater impact on the business
Hope that this transaction can be retried in other ways to reduce the probability of delay

dm-worker log：
`
[error="[code=10006:class=database:scope=not-set:level=high], Message: execute statement failed: commit, RawCause: Error 8022: Error: KV error safe to retry Txn(Mvcc(TxnLockNotFound { start_ts: TimeStamp(425904341916582174), commit_ts: TimeStamp(425904342991372376)`

glorv:
Auto retry for ~`mysql.ErrInvalidConn`~ and tidb's `kv.RetryableErr` (by https://github.com/pingcap/tidb-tools/pull/452). 

close #1825

- ~When sql client interrupt (e.g. manually killed by `kill tidb xxx`), go-mysql will return `mysql.ErrInvalidConn` and this error is safe to auto-retry build the connection~
- tidb's `kv.RetryableErr` (currently only caused by TxnLockNotFound) is retryable too.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

Related changes

### DIFF
--- a/pkg/retry/errors.go
+++ b/pkg/retry/errors.go
@@ -17,6 +17,7 @@ import (
 	"database/sql/driver"
 
 	gmysql "github.com/go-mysql-org/go-mysql/mysql"
+	"github.com/go-sql-driver/mysql"
 	"github.com/pingcap/errors"
 	tmysql "github.com/pingcap/parser/mysql"
 
@@ -82,7 +83,7 @@ var (
 func IsConnectionError(err error) bool {
 	err = errors.Cause(err)
 	switch err {
-	case driver.ErrBadConn, tmysql.ErrBadConn, gmysql.ErrBadConn:
+	case driver.ErrBadConn, tmysql.ErrBadConn, gmysql.ErrBadConn, mysql.ErrInvalidConn:
 		return true
 	}
 	return false
--- a/pkg/retry/errors.go
+++ b/pkg/retry/errors.go
@@ -17,7 +17,6 @@ import (
 	"database/sql/driver"
 
 	gmysql "github.com/go-mysql-org/go-mysql/mysql"
-	"github.com/go-sql-driver/mysql"
 	"github.com/pingcap/errors"
 	tmysql "github.com/pingcap/parser/mysql"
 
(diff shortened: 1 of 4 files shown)


Ehco1996 (approved):

syncer/dbconn/db.go
```diff
@@ -160,6 +160,8 @@ func (conn *DBConn) ExecuteSQLWithIgnore(tctx *tcontext.Context, ignoreError fun
 						log.ShortError(err))
 					return false
 				}
+				tctx.L().Warn("execute sql failed by connection error", zap.Int("retry", retryTime),
+					zap.Error(err))
```
Ehco1996, line 164: can we add test for this case?

glorv: Will add test case for this after we add retry for all other connections so we can safely kill all connection between dm-worker and tidb

pkg/retry/errors.go
```diff
@@ -82,7 +83,7 @@ var (
 func IsConnectionError(err error) bool {
 	err = errors.Cause(err)
 	switch err {
-	case driver.ErrBadConn, tmysql.ErrBadConn, gmysql.ErrBadConn:
+	case driver.ErrBadConn, tmysql.ErrBadConn, gmysql.ErrBadConn, mysql.ErrInvalidConn:
```
lance6716, line 86: we may not retry the `ErrInvalidConn` when it came from `txn.Commit()`

glorv: In the situation, do the other three error can safely retry?

glorv: ok

GMHDBJD: Currently, we will ignore invalid connection error for some DDL. https://github.com/pingcap/dm/blob/fadd54342759286e7572bb100684fe48942c0ca7/syncer/error.go#L79

GMHDBJD (approved):
LGTM

lance6716 (approved):

Issue #1790, opened 2021-06-18
Quality reinforcement (12): fix unstable test cases


lance6716:
previous one https://github.com/pingcap/dm/pull/1616

- [ ] ha_master https://github.com/pingcap/dm/issues/1762
- [x] ha_master [jenkins](https://ci.pingcap.net/blue/organizations/jenkins/dm_ghpr_new_test/detail/dm_ghpr_new_test/532/pipeline/) [jenkins](https://ci.pingcap.net/blue/organizations/jenkins/dm_ghpr_new_test/detail/dm_ghpr_new_test/676/pipeline/) [jenkins](https://ci.pingcap.net/blue/organizations/jenkins/dm_ghpr_new_test/detail/dm_ghpr_new_test/680/pipeline)
    `   [2021-06-07T02:53:21.974Z] [2021/06/07 10:53:11.136 +08:00] [ERROR] [main.go:69] ["fail to start dm-master"] [error="[code=38038:class=dm-master:scope=internal:level=high], Message: fail to start embed etcd, RawCause: wal: max entry size limit exceeded"] [errorVerbose="[code=38038:class=dm-master:scope=internal:level=high], Message: fail to start embed etcd, RawCause: wal: max entry size limit exceeded\n
...
`
- [x] trackerSuite.TestCreateTableIfNotExists (cost more than 20 seconds) [jenkins](https://ci.pingcap.net/blue/organizations/jenkins/dm_ghpr_new_test/detail/dm_ghpr_new_test/607/pipeline) [jenkins](https://ci.pingcap.net/blue/organizations/jenkins/dm_ghpr_new_test/detail/dm_ghpr_new_test/844/pipeline) [jenkins](https://ci.pingcap.net/blue/organizations/jenkins/dm_ghpr_new_test/detail/dm_ghpr_new_test/893/pipeline) [jenkins](https://ci.pingcap.net/blue/organizations/jenkins/dm_ghpr_new_test/detail/dm_ghpr_new_test/895/pipeline) [jenkins](https://ci.pingcap.net/blue/organizations/jenkins/dm_ghpr_new_test/detail/dm_ghpr_new_test/890/pipeline) [jenkins](https://ci.pingcap.net/blue/organizations/jenkins/dm_ghpr_new_test/detail/dm_ghpr_new_test/927/pipeline)
[...]

PR #1905, opened 2021-07-22
test: resume task to fix unstable test case

lance6716:
part of https://github.com/pingcap/dm/issues/1790

described in comment of test

### Check List 

Tests 

 - No code

Related changes


tests/shardddl1/run.sh
```diff
@@ -264,6 +264,12 @@ function DM_RECOVER_LOCK_CASE() {
 	# TrySync tb2: joined(a,c); tb1(a,c); tb2(a,b)
 	restart_master
 
+	# because we disabled auto-resume feature to satisfy other tests, we manually resume once to handle task was paused by
```
glorv, line 267: Shall we check the task is stopped first?

lance6716: resume-task will also return `result: true` when task is not paused

lichunzhu (approved):
LGTM

Issue #1757, opened 2021-06-09
shorter or no safe-mode when resuming or finer resuming


lance6716:
## Feature Request

**Is your feature request related to a problem? Please describe:**

when DM is paused by some bursted error and resume, safe-mode will slower sync speed

**Describe the feature you'd like:**

a) because DM could know the checkpoint and last sent binlog event, we could record them in momery so next resuming only [checkpoint, last sent binlog event] needs safe-mode.
b) We try reconnect when upstream is broken, to avoid causing a paused state. For GTID-based replication, we can't build a now connection and continue to sync from a location inside a GTID, so we may need to switch to position-based replication to continue at expected location for a short time.

**Describe alternatives you've considered:**

c) like b), to handle the scenario that "retry GTID-based replication will re-send unfinished GTID event", we might only consume event when GTID event is totally seen. Before we fully see GTID event, we need to cache it in memory so might cause OOM if GTID event is huge (like it contains many delete rows)

**Teachability, Documentation, Adoption, Migration Strategy:**

lance6716:
closed by #1906

PR #1906, opened 2021-07-22
syncer: optimize safemode mechanism to improve performance

lichunzhu:
Implement https://github.com/pingcap/dm/issues/1757 method a.

Reuse `exit_safe_binlog_name`,`exit_safe_binlog_pos`,`exit_safe_binlog_gtid` in checkpoint.
1. When parse events whose location is newer than `exit_safe_binlog`, flush `exit_safe_binlog` asap to avoid duplicate in the future.
2. Only when `syncer` fails due to errors except execution errors or meet some errors when executing, don't update global checkpoint but update safe point to downstream. In other situations, update both global checkpoint and safe point.
3. Set `syncDML` and `syncDDL`'s context to `context.Background()` to make sure ddl/dmls can be replicated to downstream as much as possible event when context is canceled.

### Check List 

Tests 

 - Unit test
 - Integration test
 
Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes


syncer/syncer.go
```diff
@@ -1295,7 +1286,10 @@ func (s *Syncer) syncDML(
 			t := v.(int)
 			time.Sleep(time.Duration(t) * time.Second)
 		})
-		return db.ExecuteSQL(tctx, queries, args...)
+		// use background context to execute sqls as much as possible
+		ctctx, cancel := tctx.WithContext(context.Background()).WithTimeout(maxDMLConnectionDuration)
```
GMHDBJD, line 1290: Seems can use tctx.WithTimeout(maxDMLConnectionDuration)

lichunzhu: Want to make sure this context is not affected by the parent context.

syncer/syncer.go
```diff
@@ -1069,7 +1069,7 @@ func (s *Syncer) flushCheckPoints() error {
 	// optimistic shard info, DM-master may resolved the optimistic lock and let other worker execute DDL. So after this
 	// worker resume, it can not execute the DML/DDL in old binlog because of downstream table structure mismatching.
 	// We should find a way to (compensating) implement a transaction containing interaction with both etcd and SQL.
-	if err != nil {
+	if err != nil && (terror.ErrDBExecuteFailed.Equal(err) || terror.ErrDBUnExpect.Equal(err)) {
```
GMHDBJD, line 1072: Maybe only flushSafeModeExitPoint in we meet error?

GMHDBJD (commented):
rest LGTM

syncer/syncer.go
glorv: I think this context with an timeout durtion of `5m` is not acceptable. In the worst case, the worker need to wait 5m before exit.

glorv (commented):
rest LGTM

syncer/syncer.go
lichunzhu: For etcd error I think it's still okay to flush global checkpoint. Because it's not saved.

GMHDBJD: I'm afraid there will be some other error that leads to data loss.

GMHDBJD (approved):
LGTM

glorv (approved):

PR #1915, opened 2021-07-26
*: read downstream table before CREATE TABLE to fix IF NOT EXISTS

lance6716:

revert some of https://github.com/pingcap/dm/pull/1699

as title, we read downstream table from checkpoint/etcd/SHOW CREATE TABLE first, to deal with IF NOT EXISTS.

in future we'll change to a better strategy.

### Check List 

Tests 

 - pass original tests

Code changes

Side effects

Related changes


tests/shardddl1_1/run.sh
```diff
@@ -252,11 +252,11 @@ function DM_027_CASE() {
 	run_sql_source1 "insert into ${shardddl1}.${tb2} values (4)"
 	run_sql_source1 "insert into ${shardddl1}.${tb3} values (5,6)"
 	# we now haven't checked table struct when create sharding table
-	# there should be a error message like "Unknown column 'val' in 'field list'", "unknown column val"
-	# but different TiDB version output different message. so we only roughly match here
+	# and we'll attach IF NOT EXISTS to every CREATE TABLE and fetch downstream table first, so downstream table strcuture
```
lance6716, line 255: this will cause CREATE TABLE in upstream becomes ineffective when downstream has already contained that table

When downstream table has same column with upstream one, this PR imtroduced more correctness for genearting WHERE
When downstream table has less column, this PR changes error message from "unknown column" to "column value mismatch"
When downstream table has more column, this PR will break the synchronizing only when DM sees CREATE TABLE before any DML of this table at sync unit.

glorv (approved):

GMHDBJD (approved):
LGTM

Ehco1996 (commented):
LGTM

PR #1924, opened 2021-07-27
fix safe mode

glorv:
Store safe-mode struct in sync to avoid depending on the safe-mode struct in event context.
Set safe-mode value by the current snapshot safe-mode value .

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

Side effects

Related changes

### DIFF
--- a/syncer/syncer.go
+++ b/syncer/syncer.go
@@ -1886,10 +1886,10 @@ type eventContext struct {
 	traceSource         string
 	// safeMode is the value of syncer.safeMode when process this event
 	// syncer.safeMode's value may change on the fly, e.g. after event by pass the safeModeExitPoint
-	safeMode            bool
-	tryReSync           bool
-	startTime           time.Time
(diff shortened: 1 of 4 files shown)


lance6716 (approved):

lichunzhu (approved):
LGTM

syncer/syncer.go
```diff
@@ -154,6 +154,7 @@ type Syncer struct {
 		sync.RWMutex
 		t time.Time
 	}
+	safeMode *sm.SafeMode
```
lichunzhu, line 157: Can we add a comment here to explain why this variable must be a local variable?

glorv: So I think we can add a comment for the `safeMode` in eventContext?

lichunzhu: That's also fine. To avoid this line is changed in the future.

Issue #1095, opened 2020-09-24
stop/pause until reached the end of a transaction


csuzhangxc:
## Feature Request

**Is your feature request related to a problem? Please describe:**

DM split transactions from upstream MySQL into rows and re-aggregate rows into a new transaction as a batch.

When `stop-task` or `pause-task`, maybe only a part of rows in an upstream MySQL transaction have been committed into the downstream TiDB, in other words, the original transaction is broken in TiDB after that.

**Describe the feature you'd like:**

stop/pause the task until reached the end of a transaction for `start-task`/`pause-task` or shutdown the DM-worker process normally.

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

lance6716:
another way is revert to last consistent point (maybe recorded in checkpoint) if TiDB could flashback table

https://github.com/pingcap/tidb/issues/20302

lichunzhu:
Maybe we can refer to tidb-binlog's logic:
1. Don't send any sqls later than the current transation.
2. Close syncer until all sqls are replicated to downstream.
3. Add a boolean column `synced` in downstream checkpoint to notify whether DM really stops at a certain transaction.

https://github.com/pingcap/tidb-binlog/blob/v4.0.13/drainer/syncer.go#L484

okJiang:
How about saving to the end of the transaction(xid event) directly every time `saveTablePoint`?

okJiang:
> How about saving to the end of the transaction(last xid event) directly every time `saveTablePoint`?

Since this cannot guarantee the consistency of upstream and downstream data, we choose to delay syncer when stop/pause until the end of the transaction, and then `flushCheckPoint`.

okJiang:
Initial idea: Delay when syncer exits and job closes. https://github.com/pingcap/dm/blob/11fb5a8b63200efa2d6bbb2fa3bcb36d90e4f575/syncer/syncer.go#L3040

- Wait for the arrival of XIDEvent before closing the job channel.

Steps:

1. Before closing the job channel, set `Syncer.waitXIDType = waiting`. Then wait `Syncer.waitXIDType = waitComplete` here.
2. After setting `Syncer.waitXIDType = waiting`, `Syncer` is running normaly until `addJob` encounter a `XIDEvent`.
3. After encountering a `XIDEvent`, 
    a. set `Syncer.waitXIDType = waitComplete`
    b. stop `addJob`
    c. continue close `jobChan` in 1
[...]

PR #1928, opened 2021-07-28
syncer: stop/pause until reached the end of a transaction

okJiang:

close https://github.com/pingcap/dm/issues/1095

- add three variable
  - waitXIDJob: the waiting XID job's status when pause/stop task
  - isTransactionEnd: check whether is a transaction end now
  - waitTransactionLock

- modify the `syncer.Run()`'s `runCtx` to `context.backgound()`, so that postpone the `Run()` quit.

### Check List 

Tests 

- integration test


syncer/syncer.go
```diff
@@ -944,9 +947,22 @@ func (s *Syncer) addJob(job *job) error {
...
+	if s.waitXIDJob == waitComplete {
+		s.tctx.L().Info("All job is completed before syncer close, the coming job will be abandon", zap.Any("job", job))
+		return nil
+	}
+
+	// failpoint.Inject("checkCheckpointInMiddleOfTransaction", func() {
+	if s.waitXIDJob == waiting {
+		s.tctx.L().Debug("not receive xid job yet", zap.Any("next job", job))
+	}
+	// })
```
okJiang, line 959: I don't known why does the log not print when uncomment? @glorv

syncer/syncer.go
```diff
@@ -1404,16 +1436,36 @@ func (s *Syncer) syncDML(
...
 	defer func() {
 		if s.done != nil {
 			close(s.done)
 		}
 	}()
 
+	go func() {
+		<-ctx.Done()
+		tctx.L().Info("received subtask's done")
+		if s.isTransactionEnd {
```
okJiang, line 1451: About the assignment of `isTransactionEnd`, need to lock? @glorv @lance6716

lance6716: if there's concurrent accessing, we should use `atomic.Bool` at lease

okJiang: `atomic.Bool` is done

syncer/syncer.go
```diff
@@ -961,6 +982,7 @@ func (s *Syncer) addJob(job *job) error {
 			// flush for every DML queue
 			metrics.AddJobDurationHistogram.WithLabelValues("flush", s.cfg.Name, s.queueBucketMapping[i], s.cfg.SourceID).Observe(time.Since(startTime).Seconds())
 		}
+		s.isTransactionEnd.Store(true)
```
glorv, line 985: A  flush event doesn't means the transaction is ended.

okJiang: Yes. I am looking for the reason why it will fail in `handle_error_3` when removing it.

syncer/syncer.go
```diff
@@ -1404,16 +1436,37 @@ func (s *Syncer) syncDML(
...
 	}()
 
+	go func() {
+		<-ctx.Done()
+		tctx.L().Info("received subtask's done")
+		if s.isTransactionEnd.Load() {
+			tctx.L().Info("the last job is transaction end, done directly")
+			runCancel()
+			return
+		}
```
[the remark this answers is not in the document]
okJiang: Seems that I also will add 

```
		if s.waitXIDJob == waiting {
			s.waitXIDJob = waitComplete
		}
```

in here? https://github.com/pingcap/dm/pull/1928/files#diff-e725a03bd475dc4ea28028613ede1fd88a8cc8f881f0a1788a189e766f191830R996

syncer/syncer.go
```diff
@@ -1404,16 +1436,37 @@ func (s *Syncer) syncDML(
 
 // Run starts running for sync, we should guarantee it can rerun when paused.
 func (s *Syncer) Run(ctx context.Context) (err error) {
-	tctx := s.tctx.WithContext(ctx)
+	runCtx, runCancel := context.WithCancel(context.Background())
```
glorv, line 1439:
```suggestion
	runCtx, runCancel := context.WithCancel(context.Background())
	defer runCancel()
```

syncer/syncer.go
```diff
@@ -1404,16 +1436,37 @@ func (s *Syncer) syncDML(
...
+	tctx := s.tctx.WithContext(runCtx)
 
 	defer func() {
 		if s.done != nil {
 			close(s.done)
 		}
 	}()
 
+	go func() {
+		<-ctx.Done()
```
glorv, line 1449: if `runCtx` is also done here, e.g. we met an unretryable error, we needn't do the following works any more

syncer/syncer.go
```diff
@@ -1404,16 +1436,37 @@ func (s *Syncer) syncDML(
...
+			return
+		}
+		waitTime := 10 * time.Second
+		s.waitXIDJob = waiting
+		tricker := time.NewTicker(waitTime)
+		defer tricker.Stop()
+		select {
+		case <-runCtx.Done():
+			tctx.L().Info("received syncer's done")
+		case <-tricker.C:
```
glorv, line 1463:
```suggestion
		case <- time.After(10 * time.Second):
```
only need to wait once, so no need to explicitly create a ticker

syncer/syncer.go
okJiang: That's the reason the CI `handle_error_3` failed.

syncer/syncer.go
```diff
@@ -1404,16 +1438,37 @@ func (s *Syncer) syncDML(
...
 	defer func() {
 		if s.done != nil {
 			close(s.done)
 		}
 	}()
 
+	go func() {
+		<-ctx.Done()
+		tctx.L().Info("received subtask's done")
+		if s.isTransactionEnd.Load() {
```
Ehco1996, line 1453: why we need isTransactionEnd? can we just use waitComplete instead?

tests/checkpoint_transaction/conf/dm-worker1.toml
```diff
@@ -0,0 +1,2 @@
+name = "worker1"
```
Ehco1996, line 1: please add worker2 and source2 into test, and make sure shardddl is working properly

syncer/syncer.go
```diff
@@ -1360,6 +1386,14 @@ func (s *Syncer) syncDML(
 		case sqlJob, ok := <-jobChan:
 			metrics.QueueSizeGauge.WithLabelValues(s.cfg.Name, queueBucket, s.cfg.SourceID).Set(float64(len(jobChan)))
 			if !ok {
+				tctx.L().Info("close job chan", zap.Any("rest job", jobs))
+				affect, err = executeSQLs()
```
Ehco1996, line 1390: L1390-L1396 seems appear many times, can we put them in one function

Ehco1996: or maybe move the logic for`executing the remaining tasks in the queue ti downstream` to`s.closeJobChans`

syncer/syncer.go
okJiang: This is to determine whether the current transaction is over

syncer/syncer.go
okJiang: > L1390-L1396 seems appear many times, can we put them in one function

Can you help me think of a function name?

tests/checkpoint_transaction/conf/dm-worker1.toml
okJiang: I think ddl does not need to be tested, because the code will wait for all ddl to finish executing in `s.jobwg.Wait()`

syncer/syncer.go
okJiang: fixed in https://github.com/pingcap/dm/pull/1928/commits/b2d6355f9dfe3f5241bde6edbe5f4aaad280570b

syncer/syncer.go
```diff
@@ -944,13 +950,28 @@ func (s *Syncer) addJob(job *job) error {
 			s.tctx.L().Info("meet the second job of a GTID", zap.Any("binlog position", lastLocation))
 		}
 	})
+	if s.waitXIDJob == waitComplete {
```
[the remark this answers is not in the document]
okJiang: fixed in https://github.com/pingcap/dm/pull/1928/commits/b2d6355f9dfe3f5241bde6edbe5f4aaad280570b

syncer/syncer.go
Ehco1996: I just noticed something a little different，why does the L1388 need to continue

okJiang: > I just noticed something a little different，why does the L1388 need to continue

Because the `jobs` may not clean completely.

Ehco1996: in `fatalF` jobs will clear by `clearF`, i think't we don't need into another round of for loop, because jobChan already closed

okJiang: My bad, it shall not be `continue`.

okJiang:
During the integration test, there is a situation that the big transaction has not started yet, but the task has already started to pause. At this time, the test will fail in the first loop. But it will also not succeed until the second loop. Because the first transaction in the front will start to synchronize in second loop, then syncer pause again, and downstream only synchronizes the first transaction, and the second transaction is not synchronized. This will not achieve the test results. 

So in this case, it is useless to loop to reach the middle of the transaction and delete it directly.

The current method is to wait for 0.1s to reach the beginning of the transaction. But this will make the test unstable. For example, if the default pre-operation of syncer before synchronization exceeds 0.1s, the test will fail again. If it happens, manually adjust the waiting time?

syncer/syncer.go
```diff
@@ -1895,6 +1945,10 @@ func (s *Syncer) Run(ctx context.Context) (err error) {
 
 			job := newXIDJob(currentLocation, startLocation, currentLocation)
 			err2 = s.addJobFunc(job)
+			if s.waitXIDJob == waitComplete {
+				runCancel()
```
glorv, line 1949: Should not call this cancel, it will cause other sub routine like syncDML or syncDDL exit before finish all remain tasks.

syncer/syncer.go
```diff
@@ -1360,6 +1381,14 @@ func (s *Syncer) syncDML(
 		case sqlJob, ok := <-jobChan:
 			metrics.QueueSizeGauge.WithLabelValues(s.cfg.Name, queueBucket, s.cfg.SourceID).Set(float64(len(jobChan)))
 			if !ok {
+				tctx.L().Info("close job chan", zap.Any("rest job", jobs))
```
glorv, line 1384: I think we can change ctx in L1340-1342 depend on input ctx directly instead of background

okJiang: @lance6716 @lichunzhu PTAL, I don't know why use background.

syncer/syncer.go
okJiang: I think we can retain it. Create `dmlCtxs` and `ddlCtx` depend on `runCtx` in syncDML and syncDDL. When the `runCtx` canceled, `dmlCtxs` and `ddlCtx` will cancel until theri job completed.

glorv: Simple remove this line and depend on the defer block to do similar things is enough.

okJiang: > Simply remove this line and depend on the defer block to do similar things is enough.

Is it also exit above problem?

> Should not call this cancel, it will cause other sub routine like syncDML or syncDDL exit before finish all remain tasks.

syncer/syncer.go
lance6716: I remember the original background context is used to make sure after user paused the task, some remaining job could still be executed because the `tctx` is passed from `(*SubTask).run`, and will be canceled when `(*SubTask).callCurrCancel`

Now in `(*Syncer).Run` we have changed the context chain, I'll take a look soon

syncer/syncer.go
```diff
@@ -1685,6 +1741,11 @@ func (s *Syncer) Run(ctx context.Context) (err error) {
 		startTime := time.Now()
 		e, err = s.getEvent(tctx, currentLocation)
 
+		if waitXIDStatus(s.waitXIDJob.Load()) == waitComplete {
+			s.tctx.L().Info("All event is completed before syncer close, the coming event will be reject", zap.Any("event", e))
+			return nil
```
lance6716, line 1746: this return may trigger the `defer runCancel()` in line 1448, so `case <-runCtx.Done():` in line 1470 is entered and we have no time to wait job executed. (we change `s.waitXIDJob` to `waitComplete` based on receiving the XID event, at that time previous DML jobs is not executed) 🤔

okJiang: We will wait job in L1639 first, then `defer runCancel()`

syncer/syncer.go
```diff
@@ -1895,6 +1952,9 @@ func (s *Syncer) Run(ctx context.Context) (err error) {
 
 			job := newXIDJob(currentLocation, startLocation, currentLocation)
 			err2 = s.addJobFunc(job)
+			if waitXIDStatus(s.waitXIDJob.Load()) == waitComplete {
```
lance6716, line 1955: is this branch needed 🤔

tests/checkpoint_transaction/conf/dm-task.yaml
```diff
@@ -0,0 +1,49 @@
...
+  password: ""
+
+mysql-instances:
+  - source-id: "mysql-replica-01"
+    block-allow-list:  "instance"
+    mydumper-config-name: "global"
+    loader-config-name: "global"
+    syncer-config-name: "global"
+
+  # - source-id: "mysql-replica-02"
```
lance6716, line 23: we can delete the config now

tests/checkpoint_transaction/run.sh
```diff
@@ -0,0 +1,81 @@
...
+	# start a task in all mode
+	dmctl_start_task_standalone $cur/conf/dm-task.yaml
+
+	# check diff
+	check_sync_diff $WORK_DIR $cur/conf/diff_config.toml
+
+	# copy file
+	cp $cur/data/db1.increment1.sql $WORK_DIR/db1.increment1.sql
+	run_sql_file $WORK_DIR/db1.increment1.sql $MYSQL_HOST1 $MYSQL_PORT1 $MYSQL_PASSWORD1
+	sleep 0.1 # wait big_transaction start
```
lance6716, line 34: not sure if this will introduce an unstable test. We can inject a sleep to make sure the sleep time will end in a `false` `isTransactionEnd` status.

tests/checkpoint_transaction/run.sh
```diff
@@ -0,0 +1,81 @@
...
+	check_rpc_alive $cur/../bin/check_worker_online 127.0.0.1:$WORKER1_PORT
+	dmctl_operate_source create $cur/conf/source1.yaml $SOURCE_ID1
+
+	# start a task in all mode
+	dmctl_start_task_standalone $cur/conf/dm-task.yaml
+
+	# check diff
+	check_sync_diff $WORK_DIR $cur/conf/diff_config.toml
+
+	# copy file
```
lance6716, line 31: seems the copy is not needed because we didn't edit this file.

lance6716 (commented):
rest lgtm

syncer/syncer.go
okJiang: If remove this branch, when dose `runCtx.cancel()`?

lance6716: there's another check under `s.getEvent`, at line 1740

okJiang:
Sleep 100ms, when each dml job comes, and here sleep means that the transaction has started. As long as the sleep time in the test is slightly longer than the injected sleep, the transaction can be guaranteed to start. Then since the transaction has `30` dmls and each has 100ms sleep, it can be guaranteed that after 2s sleep, pause/stop can be started before the end of the transaction.

syncer/syncer.go
```diff
@@ -979,6 +1006,11 @@ func (s *Syncer) addJob(job *job) error {
 		startTime := time.Now()
 		s.tctx.L().Debug("queue for key", zap.Int("queue", queueBucket), zap.String("key", job.key))
 		s.jobs[queueBucket] <- job
+		s.isTransactionEnd.Store(false)
```
glorv, line 1009: should also set transaction end to true after ddl event

okJiang: > from lance https://github.com/pingcap/dm/pull/1928#discussion_r684867458
> we initialized `s.isTransactionEnd.Store(true)`, and there should always be a XID event after DML events commit, so I think this logic to handle DDL event is not needed.

lichunzhu, line 1009: Will this scene cause a problem?
1. run `s.jobs[queueBucket] <- job` first, dml are sent to `jobs` channel.
2. `jobs` goroutine starts to execute this dml SQL. Some sqls are written to downstream.
3. `s.isTransactionEnd.Load()` is true, `runCancel()` directly.
4. run L1009, `s.isTransactionEnd.Store(false)`

Some sqls out of this transaction are still written to downstream.

lichunzhu (commented):
I think we also need to think about the DDLs. Here is a part of MySQL's binlog. Query event expect `BEGIN` doesn't have a XID event but I think we can still use them as "the end of a transaction" described in this PR.
```
# at 522
#210803 21:48:01 server id 123421  end_log_pos 587 CRC32 0x135c1d9d 	GTID	last_committed=1	sequence_number=2	rbr_only=yes
/*!50718 SET TRANSACTION ISOLATION LEVEL READ COMMITTED*//*!*/;
SET @@SESSION.GTID_NEXT= 'e8e592a6-7a59-11eb-8da1-0242ac110002:45'/*!*/;
# at 587
#210803 21:48:01 server id 123421  end_log_pos 655 CRC32 0x0b6de26e 	Query	thread_id=529	exec_time=0	error_code=0
SET TIMESTAMP=1627998481/*!*/;
BEGIN
/*!*/;
# at 655
#210803 21:48:01 server id 123421  end_log_pos 709 CRC32 0xaab1e461 	Table_map: `dmctl`.`t_2` mapped to number 1484
# at 709
#210803 21:48:01 server id 123421  end_log_pos 792 CRC32 0xde94ec35 	Update_rows: table id 1484 flags: STMT_END_F
### UPDATE `dmctl`.`t_2`
### WHERE
###   @1=12 /* LONGINT meta=0 nullable=0 is_null=0 */
###   @2=683005299 /* INT meta=0 nullable=1 is_null=0 */
###   @3='lsATr' /* VARSTRING(80) meta=80 nullable=1 is_null=0 */
###   @4='mNOpyfu' /* VARSTRING(40) meta=40 nullable=1 is_null=0 */
### SET
###   @1=12 /* LONGINT meta=0 nullable=0 is_null=0 */
###   @2=683005299 /* INT meta=0 nullable=1 is_null=0 */
###   @3='lsATr' /* VARSTRING(80) meta=80 nullable=1 is_null=0 */
###   @4='' /* VARSTRING(40) meta=40 nullable=1 is_null=0 */
# at 792
#210803 21:48:01 server id 123421  end_log_pos 823 CRC32 0xaeeaca20 	Xid = 4490
COMMIT/*!*/;
# at 823
#210803 21:48:03 server id 123421  end_log_pos 888 CRC32 0x00de7cf5 	GTID	last_committed=2	sequence_number=3	rbr_only=no
SET @@SESSION.GTID_NEXT= 'e8e592a6-7a59-11eb-8da1-0242ac110002:46'/*!*/;
# at 888
#210803 21:48:03 server id 123421  end_log_pos 999 CRC32 0xd99466f0 	Query	thread_id=532	exec_time=0	error_code=0
SET TIMESTAMP=1627998483/*!*/;
alter table dmctl.t_1 add column aaa int
/*!*/;
# at 999
#210803 21:48:03 server id 123421  end_log_pos 1064 CRC32 0xc5988002 	GTID	last_committed=3	sequence_number=4	rbr_only=no
SET @@SESSION.GTID_NEXT= 'e8e592a6-7a59-11eb-8da1-0242ac110002:47'/*!*/;
# at 1064
#210803 21:48:03 server id 123421  end_log_pos 1175 CRC32 0xe9087555 	Query	thread_id=533	exec_time=0	error_code=0
SET TIMESTAMP=1627998483/*!*/;
alter table dmctl.t_2 add column aaa int
/*!*/;
```

syncer/syncer.go
```diff
@@ -971,6 +998,10 @@ func (s *Syncer) addJob(job *job) error {
 		queueBucket = s.cfg.WorkerCount
 		startTime := time.Now()
 		s.jobs[queueBucket] <- job
+		if waitXIDStatus(s.waitXIDJob.Load()) == waiting {
```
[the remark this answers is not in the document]
lichunzhu: If we are executing a DDL when we receive a `<-ctx.Done()`, the current DDL will definitely be canceled after this PR.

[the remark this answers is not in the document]
lance6716: IMO this is expected behaviour because DDL may cost a long time. Do you mean we should wait 10s?

syncer/syncer.go
```diff
@@ -1316,7 +1348,7 @@ func (s *Syncer) syncDML(
 			time.Sleep(time.Duration(t) * time.Second)
 		})
 		// use background context to execute sqls as much as possible
-		ctctx, cancel := tctx.WithContext(context.Background()).WithTimeout(maxDMLExecutionDuration)
+		ctctx, cancel := tctx.WithTimeout(maxDMLExecutionDuration)
```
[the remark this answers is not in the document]
glorv: So what about add a wg.wait when find isTransactionEnd before do `runCancel`. Or we should rollback to use the context.Background to handle this situation.

syncer/syncer.go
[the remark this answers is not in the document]
lichunzhu: Maybe. I'm afraid some DDLs are not reentrant. But since we have ignored some database/table/index/column exists/notExists errors, it may also be fine.

syncer/syncer.go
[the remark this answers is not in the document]
lichunzhu: I think both SGTM.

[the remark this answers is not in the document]
okJiang: > Suggest using background context here. To avoid being affected by other contexts.
> If `isTransactionEnd` is true and we directly run `runCancel`, the safePoint and new global checkpoint won't be saved to the downstream database.

Are you talking about this problem?

https://github.com/pingcap/dm/pull/1928/#discussion_r684871196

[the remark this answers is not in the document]
glorv: Maybe just revert this change and use the origin background ctx can make things a little easier @okJiang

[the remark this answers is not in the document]
okJiang: I think add a `jobwg.Wait` before `runCancel()` is enough.

glorv (approved):

syncer/syncer.go
okJiang: You can see the latest commit https://github.com/pingcap/dm/pull/1928/commits/d0f7737dfa6874a4a8aaa7ff5f71a3ae81281513 whether solve this problem.

okJiang: Maybe:
```
                                  <- xid event
                                  <- s.isTransactionEnd.Store(true)
if s.isTransactionEnd.Load() {    <- enter if
	                              <- dml event
                                  <- s.isTransactionEnd.Store(false)
	                              <- s.jobs[queueBucket] <- job
	s.jobWg.Wait()
	tctx.L().Info("the last job is transaction end, done directly")
	runCancel()
	s.waitTransactionLock.Unlock()
	return
}
s.isTransactionEnd.Load()
```

also will write other sqls to downstream.

I also thought about that 

```
if s.isTransactionEnd.Load() {
    s.waitXIDJob.Store(int64(waitComplete)) // add
	s.jobWg.Wait()
	tctx.L().Info("the last job is transaction end, done directly")
	runCancel()
	s.waitTransactionLock.Unlock()
	return
}
```

However, once the `s.waitXIDJob.Store(int64(waitComplete))` executed after the check https://github.com/pingcap/dm/pull/1928/files#diff-e725a03bd475dc4ea28028613ede1fd88a8cc8f881f0a1788a189e766f191830R1753, the problem also exits.

lance6716 (approved):

syncer/syncer.go
```diff
@@ -1404,16 +1446,47 @@ func (s *Syncer) syncDML(
...
+	go func() {
+		<-ctx.Done()
+		select {
+		case <-runCtx.Done():
+		default:
+			tctx.L().Info("received subtask's done")
+
+			s.waitTransactionLock.Lock()
+			if s.isTransactionEnd {
+				s.jobWg.Wait()
```
lichunzhu, line 1468: If we wait here, it seems that we can't add dml job anymore. Will this cause a problem?

okJiang: It just waits added job completed.

syncer/syncer.go
```diff
@@ -908,6 +926,9 @@ var (
 )
 
 func (s *Syncer) addJob(job *job) error {
+	s.waitTransactionLock.Lock()
```
lichunzhu, line 929: I think we'd better check `waitXIDJob` after L929-L930.
Suggest the following case:
1. run `runCancel()` and `s.waitTransactionLock.Unlock()` in L1470-L1471.
2. `syncer.Run` quits and `s.done` is closed.
3. run `syncer.stopSync` and `s.closeJobChans`
4. `syncer.addJob` sends a dml job to the closed job channel.

lichunzhu (commented):
Rest LGTM

syncer/syncer.go
okJiang: SGTM. 

How about delete L1397-L1404? Because the job in jobQueue have been executed before `closeJobChan()`.

lance6716:
@lichunzhu please check if this PR can be merged

syncer/syncer.go
```diff
@@ -1360,6 +1399,9 @@ func (s *Syncer) syncDML(
 		case sqlJob, ok := <-jobChan:
 			metrics.QueueSizeGauge.WithLabelValues(s.cfg.Name, queueBucket, s.cfg.SourceID).Set(float64(len(jobChan)))
 			if !ok {
+				if len(jobs) > 0 {
+					tctx.L().Warn("have unexecuted jobs when close job chan!", zap.Any("rest job", jobs))
+				}
```
lichunzhu, line 1404: Why do we directly return here instead of executing the remained SQLs?

syncer/syncer.go
```diff
@@ -1404,16 +1446,48 @@ func (s *Syncer) syncDML(
...
+				s.waitXIDJob.Store(int64(waitComplete))
+				s.jobWg.Wait()
+				tctx.L().Info("the last job is transaction end, done directly")
+				runCancel()
+				s.waitTransactionLock.Unlock()
+				return
+			}
+			s.waitTransactionLock.Unlock()
+
+			s.waitXIDJob.Store(int64(waiting))
```
lichunzhu, line 1477: Can we switch these two lines?
Imagine that we run L982 to set `waitComplete` first, and then we set a `waiting` here. We must wait for one more transaction after that.

glorv: L982 depends on CAS operation, so if the current status is not waiting, then it'll do nothing 🤔️

syncer/syncer.go
glorv: This is the same logic as before. I prefer to execute the remaining jobs before return. Though currently, the caller can guarantee that all messages have been processed before the chan is closed.

syncer/syncer.go
lichunzhu: > L982 depends on CAS operation, so if the current status is not waiting, then it'll do nothing 🤔️

I know. Then we will set `waitXIDJob` to `waiting` in L1477 and syncer will continue sending dml event to jobChan but we don't have to do that.

okJiang: > Can we switch these two lines?
> Imagine that we run L982 to set `waitComplete` first, and then we set a `waiting` here. We must wait for one more transaction after that.

SGTM, I will switch them in next commit.

syncer/syncer.go
okJiang: Because the job in jobQueue have been executed before `closeJobChan()`. This is because subtask's(also syncer's) ctx must cancel before `closeJobChan()`, and that means all jobs will be executed before exit `syncer.Run()`.

tests/checkpoint_transaction/run.sh
```diff
@@ -0,0 +1,81 @@
...
+	run_sql_file $cur/data/db1.increment2.sql $MYSQL_HOST1 $MYSQL_PORT1 $MYSQL_PASSWORD1
+	# wait transaction start
+	# you can see why sleep in https://github.com/pingcap/dm/pull/1928#issuecomment-895820239
+	sleep 2
+	echo "stop task"
+	run_dm_ctl $WORK_DIR "127.0.0.1:$MASTER_PORT" \
+		"stop-task test" \
+		"\"result\": true" 2
+	# check the point is the middle of checkpoint
+	num=$(grep "not receive xid job yet" $WORK_DIR/worker1/log/dm-worker.log | wc -l)
```
lichunzhu, line 66: Should we truncate $WORK_DIR/worker1/log/dm-worker.log before?

lichunzhu (commented):
Rest LGTM

syncer/syncer.go
lichunzhu: It's a little tricky here but SGTM.

syncer/syncer.go
```diff
@@ -944,10 +965,23 @@ func (s *Syncer) addJob(job *job) error {
...
 		}
 	})
+
+	failpoint.Inject("checkCheckpointInMiddleOfTransaction", func() {
+		if waitXIDStatus(s.waitXIDJob.Load()) == waiting {
+			s.tctx.L().Info("not receive xid job yet", zap.Any("next job", job))
+		}
+	})
+
+	if waitXIDStatus(s.waitXIDJob.Load()) == waitComplete {
```
okJiang, line 975: If only check here and there is no subsequent job entry, it will wait for `maxPauseOrStopWaitTime`.

So add a check in `syncer.Run()`.

lichunzhu (approved):
LGTM

Issue #1212, opened 2020-10-23
evict leader can't take effect


lance6716:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do? If possible, provide a recipe for reproducing the error.

in a test case [jenkins](https://internal.pingcap.net/idc-jenkins/blue/organizations/jenkins/dm_ghpr_test/detail/dm_ghpr_test/7165/pipeline/), evicting leader failed.

2. What did you expect to see?

evict success

3. What did you see instead?

the DM-master which should be retired, gets the election key again so become leader.

[Untitled-2.log](https://github.com/pingcap/dm/files/5427934/Untitled-2.log)

lance6716:
evictLeader did take effect, because becoming -> found evict -> retiring. but why keeps becoming again 🤔

lance6716:
we may let `elec` explictly `Resign` to remove its "campaign key", otherwise `Observe` will read this old key

lance6716:
need a new accessible CI log with etcd DEBUG log to investigate

lance6716:
[master-cant-resign.log](https://github.com/pingcap/dm/files/6943674/master-cant-resign.log)

from the log we can see

the first time DM failed the campaign, etcd has logged a txn to delete old election key

```
[2021-08-05T08:57:07.415Z] [2021/08/05 16:54:04.297 +08:00] [DEBUG] [interceptor.go:181] ["request stats"] [component="embed etcd"] ["start time"=2021/08/05 16:54:04.293 +08:00] ["time spent"=3.922985ms] [remote=127.0.0.1:50806] ["response type"=/etcdserverpb.KV/Txn] ["request count"=1] ["request size"=38] ["response count"=0] ["response size"=40] ["request content"="compare:<target:CREATE key:\"/dm-master/leader/45157b1585041104\" create_revision:4 > success:<request_delete_range:<key:\"/dm-master/leader/45157b1585041104\" > > failure:<>"]
[2021-08-05T08:57:07.415Z] [2021/08/05 16:54:04.298 +08:00] [INFO] [election.go:271] ["fail to campaign"] [component=election] ["current member"="{\"id\":\"master5\",\"addr\":\"localhost:8661\"}"] [error="context canceled"]
```

but the second time etcd runs an empty txn

```
[...]

PR #1946, opened 2021-08-02
test: fix unstable evict test

gozssky:
 Fix unstable evict leader test. Part of https://github.com/pingcap/dm/issues/1790.

When evicting a leader, we just signal the `resignCh` channel. This operation is asynchronous, so new leader may not be elected immediately after the request returns. We need to wait for a proper period of time.

https://github.com/pingcap/dm/blob/ccddd7dd71019a07695ad0d208da29a1442816e1/pkg/election/election.go#L401-L405

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

Related changes


tests/ha_master/run.sh
```diff
@@ -58,12 +58,17 @@ function test_evict_leader() {
...
+		  [ "$NEW_LEADER_NAME" = "$LEADER_NAME" ] && continue
+		  sleep 1
+		done
 		if [ "$NEW_LEADER_NAME" = "$LEADER_NAME" ]; then
-			echo "leader evict failed"
-			exit 1
+		  echo "leader evict failed"
+		  exit 1
 		fi
+    echo "new leader is $NEW_LEADER_NAME"
```
lance6716, line 71: make fmt

lance6716:
in my mind the behaviour is not "wait for a proper period of time", but reading the debug log of etcd I found the election key is not deleted for a long time. Is this the same thing?

pkg/election/election.go
```diff
@@ -246,14 +246,15 @@ func (e *Election) campaignLoop(ctx context.Context, session *concurrency.Sessio
 		elec := concurrency.NewElection(session, e.key)
 		ctx2, cancel2 := context.WithCancel(ctx)
 
+		campaignWg.Add(1)
```
lance6716, line 249: does this change has some effects?

gozssky:
I'm not sure whether this PR can fix https://github.com/pingcap/dm/issues/1212 😂.  I add some sleep after `case <-e.resignCh`, the integration test will fail.

pkg/election/election.go
gozssky: Probably not. This is a problem that I found by reading code, but the probability of this problem triggering is very small.

lance6716 (approved):

Ehco1996 (approved):
LGTM

pkg/election/election.go
gozssky: @lance6716  If we evict leader before line 246, can campaign be canceled successfully?

lance6716: then in line 260, the atomic variable loaded `false` and the campaign goroutine will not work

PR #1947, opened 2021-08-02
*: Add OpenAPI

Ehco1996:

**This PR contains all the changes related to OpenApi, but is too big for review**

**So I will break this PR into smaller PRs and associate them to this** https://github.com/pingcap/dm/issues/1982

1. Prepare an openAPI 3.0 compliant yaml file
2. The yaml file is used to generate the structs needed at api level, such as the request and return bodies
3. Start a [echo](https://github.com/labstack/echo) http server on dm-master with a new port
4. Implement all corresponding interfaces in dm-master

### A few points that might be helpful in review

* Added `make generate_openapi` command to generate new golang code files when the [yaml file ](https://github.com/Ehco1996/dm/blob/api/openapi/spec/dm.yaml)changes
* All generated files are under the `openapi/` folder, and all files are named `gen.xx.go`
* The logic for the API implementation is in `dm/master/openapi.go`.
* Only the dm-master that is currently the leader can process the request. When a non-`leader` node accepts the request, it will return an http redirect, redirecting the request to the current leader node

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes


tests/openapi/run.sh
```diff
@@ -0,0 +1,62 @@
...
+	check_rpc_alive $cur/../bin/check_master_online 127.0.0.1:$MASTER_PORT2
+
+	test_source
+
+}
+
+cleanup_data openapi
+cleanup_process
+
+run $*
```
okJiang, line 58:
```suggestion
run
```

Ehco1996:
closed for https://github.com/pingcap/dm/pull/1999

Issue #1845, opened 2021-07-07
task pause spent a lot of time


GMHDBJD:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do? If possible, provide a recipe for reproducing the error.
upstream database connection error(master-slave switch/network connection error)
2. What did you expect to see?
task pause quickly
3. What did you see instead?
```
[2021/07/07 15:09:36.460 +08:00] [INFO] [syncer.go:1494] ["deadline exceeded when fetching binlog event"] [task=test] [unit="binlog replication"]
[...]

PR #1954, opened 2021-08-03
syncer: remove unreasonable context.Background()

okJiang:
close https://github.com/pingcap/dm/issues/1845

replace `Query` with `QueryContext`

### Check List 

Tests


syncer/syncer.go
```diff
@@ -2894,7 +2895,6 @@ func (s *Syncer) printStatus(ctx context.Context) {
 				}
 			}
 
-			ctx2, cancel2 := context.WithTimeout(ctx, utils.DefaultDBTimeout)
 			latestMasterPos, latestmasterGTIDSet, err = s.getMasterStatus(ctx2)
```
GMHDBJD, line 2898: Maybe use another context ctx3?

GMHDBJD (commented):
rest LGTM

syncer/syncer.go
lichunzhu:
```suggestion
			ctx2, cancel2 = context.WithTimeout(ctx, utils.DefaultDBTimeout)
```

lance6716: call previous cancel2 before it's overwritten

okJiang: fixed, PTAL

lichunzhu (approved):
LGTM

PR #1955, opened 2021-08-04
[WIP] scheduler: add latch and support concurrent create sources

lance6716:
### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

### DIFF
--- a/dm/master/scheduler/latch.go
+++ b/dm/master/scheduler/latch.go
@@ -1,0 +1,61 @@
+// Copyright 2021 PingCAP, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package scheduler
+
+import (
+	"sync"
+
+	"github.com/pingcap/errors"
+)
+
+// latches provides a simple way to block concurrent accessing to one resource, if caller tries to acquire latch before
+// accessing resources.
+type latches struct {
+	mu    sync.Mutex
+	inUse map[string]struct{}
+}
+
+// releaseFunc wraps on releasing a latch. It is safe to call multiple times and compiler can warn you of not used
+// releaseFunc variables.
+type releaseFunc func()
+
+func newLatches() *latches {
+	return &latches{
+		inUse: map[string]struct{}{},
+	}
+}
+
+func (l *latches) acquire(name string) (releaseFunc, error) {
+	l.mu.Lock()
+	defer l.mu.Unlock()
+	if _, ok := l.inUse[name]; ok {
+		return nil, errors.Errorf("%s is in use by other client", name)
+	}
+
+	l.inUse[name] = struct{}{}
+	var once sync.Once
+	return func() {
+		once.Do(func() {
+			l.release(name)
+		})
+	}, nil
+}
+
+// release should not be called directly, it's recommended to wrap it with releaseFunc to avoid release a latch that not
+// belongs to caller.
+func (l *latches) release(name string) {
+	l.mu.Lock()
+	delete(l.inUse, name)
+	l.mu.Unlock()
+}
--- a/dm/master/scheduler/latch_test.go
+++ b/dm/master/scheduler/latch_test.go
@@ -1,0 +1,162 @@
+// Copyright 2021 PingCAP, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package scheduler
+
+import (
+	"sort"
+	"sync"
+	"time"
+
+	. "github.com/pingcap/check"
+)
+
+type testLatch struct{}
+
+var _ = Suite(&testLatch{})
+
+func (t *testLatch) TestOneAcquireSuccess(c *C) {
+	var (
+		l = newLatches()
+		fire = make(chan struct{})
+		success = make(chan string, 2)
+		fail = make(chan struct{}, 8)
+		group1 = "group1"
+		group2 = "group2"
+		wg sync.WaitGroup
+	)
+
+	for i := 0; i < 10; i++ {
+		var group string
+		if i < 5 {
+			group = group1
+		} else {
+			group = group2
+		}
+
+		wg.Add(1)
+		go func(name string) {
+			defer wg.Done()
+			<- fire
+			_, err := l.acquire(name)
+			if err != nil {
+				fail <- struct{}{}
+			} else {
+				success <- name
+			}
+		}(group)
+	}
+
+	for i := 0; i < 10; i++ {
+		fire <- struct{}{}
+	}
+	wg.Wait()
+
+	c.Assert(fail, HasLen, 8)
+
+	var succNames []string
+	succNames = append(succNames, <-success)
+	succNames = append(succNames, <-success)
+	sort.Strings(succNames)
+	c.Assert(succNames, DeepEquals, []string{group1, group2})
+}
+
+func (t *testLatch) TestAcquireAfterRelease(c *C) {
+	var (
+		l = newLatches()
+		fire = make(chan struct{})
+		success int
+		group = "group1"
+		wg sync.WaitGroup
+	)
+
+	for i := 0; i < 5; i++ {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			<- fire
+
+			for {
+				release, err := l.acquire(group)
+				if err != nil {
+					time.Sleep(10*time.Millisecond)
+				} else {
+					time.Sleep(20*time.Millisecond)
+					success++
+					release()
+					return
+				}
+			}
+		}()
+	}
+
+	for i := 0; i < 5; i++ {
+		fire <- struct{}{}
+	}
+
+	wg.Wait()
+	c.Assert(success, Equals, 5)
+}
+
+func (t *testLatch) TestMultiRelease(c *C) {
+	var (
+		l = newLatches()
+		names = []string{"name1", "name2", "name3"}
+		fire = make(chan struct{}, len(names))
+		wg sync.WaitGroup
+	)
+
+	for repeat := 0; repeat < 3; repeat ++ {
+		for i := range names {
+			wg.Add(1)
+			go func(name string) {
+				defer wg.Done()
+				<- fire
+				release, err := l.acquire(name)
+				c.Assert(err, IsNil)
+				release()
+				// will not panic or cause other error
+				release()
+			}(names[i])
+		}
+
+		for _ = range names {
+			fire <- struct{}{}
+		}
+		wg.Wait()
+	}
+}
+
+func (t *testLatch) TestWontReleaseOther(c *C) {
+	var (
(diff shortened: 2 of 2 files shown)


dm/master/scheduler/scheduler.go
```diff
@@ -84,7 +88,7 @@ type Scheduler struct {
 	// - recover from etcd (calling `recoverSubTasks`).
 	// delete:
 	// - remove/stop subtask by user request (calling `RemoveSubTasks`).
-	subTaskCfgs map[string]map[string]config.SubTaskConfig
+	subTaskCfgs sync.Map
```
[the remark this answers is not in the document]
Ehco1996: By the way, I would like to ask again if sync.Map is used here, is there no need to use a lock at the outside?

dm/master/scheduler/latch.go
```diff
@@ -0,0 +1,62 @@
...
+	"sync"
+
+	"github.com/pingcap/errors"
+)
+
+// latches provides a simple way to block concurrent accessing to one resource, if caller tries to acquire latch before
+// accessing resources.
+type latches struct {
+	mu    sync.Mutex
+	inUse map[string]struct{}
```
Ehco1996, line 26: can we consider using some third party implementation? such as [kmutex](https://github.com/im7mortal/kmutex)

Ehco1996 (commented):
I haven't fully understood it yet, I'll read it again in a while

dm/master/scheduler/latch.go
lance6716: the API is slightly different, kmutex will be blocked which will slow the response to user. Let's wait other reviewers' opinion.

Also we can fork or contribute to that repo for a TryLock method.

dm/master/scheduler/scheduler.go
```diff
@@ -70,6 +70,10 @@ type Scheduler struct {
 
 	etcdCli *clientv3.Client
 
+	// must acquire latch from subtaskLatch before accessing subTaskCfgs and expectSubTaskStages
+	// TODO: also sourceLatch, relayLatch?
+	subtaskLatch *latches
```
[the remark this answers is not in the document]
lance6716: will address in next commit

dm/master/scheduler/scheduler.go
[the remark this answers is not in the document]
lance6716: other key of map is changing.

what's your meaning of "a lock at the outside"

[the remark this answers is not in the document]
Ehco1996: > Please tell me if the write operation is already locked with latch lock , why still need to change to sync.Map here ?

I figured it out, because those who don't use sync.Map will have problems with concurrent map read and write, golang will report a `fatal error: concurrent map read and map write`.

> a lock at the outside

i mean https://github.com/lance6716/dm/blob/latch/dm/master/scheduler/scheduler.go#L671

I may have missed something but i don't see what the correctness problem would be if I read and write `s.subTaskCfgs ` by calling `s.AddSubTasks` concurrently

dm/master/scheduler/scheduler.go
```diff
@@ -724,14 +749,22 @@ func (s *Scheduler) RemoveSubTasks(task string, sources ...string) error {
...
+	if err != nil {
+		return terror.ErrSchedulerLatchInUse.Generate("RemoveSubTasks", task)
+	}
+	defer release()
+
 	// 1. check the task exists.
-	stagesM, ok1 := s.expectSubTaskStages[task]
-	cfgsM, ok2 := s.subTaskCfgs[task]
+	v1, ok1 := s.expectSubTaskStages.Load(task)
+	v2, ok2 := s.subTaskCfgs.Load(task)
```
okJiang, line 760: Maybe more understandable naming. For all `v`.

```suggestion
    stagesMapV, ok1 := s.expectSubTaskStages.Load(task)
	cfgsMapV, ok2 := s.subTaskCfgs.Load(task)
```

dm/master/server.go
```diff
@@ -450,23 +447,27 @@ func (s *Server) StartTask(ctx context.Context, req *pb.StartTaskRequest) (*pb.S
...
-				s.removeMetaLock.Unlock()
 				return resp, nil
 			}
 			err = s.removeMetaData(ctx, cfg)
 			if err != nil {
 				resp.Msg = terror.Annotate(err, "while removing metadata").Error()
-				s.removeMetaLock.Unlock()
 				return resp, nil
 			}
+			release()
```
okJiang, line 468: Is it a repeat?

dm/master/scheduler/scheduler.go
lance6716: they will be quickly converted to a meaningful name, just 7 lines below 😂

dm/master/server.go
lance6716: yes, like defer cancel() and cancel()

dm/master/scheduler/scheduler.go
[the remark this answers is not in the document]
lance6716: acquire the lock then modify the object 🤔 normal case for the usage of lock

for example calling s.AddSubTasks concurrently with same name

dm/master/scheduler/scheduler.go
okJiang: I see. But it’s better to give it a meaningful name also here.

dm/master/scheduler/latch.go
lance6716: seems we can't use a library of GPL-3.0 License

dm/master/server.go
```diff
@@ -450,23 +447,27 @@ func (s *Server) StartTask(ctx context.Context, req *pb.StartTaskRequest) (*pb.S
 		for _, stCfg := range stCfgs {
 			sources = append(sources, stCfg.SourceID)
 		}
-		s.removeMetaLock.Lock()
 		if req.RemoveMeta {
+			release, err3 := s.scheduler.AcquireSubtaskLatch(cfg.Name)
```
Ehco1996, line 451: Will there be a situation where：

Two start-task requests A and B, A requires remove-meta but B does not

t1: A grabs the task's lock (starts remove-meta)
t2: A finishes the remove-meta and releases the lock on the task, but at the same time request B is called in
t2: B grabs the lock on the task and starts execution, but A returns with an exception

Is this not as expected? Perhaps adding a lock to the start of each start-task request would solve this problem?

lance6716: the result is same, meta removed and task started

Ehco1996: but if request A task.yaml is different B task.yaml (only task name is same)?

I know this may rarely happen 😂, But I think we should be more critical?

lance6716: we didn't provide transactional guarantee of DM API (for example, some subtask will fail while other succeeded. so remove meta may success and start task failed) 🤔 can you propose a boundary of API guarantee

Ehco1996: > the result is same, meta removed and task started

Think it over. The result is not really the same
Since A reaches the server first, request A should be fully executed before request B is executed

The expected situation is that when request B reaches the server, it should either return that task is being created, or that task has already been created

Ehco1996: There is no requirement that if a task is created all subtasks are created successfully

But each individual request should not be interrupted by concurrent requests
In the previous implementation, when a remote-meta was required, each start-task would
1. get the remove-meta lock 
2. get master-lock
3. release master-lock
4. release the remove-meta lock

With this implementation, each request is independent and atomic

lance6716:
/hold

waiting for API transactional guarantee

dm/config/subtask.go
```diff
@@ -124,6 +124,26 @@ func (db *DBConfig) Adjust() {
 	AdjustTargetDBTimeZone(db)
 }
 
+// Clone returns a deep copy of DBConfig. This function only fixes data race when adjusting Session.
+func (db *DBConfig) Clone() (DBConfig, error) {
```
glorv, line 128: Shall we just depend on pointer dereference to avoid this kind of special logic and notorious error handling.

lance6716: not understand, pointer dereferencing is  one level shallow copy, at some cases it will cause data race.

Do you mean when we meet data race, we should use lock or adjust the logic to avoid concurrent accessing?

Ehco1996 (approved):
LGTM

dm/config/subtask.go
glorv: I mean we should manually do derefernce for all the pointers. Depend on marshal is some what weird and not ergonomic.

glorv: BWT, Clone an object may return an error is counter-intuitive

dm/master/server.go
```diff
@@ -450,29 +447,46 @@ func (s *Server) StartTask(ctx context.Context, req *pb.StartTaskRequest) (*pb.S
...
+
+		var (
+			latched = false
+			release scheduler.ReleaseFunc
+			err3    error
+		)
+
 		if req.RemoveMeta {
+			// use same latch for remove-meta and start-task
+			release, err3 = s.scheduler.AcquireSubtaskLatch(cfg.Name)
```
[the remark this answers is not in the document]
lance6716: now rely on manually retry on user side, to try to quickly respond 😂

dm/config/task.go
```diff
@@ -714,8 +714,16 @@ func (c *TaskConfig) SubTaskConfigs(sources map[string]DBConfig) ([]*SubTaskConf
 		cfg.HeartbeatReportInterval = c.HeartbeatReportInterval
 		cfg.Meta = inst.Meta
 
-		cfg.From = dbCfg
-		cfg.To = *c.TargetDB
+		fromClone := dbCfg.Clone()
+		if fromClone == nil {
```
Ehco1996, line 718: maybe check (dbCfg!=nil) before clone looks a little more reasonable

dm/config/task.go
```diff
@@ -714,8 +714,16 @@ func (c *TaskConfig) SubTaskConfigs(sources map[string]DBConfig) ([]*SubTaskConf
...
+			return nil, terror.ErrConfigMySQLInstNotFound
+		}
+		cfg.From = *fromClone
+		toClone := c.TargetDB.Clone()
+		if toClone == nil {
```
Ehco1996, line 723: ditto

dm/config/task.go
lance6716: these checking are placed before pointer dereference, in case of `Clone` has a bug.

dm/config/subtask_test.go
```diff
@@ -168,3 +170,33 @@ func (t *testConfig) TestSubTaskBlockAllowList(c *C) {
...
+		Password: "123",
+		Session:  map[string]string{"1": "1"},
+		RawDBCfg: DefaultRawDBConfig(),
+	}
+
+	// When add new fields, also update this value
+	c.Assert(reflect.Indirect(reflect.ValueOf(a)).NumField(), Equals, 8)
+
+	b := a.Clone()
+	c.Assert(a, DeepEquals, b)
```
glorv, line 188: Please also check that the inner pointer are point to different addresses

dm/master/server.go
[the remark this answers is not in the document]
glorv: If we output an error with the message something like "conflict with other commend please retry ...", and the user need to do nothing but retry,  then is it better to automatically retry by the server. 🤔

glorv (commented):
rest LGTM

dm/master/server.go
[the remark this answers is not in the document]
lance6716: If we should let the server perform a retry, it is equal to we use a blocking lock (maybe with timeout) rather than fail-fast lock. I prefer implement `Acquire` in future PRs (now we only have `TryAcquire`)

dm/config/subtask_test.go
lance6716: https://github.com/pingcap/dm/pull/1955/commits/d2a4d6fcedb6c3e8d858a9c9bc89d9f0e026a9b6

lance6716:
ping @glorv

glorv (approved):

lance6716:
Error 1105: item 1 is not in enum [a b c]"

🤔

Issue #1482, opened 2021-03-05
Check important environment variables


overvenus:
## Feature Request

Some important environment variables  affects BR and TiDB-Lightning, eg. http_proxy which may break gRPC connection.

The following link is how TiKV checks environment variables, we can do the same.

https://github.com/tikv/tikv/blob/2cee812f236e3134c9df8b37e73ac51a5f90690f/components/tikv_util/src/lib.rs#L509

lance6716:
specifically, 
- we could add a WARN log when proxies are set in DM master and worker
- we could print warning message to stdout when using dmctl, and this could be suppressed by a flag

lance6716:
https://github.com/pingcap/ticdc/pull/1492

PR #1960, opened 2021-08-04
cmd: add warning message when grpc proxy address env exists

db-will:
#1482

Add warning log to dm-master, dm-worker and dm-ctl when grpc proxy address environment variables are set(HTTP_PROXY, HTTPS_PROXY, NO_PROXY)

### Check List 

Tests 

 - Unit test: pkg/utils/util_test.go
 - Integration test: tests/http_proxies

Related changes


tests/http_proxies/run.sh
```diff
@@ -0,0 +1,51 @@
...
+	# check dm master log
+	check_log_contains "$WORK_DIR/master/log/dm-master.log" $expected_str 1
+
+	# rum dm worker
+	run_dm_worker $WORK_DIR/worker1 $WORKER1_PORT $cur/conf/dm-worker1.toml
+	check_rpc_alive $cur/../bin/check_worker_online 127.0.0.1:$WORKER1_PORT
+
+	# check dm worker log
+	check_log_contains "$WORK_DIR/worker1/log/dm-worker.log" $expected_str 1
+
```
lance6716, line 30: a more troublesome scenario is `dmctl` has set a wrong proxy environment, so user can't control the DM cluster. we might also check `dmctl` here.

and maybe it's hard to find the log file of `dmctl`, we can also print the warning message to stdout as an alternative.

lance6716 (commented):
new tests should add a line of its name in `others_integration.txt`, because we run test based on some hard-coded names in CI script and content of this txt file.

tests/http_proxies/run.sh
```diff
@@ -0,0 +1,51 @@
...
+	unset $env_name
+}
+
+function run() {
+	echo "test dm grpc proxy env setting checking for http_proxy=http://127.0.0.1:8080"
+	test_dm "http_proxy" "http://127.0.0.1:8080"
```
okJiang, line 36: The issue seems to mean that grpc will use the system default environment variables.

So, maybe you can obtain them and input them here?

lance6716: he set environment variables in this test 🤔

okJiang: Sorry, I didn't see it just now....

pkg/utils/util_test.go
```diff
@@ -197,3 +198,34 @@ func (t *testUtilsSuite) TestIsBuildInSkipDDL(c *C) {
...
 	c.Assert(IsBuildInSkipDDL("DROP PROCEDURE"), IsTrue)
 }
+
+func (t *testUtilsSuite) TestProxyFields(c *C) {
+	revIndex := map[string]int{
+		"http_proxy":  0,
+		"https_proxy": 1,
+		"no_proxy":    2,
+	}
+	envs := [...]string{"http_proxy", "https_proxy", "no_proxy"}
```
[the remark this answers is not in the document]
db-will: `[...]` is short for creating array without specifying len of array.

tests/http_proxies/run.sh
db-will: the log file is actually hardcoded in `cmd/dm-ctl/main.go` as `/dmctl.log`

pkg/utils/util_test.go
```diff
@@ -197,3 +198,34 @@ func (t *testUtilsSuite) TestIsBuildInSkipDDL(c *C) {
...
+		for i := 0; i < 3; i++ {
+			if (1<<i)&mask != 0 {
+				c.Assert(os.Setenv(envs[i], envPreset[i]), IsNil)
+			}
+		}
+
+		for _, field := range proxyFields() {
+			idx, ok := revIndex[field.Key]
+			c.Assert(ok, IsTrue)
+			c.Assert((1<<idx)&mask, Not(Equals), 0)
```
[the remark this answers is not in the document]
lance6716: I have checked that they are both `int` type. It's OK.

lance6716 (approved):

pkg/utils/util.go
```diff
@@ -28,7 +28,11 @@ import (
 	"github.com/pingcap/tidb/errno"
 
 	"github.com/pingcap/dm/dm/pb"
+	"github.com/pingcap/dm/pkg/log"
 	"github.com/pingcap/dm/pkg/terror"
+
+	"go.uber.org/zap"
+	"golang.org/x/net/http/httpproxy"
```
lichunzhu, line 35: merge these with L24~28?

lichunzhu (commented):
Rest LGTM
Should we print warning message to stdout for `dmctl`?

glorv (commented):
LGTM after remaining comments are resolves

lichunzhu (approved):
LGTM

PR #1975, opened 2021-08-10
syncer: schema tracker support track skipped ddl

okJiang:
close https://github.com/pingcap/dm/issues/1717

track ddl when ignored it.

### Check List 

Tests 

 - Integration test

Related changes


syncer/filter.go
```diff
@@ -29,26 +29,27 @@ import (
 // - any schema of table names is system schema.
 // - any table name doesn't pass block-allow list.
 // - type of SQL doesn't pass binlog filter.
-func (s *Syncer) skipQuery(tables []*filter.Table, stmt ast.StmtNode, sql string) (bool, error) {
+// skipQuery return (ignored, isBaList, error).
```
lance6716, line 32: isBaList means in block list or allow list?

syncer/filter.go
```diff
@@ -29,26 +29,27 @@ import (
...
+func (s *Syncer) skipQuery(tables []*filter.Table, stmt ast.StmtNode, sql string) (bool, bool, error) {
 	if utils.IsBuildInSkipDDL(sql) {
-		return true, nil
+		return true, false, nil
 	}
 
 	for _, table := range tables {
 		if filter.IsSystemSchema(table.Schema) {
```
lance6716, line 39: seems this should be treated as "default block list"

syncer/ddl.go
```diff
@@ -151,13 +153,38 @@ func (s *Syncer) splitAndFilterDDL(
...
+		shouldSkip, isBaList, err2 := s.skipQuery(tableNames, stmt2, sql)
 		if err2 != nil {
 			return nil, nil, err2
 		}
 		if shouldSkip {
 			metrics.SkipBinlogDurationHistogram.WithLabelValues("query", s.cfg.Name, s.cfg.SourceID).Observe(time.Since(ec.startTime).Seconds())
-			ec.tctx.L().Warn("skip event", zap.String("event", "query"), zap.String("statement", sql), zap.String("schema", schema))
+			ec.tctx.L().Warn("skip event", zap.String("event", "query"), zap.String("sql", sql), zap.String("schema", schema), zap.Any("tableNames", tableNames))
+
+			if s.cfg.ShardMode == config.ShardOptimistic || isBaList || ec.currentLocation == nil {
```
lance6716, line 164: what's the reason we skip for `ShardOptimistic` and `currentLocation` 🤔

syncer/filter.go
```diff
@@ -29,26 +29,27 @@ import (
 // - any schema of table names is system schema.
 // - any table name doesn't pass block-allow list.
 // - type of SQL doesn't pass binlog filter.
-func (s *Syncer) skipQuery(tables []*filter.Table, stmt ast.StmtNode, sql string) (bool, error) {
+// skipQuery return (ignored, isBaList, error).
+func (s *Syncer) skipQuery(tables []*filter.Table, stmt ast.StmtNode, sql string) (bool, bool, error) {
 	if utils.IsBuildInSkipDDL(sql) {
-		return true, nil
+		return true, false, nil
```
lance6716, line 35: because we will `trackDDL` for false isBaList, return false here will cause schema tracker execute a not supported DDL.

I recommend change `isBaList` to `needTrack` and adjust the logic of returning value.

syncer/ddl.go
```diff
@@ -151,13 +153,38 @@ func (s *Syncer) splitAndFilterDDL(
...
+			}
+
+			if err = s.trackDDL(schema, sql, [][]*filter.Table{tableNames, targetTableNames}, stmt2, &ec); err != nil {
+				return nil, nil, err
+			}
+
+			ec.tctx.L().Info("save table checkpoint for source", zap.String("event", "query"), zap.String("source", ec.traceSource), zap.Stringer("start location", *ec.startLocation), log.WrapStringerField("end location", *ec.currentLocation))
+			s.saveGlobalPoint(*ec.currentLocation)
+
+			if err = s.flushCheckPoints(); err != nil {
```
lance6716, line 185: we may need to add a flush job, to wait existing jobs executed. Because we saved the global checkpoint at above line but the jobs before that checkpoint may not be executed

syncer/filter.go
okJiang: Whatever block list or allow list may lead to ignore this query event.

syncer/ddl.go
okJiang: `currentLocation` is to pass all unit test. The reason we skip `currentLocation` is that we all do not initialize `ec.currentLocation` in unit test.

okJiang: After skipping `isBaList(needTrack)`, I find we can dismiss `ShardOptimistic` check. I will remove it in next commit.

tests/tracker_ignored_ddl/run.sh
```diff
@@ -0,0 +1,59 @@
...
+	check_sync_diff $WORK_DIR $cur/conf/diff_config.toml
+
+	run_sql_file $cur/data/db.increment1.sql $MYSQL_HOST1 $MYSQL_PORT1 $MYSQL_PASSWORD1
+	run_sql_tidb "show create table $TEST_NAME.t1;"
+	check_not_contains "ignore_1"
+	echo "increment1 check success"
+
+	run_sql_file $cur/data/db.increment2.sql $MYSQL_HOST1 $MYSQL_PORT1 $MYSQL_PASSWORD1
+	run_dm_ctl_with_retry $WORK_DIR "127.0.0.1:$MASTER_PORT" \
+		"query-status test" \
```
Ehco1996, line 33: I'm curious if this solves the problem at all
It just changes the error message from `count unmatch errors` to `Unknown column`.

Maybe we can go a step further and keep the sync task from being interrupted?

If a column is added upstream (which is ignored), the DM generated DML should also ignore that column

If a column is subtracted upstream (which is ignored), the DM generated DML should also take the default value of that column

also cc @lance6716  @lichunzhu

okJiang: also cc @glorv

syncer/ddl.go
```diff
@@ -153,15 +152,15 @@ func (s *Syncer) splitAndFilterDDL(
...
+		shouldSkip, needTrack, err2 := s.skipQuery(tableNames, stmt2, sql)
 		if err2 != nil {
 			return nil, nil, err2
 		}
 		if shouldSkip {
 			metrics.SkipBinlogDurationHistogram.WithLabelValues("query", s.cfg.Name, s.cfg.SourceID).Observe(time.Since(ec.startTime).Seconds())
 			ec.tctx.L().Warn("skip event", zap.String("event", "query"), zap.String("sql", sql), zap.String("schema", schema), zap.Any("tableNames", tableNames))
 
-			if s.cfg.ShardMode == config.ShardOptimistic || isBaList || ec.currentLocation == nil {
+			if !needTrack || ec.currentLocation == nil {
```
lance6716, line 163: better to explain the reason of checking `ec.currentLocation == nil` in comment

syncer/ddl.go
```diff
@@ -40,7 +41,7 @@ func (s *Syncer) parseDDLSQL(sql string, p *parser.Parser, schema string) (resul
 	sql = utils.TrimCtrlChars(sql)
 
 	// check skip before parse (used to skip some un-supported DDLs)
-	ignore, err := s.skipQuery(nil, nil, sql)
+	ignore, _, err := s.skipQuery(nil, nil, sql)
```
lance6716, line 44: if here a binlog filter using RE to filter a SQL, `splitAndFilterDDL` will not be called and we won't track that DDL

okJiang: What does RE means?

tests/tracker_ignored_ddl/run.sh
Ehco1996: there's already a lot of content in this PR, so maybe we can implement this in next PR

syncer/filter.go
```diff
@@ -58,26 +60,26 @@ func (s *Syncer) skipQuery(tables []*filter.Table, stmt ast.StmtNode, sql string
 	if len(tables) == 0 {
 		action, err := s.binlogFilter.Filter("", "", et, sql)
 		if err != nil {
-			return false, terror.Annotatef(terror.ErrSyncerUnitBinlogEventFilter.New(err.Error()), "skip query %s", sql)
+			return false, false, terror.Annotatef(terror.ErrSyncerUnitBinlogEventFilter.New(err.Error()), "skip query %s", sql)
 		}
 
 		if action == bf.Ignore {
-			return true, nil
+			return true, true, nil
```
Ehco1996, line 67: I'm not sure if this will cause us to track statements that shouldn't be tracked

For example, `truncate table` `rename table`

okJiang:
Wait refined.
/hold

tests/tracker_ignored_ddl/run.sh
glorv: > If a column is subtracted upstream (which is ignored), the DM generated DML should also take the default value of that column
@Ehco1996  We don't support this kind of operation and don't have plan to support this kind of operation at least now. Since this kind of operations will lead to data inconsistency.  And #1717 is not related to it either, the problem is that even if the down stream is compatible with upstream, e.g. user manually changed the schema, dm still reported an error which is not acceptable. 

So I think this PR is ok.

glorv: I think we can add the alter table ddl in the script before executing the incremental phase, then the task should finish successfully.

PR #1981, opened 2021-08-11
tests: fix unstable tests

lance6716:
part of #1790 

- don't check result of resume. Because if worker runs quickly, resume-task will see the expected error of next query-status and not return "result: true"
- wait DM-worker clean dump files and then using `ls` to check files
- also check master peer port before restart master, to avoid port in use

### Check List 

Tests 

 - No code

Related changes


tests/sharding/run.sh
```diff
@@ -80,17 +80,15 @@ function run() {
...
+		"resume-task test"
 
 	run_dm_ctl_with_retry $WORK_DIR "127.0.0.1:$MASTER_PORT" \
 		"query-status test" \
 		"failpoint error for FlushCheckpointStage before track DDL" 1
 
 	# resume-task to next stage
 	run_dm_ctl $WORK_DIR "127.0.0.1:$MASTER_PORT" \
-		"resume-task test" \
-		"\"result\": true" 3
```
Ehco1996, line 93: why don't you need to test the number of `result: true`
Is it because the following is tested with `dmctl query-status`?

lance6716: see PR description

Ehco1996 (approved):

GMHDBJD (approved):
LGTM

Issue #1983, opened 2021-08-11
evict leader can't take effect


lance6716:
## Bug Report

re-explain of https://github.com/pingcap/dm/issues/1212

Please answer these questions before submitting your issue. Thanks!

1. What did you do? If possible, provide a recipe for reproducing the error.

evict leader

2. What did you expect to see?

evict success

3. What did you see instead?

leader can't be evicted

4. Root cause

[dm-ci.log.zip](https://github.com/pingcap/dm/files/6967781/dm-ci.log.zip)

in the log we can see

the current election key (consist of a user specified prefix and etcd session lease) is `/dm-master/leader/690e7b3305c06104`
```
[2021/08/11 10:23:46.449 +08:00] [DEBUG] [interceptor.go:181] ["request stats"] [component="embed etcd"] ["start time"=2021/08/11 10:23:46.441 +08:00] ["time spent"=7.229401ms] [remote=127.0.0.1:50072] ["response type"=/etcdserverpb.KV/Txn] ["request count"=1] ["request size"=90] ["response count"=0] ["response size"=38] ["request content"="compare:<target:CREATE key:\"/dm-master/leader/690e7b3305c06104\" create_revision:0 > success:<request_put:<key:\"/dm-master/leader/690e7b3305c06104\" value_size:40 lease:7570123482726424836 >> failure:<request_range:<key:\"/dm-master/leader/690e7b3305c06104\" > >"]
```

last election ends gracefully
```
[2021/08/11 10:23:53.483 +08:00] [INFO] [election.go:284] ["fail to campaign"] [component=election] ["current member"="{\"id\":\"master3\",\"addr\":\"localhost:8461\"}"] [error="context canceled"]
```

start a new election
```
[2021/08/11 10:23:53.483 +08:00] [DEBUG] [election.go:266] ["begin to campaign"] [component=election] ["current member"="{\"id\":\"master3\",\"addr\":\"localhost:8461\"}"]
```

this new election failed, and the inner recorded election key is empty
```
[2021/08/11 10:24:00.488 +08:00] [DEBUG] [election.go:271] ["before manually resign"] [component=election] ["current election key"=] ["current election header"=<nil>]
[2021/08/11 10:24:00.488 +08:00] [DEBUG] [election.go:279] ["after manually resign"] [component=election] ["current election key"=] ["current election header"=<nil>]
[...]

PR #1986, opened 2021-08-12
election: always try to clean last election key

lance6716:
close https://github.com/pingcap/dm/issues/1983

when `Campaign` returned an error, always try to clean last election key

### Check List 

Tests 

 - hard to test

Related changes

### DIFF
--- a/pkg/election/election.go
+++ b/pkg/election/election.go
@@ -81,7 +81,6 @@ type Election struct {
 	// so do not close it in the methods of Election.
 	cli        *clientv3.Client
 	sessionTTL int
-	session    *concurrency.Session
 	key        string
 
 	info    *CampaignerInfo
--- a/pkg/election/election.go
+++ b/pkg/election/election.go
@@ -81,6 +81,7 @@ type Election struct {
 	// so do not close it in the methods of Election.
 	cli        *clientv3.Client
 	sessionTTL int
+	session    *concurrency.Session
 	key        string
 
 	info    *CampaignerInfo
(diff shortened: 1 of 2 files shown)


tests/ha_master/run.sh
```diff
@@ -58,14 +58,8 @@ function test_evict_leader() {
...
 		fi
-		# Leader evict is not effective immediately, we need to wait for a proper period of time.
-		for _ in {0..30}; do
-			NEW_LEADER_NAME=$(get_leader "$WORK_DIR" 127.0.0.1:${MASTER_PORT1})
-			if [ "$NEW_LEADER_NAME" != "$LEADER_NAME" ]; then
-				break
-			fi
-			sleep 1
-		done
+		NEW_LEADER_NAME=$(get_leader $WORK_DIR 127.0.0.1:${MASTER_PORT1})
```
lance6716, line 61: revert #1946

glorv (approved):

tests/ha_master/run.sh
gozssky: https://github.com/pingcap/dm/pull/1946 shouldn't be reverted? What it focus on is different. The evict request is asynchronous, so new leader may not be elected immediately after the request returns.

pkg/election/election.go
```diff
@@ -267,18 +267,14 @@ func (e *Election) campaignLoop(ctx context.Context, session *concurrency.Sessio
...
-				// sometimes we noticed auto Resign does not work, so we try to manually Resign.
-				e.l.Debug("before manually resign",
-					zap.String("current election key", elec.Key()),
-					zap.Any("current election header", elec.Header()))
-				resignCtx, cancel := context.WithTimeout(ctx, 5*time.Second)
-				defer cancel()
-				if err3 := elec.Resign(resignCtx); err3 != nil {
-					e.l.Warn("failed to manually resign", zap.Error(err3))
+				// because inner commit may return undetermined error, we try to delete the election key manually
+				deleted, err3 := e.ClearSessionIfNeeded(ctx, e.ID())
```
Ehco1996, line 271: Can you please tell me why I need to delete this key manually ?

and in old logic, we do `Resign` if `Campaign` failed, after this change ,why don't we do this anymore?

lance6716: see the issue in PR description

Ehco1996 (approved):

Issue #1982, opened 2021-08-11
Develop OpenAPI for DM


Ehco1996:
## Feature Request

**Is your feature request related to a problem? Please describe:**

To enhance DM's operations and maintenance capabilities, DM needs to support a fully functional OpenAPI
I don't know why, but I guess we didn't put this configuration option out in the documentation before

**Describe the feature you'd like:**

- [x]  API is simple, unified and easy to use.
- [x] Provide complete API usage documents.
- [x]  Provide Integration test and be able to conduct complete function test on API
- [x]  Provide unit test and be able to conduct complete function test on API
- [x]  Automatically forward the request to the master

**Related job content**

- [x] Support for uploading the contents of tls certificates via the API
- [x] Compliant with openapi 3.0 spec and documentation
- [x] Source Related APIs
- [x] Relay Log related APIs
- [x] Task-related APIs
- [x] ...

**Related job V2 content**

- [x] New part of openapi 3.0 spec and documentation
- [x] Cluster Related APIs
- [ ] https://github.com/pingcap/dm/pull/2209
- [ ] Task  related APIs
- [ ] doc of all APIs

**Teachability, Documentation, Adoption, Migration Strategy:**

* [openapi 3.0 doc](https://app.swaggerhub.com/apis/Ehco19969/DM-OpenAPI/2.1.0#/)
* [design doc](https://docs.google.com/document/d/1GNOR1a5ictuen-8zBGaQRTx2yOCiR_bJAi14zyt50sg/edit)

PR #1999, opened 2021-08-17
openapi: init openapi codebase

Ehco1996:
#1982  
* Part2 - Compliant with openapi 3.0 spec and documentation

1. Prepare an openAPI 3.0 compliant yaml file
2. The yaml file is used to generate the structs needed at api level, such as the request and return bodies
3. Start a [echo](https://github.com/labstack/echo) http server on dm-master with a **new port**

### Check List 

Tests 

 - Unit test

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes


dm/master/dm-master.toml
```diff
@@ -10,6 +10,9 @@ master-addr = ":8261"
 # addr(i.e. 'host:port') to advertise to the public
 advertise-addr = "127.0.0.1:8261"
 
+# openapi-addr(i.e. 'host:port') is dm open api server listen address
```
Ehco1996, line 13: openapi

dm/master/openapi.go
```diff
@@ -0,0 +1,171 @@
...
+	if err != nil {
+		exitServer(err)
+	}
+	swagger.AddServer(&openapi3.Server{URL: fmt.Sprintf("http://%s", s.cfg.OpenAPIAddr)})
+	swaggerJSON, err := swagger.MarshalJSON()
+	if err != nil {
+		exitServer(err)
+	}
+	docMW := openapi.NewSwaggerDocUI(openapi.NewSwaggerConfig(docBasePath, docJSONBasePath, ""), swaggerJSON)
+	// Echo instance
```
Ehco1996, line 55: Uniform case for comments

dm/master/server.go
```diff
@@ -458,7 +467,12 @@ func (s *Server) StartTask(ctx context.Context, req *pb.StartTaskRequest) (*pb.S
 				s.removeMetaLock.Unlock()
 				return resp, nil
 			}
-			err = s.removeMetaData(ctx, cfg)
+			toDB := cfg.TargetDB
+			toDB.Adjust()
+			if len(toDB.Password) > 0 {
+				toDB.Password = utils.DecryptOrPlaintext(toDB.Password)
+			}
+			err = s.removeMetaData(ctx, cfg.Name, cfg.MetaSchema, *toDB)
```
Ehco1996, line 475: The early change here is due to the need to call the RemoveMeta method when creating a task via the API later
https://github.com/Ehco1996/dm/blob/7c01c9cacf60d882548e0bbcc2486d444c9bace2/dm/master/openapi.go#L339

dm/master/server_test.go
```diff
@@ -175,8 +175,16 @@ func TestMaster(t *testing.T) {
...
-	s.testT = t
+	s, ok := testSuite.(*testMaster)
+	if ok {
+		s.testT = t
+	}
+
+	// inject *testing.T to openAPISuite
+	os, ok := ts.(*openAPISuite)
+	if ok {
+		os.testT = t
```
Ehco1996, line 186: This is because the unit tests that follow will need to use the 
`testing.T`

https://github.com/Ehco1996/dm/blob/7c01c9cacf60d882548e0bbcc2486d444c9bace2/dm/master/openapi_test.go#L811

pkg/election/election.go
```diff
@@ -54,13 +54,15 @@ type CampaignerInfo struct {
 	ID string `json:"id"`
 	// addr is the campaigner's advertise address
 	Addr string `json:"addr"`
+	// openAddr is the campaigner's advertise address
```
Ehco1996, line 57: typo

tools/go.mod
```diff
@@ -3,6 +3,8 @@ module github.com/pingcap/dm/tools
 go 1.16
 
 require (
+	github.com/deepmap/oapi-codegen v1.8.2 // indirect
```
Ehco1996, line 6: fix indirect

dm/master/config.go
```diff
@@ -111,6 +111,7 @@ type Config struct {
 
 	MasterAddr    string `toml:"master-addr" json:"master-addr"`
 	AdvertiseAddr string `toml:"advertise-addr" json:"advertise-addr"`
+	OpenAPIAddr   string `toml:"openapi-addr" json:"openapi-addr"`
```
glorv, line 114: Can we just share the same address with `advertise-addr`

dm/master/server_test.go
```diff
@@ -175,8 +175,16 @@ func TestMaster(t *testing.T) {
 		t.Fatal(err)
 	}
 	// inject *testing.T to testMaster
-	s := testSuite.(*testMaster)
-	s.testT = t
+	s, ok := testSuite.(*testMaster)
+	if ok {
```
glorv, line 179: How can this type cast fail, I think this check is redundant

dm/master/config.go
Ehco1996: yes  and I think it's a really good idea! fixed in https://github.com/pingcap/dm/pull/1999/commits/4536f64657cbbadbbf9bd13edcc2e521edc1bacf

dm/master/server_test.go
Ehco1996: in L184, `s` possibly be `openAPISuite`

glorv: No. But the `testSuite` and `ts` are never mistaken with each other.

Ehco1996: 😬 yes, my bad , fixed in https://github.com/pingcap/dm/pull/1999/commits/a6935915d18e82d776e4cad867cc8bca62ef14a1

dm/master/server.go
```diff
@@ -1395,29 +1408,23 @@ func withHost(addr string) string {
...
 }
 
-func (s *Server) removeMetaData(ctx context.Context, cfg *config.TaskConfig) error {
-	toDB := *cfg.TargetDB
-	toDB.Adjust()
-	if len(toDB.Password) > 0 {
-		toDB.Password = utils.DecryptOrPlaintext(toDB.Password)
-	}
-
+func (s *Server) removeMetaData(ctx context.Context, taskName, metaSchema string, toDBCfg config.DBConfig) error {
```
lance6716, line 1411:
```suggestion
func (s *Server) removeMetaData(ctx context.Context, taskName, metaSchema string, toDBCfg *config.DBConfig) error {
```

lance6716: can reduce the anooying `*` when invoking.

dm/master/server.go
lance6716: DecryptOrPlaintext can still moved inside removeMetaData

dm/master/openapi_test.go
```diff
@@ -0,0 +1,82 @@
...
+	"github.com/pingcap/check"
+	"github.com/tikv/pd/pkg/tempurl"
+
+	"github.com/pingcap/dm/pkg/utils"
+)
+
+var openAPITestSuite = check.Suite(&openAPISuite{})
+
+type openAPISuite struct {
+	testT *testing.T
```
lance6716, line 33: where is it used?

openapi/spec/dm.yaml
```diff
@@ -0,0 +1,876 @@
+openapi: "3.0.0"
+info:
+  title: DM OpenAPI 文档
```
lance6716, line 3: no chinese

dm/master/openapi_test.go
Ehco1996: it will be used in next[ PR ](https://github.com/Ehco1996/dm/blob/7c01c9cacf60d882548e0bbcc2486d444c9bace2/dm/master/openapi_test.go#L811)

dm/master/openapi.go
```diff
@@ -0,0 +1,154 @@
...
+	swagger, err := openapi.GetSwagger()
+	if err != nil {
+		exitServer(err)
+	}
+	swagger.AddServer(&openapi3.Server{URL: fmt.Sprintf("http://%s", s.cfg.AdvertiseAddr)})
+	swaggerJSON, err := swagger.MarshalJSON()
+	if err != nil {
+		exitServer(err)
+	}
+	docMW := openapi.NewSwaggerDocUI(openapi.NewSwaggerConfig(docBasePath, docJSONBasePath, ""), swaggerJSON)
```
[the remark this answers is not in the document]
Ehco1996: `echo.MiddlewareFunc` don't  allow return error and all mw funcs  must checked before server start 😂

Makefile
```diff
@@ -90,6 +90,11 @@ generate_proto: tools_setup
 generate_mock: tools_setup
 	./tests/generate-mock.sh
 
+generate_openapi: tools_setup
+	@echo "generate_openapi"
+	tools/bin/oapi-codegen --config=openapi/spec/server-gen-cfg.yaml openapi/spec/dm.yaml
```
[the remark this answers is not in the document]
Ehco1996: `oapi` does not yet offer such functionality
the two configuration files are used to generate two separate files for the` http server `related code and the `request/response` related code

_utils/terror_gen/errors_release.txt
```diff
@@ -514,4 +514,6 @@ ErrSchedulerSourceOpRelayExist,[code=46023:class=scheduler:scope=internal:level=
 ErrCtlGRPCCreateConn,[code=48001:class=dmctl:scope=internal:level=high], "Message: can not create grpc connection, Workaround: Please check your network connection."
 ErrCtlInvalidTLSCfg,[code=48002:class=dmctl:scope=internal:level=medium], "Message: invalid TLS config, Workaround: Please check the `ssl-ca`, `ssl-cert` and `ssl-key` config in command line."
 ErrCtlLoadTLSCfg,[code=48003:class=dmctl:scope=internal:level=high], "Message: can not load tls config, Workaround: Please ensure that the tls certificate is accessible on the node currently running dmctl."
+ErrOpenAPICommonError,[code=49001:class=unknown "error class: 19:scope=internal:level=high], Message: some unforeseen errors have occurred, please check the detailed error message"
```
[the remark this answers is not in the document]
Ehco1996: I have used the terror module incorrectly
I only added `ClassOpenAPI` but not the corresponding description in `errClass2Str`.  👍

dm/master/server.go
Ehco1996: can we move `DecryptOrPlaintext` into `DBConfig.Adjust` ?

dm/master/openapi.go
```diff
@@ -0,0 +1,154 @@
...
+		c.Logger().Error(sendErr)
+	}
+}
+
+func sendHTTPErrorResp(ctx echo.Context, code int, message string) error {
+	err := openapi.ErrorWithMessage{ErrorMsg: message, ErrorCode: code}
+	return ctx.JSON(http.StatusBadRequest, err)
+}
+
+func exitServer(err error) {
```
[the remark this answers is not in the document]
Ehco1996: fixed in https://github.com/pingcap/dm/pull/1999/commits/8ca550e080f06a33d6d93314baad7152780687ed

dm/master/server.go
Ehco1996: fixed in https://github.com/pingcap/dm/pull/1999/commits/779794675eaa37084ab15e1b70447bbfe4538257

tests/_utils/test_prepare
```diff
@@ -11,8 +11,16 @@ function cleanup_data() {
 }
 
 function cleanup_process() {
-    pkill -hup dm-worker.test 2>/dev/null || true
+    dm_master_num=$(ps aux >temp && grep "dm-master.test" temp | wc -l && rm temp)
```
Ehco1996, line 14: print some debug msg

dm/master/openapi.go
```diff
@@ -0,0 +1,152 @@
...
+		msg = err.Error()
+	}
+	if sendErr := sendHTTPErrorResp(c, code, msg); sendErr != nil {
+		c.Logger().Error(sendErr)
+	}
+}
+
+func sendHTTPErrorResp(ctx echo.Context, code int, message string) error {
+	err := openapi.ErrorWithMessage{ErrorMsg: message, ErrorCode: code}
+	return ctx.JSON(http.StatusBadRequest, err)
```
lance6716, line 144: maybe we should use 5xx because this is server side error?

dm/master/server.go
```diff
@@ -25,6 +25,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/labstack/echo/v4"
```
lance6716, line 28: will this library be used for demonstrating only in this PR, or used from now on?

dm/master/server.go
```diff
@@ -458,7 +462,9 @@ func (s *Server) StartTask(ctx context.Context, req *pb.StartTaskRequest) (*pb.S
 				s.removeMetaLock.Unlock()
 				return resp, nil
 			}
-			err = s.removeMetaData(ctx, cfg)
+			toDB := cfg.TargetDB
+			toDB.Adjust()
```
lance6716, line 466: `Adjust` can also be moved into `removeMetaData`. And please check we should clone the toDB when `Adjust`, to avoid unexpected changes.

dm/master/server.go
```diff
@@ -1094,31 +1100,32 @@ func parseAndAdjustSourceConfig(ctx context.Context, contents []string) ([]*conf
...
 
+func checkAndAdjustSourceConfig(ctx context.Context, cfg *config.SourceConfig) error {
+	dbConfig := cfg.GenerateDBConfig()
+	fromDB, err := conn.DefaultDBProvider.Apply(*dbConfig)
+	if err != nil {
+		return err
+	}
+	defer fromDB.Close()
+	if err = cfg.Adjust(ctx, fromDB.DB); err != nil {
+		fromDB.Close()
```
lance6716, line 1119: no need because we have defered, and remove another one some lines below

pkg/terror/error_list.go
```diff
@@ -1245,6 +1251,10 @@ var (
 	ErrCtlInvalidTLSCfg  = New(codeCtlInvalidTLSCfg, ClassDMCtl, ScopeInternal, LevelMedium, "invalid TLS config", "Please check the `ssl-ca`, `ssl-cert` and `ssl-key` config in command line.")
 	ErrCtlLoadTLSCfg     = New(codeCtlLoadTLSCfg, ClassDMCtl, ScopeInternal, LevelHigh, "can not load tls config", "Please ensure that the tls certificate is accessible on the node currently running dmctl.")
 
+	// openapi.
+	ErrOpenAPICommonError        = New(codeOpenAPICommon, ClassOpenAPI, ScopeInternal, LevelHigh, "some unforeseen errors have occurred, please check the detailed error message", "")
```
lance6716, line 1255:
```suggestion
	ErrOpenAPICommonError        = New(codeOpenAPICommon, ClassOpenAPI, ScopeInternal, LevelHigh, "some unexpected errors have occurred, please check the detailed error message", "")
```

openapi/logger.go
```diff
@@ -0,0 +1,52 @@
...
+				zap.String("remote_ip", c.RealIP()),
+				zap.String("user_agent", req.UserAgent()),
+				zap.String("duration", time.Since(start).String()),
+				zap.String("request", fmt.Sprintf("%s %s", req.Method, req.RequestURI)),
+			}
+
+			n := res.Status
+			switch {
+			case n >= 500:
+				log.With(zap.Error(err)).Error("Server error", fields...)
```
lance6716, line 41: do we still need to print the error? line 20 has called it

openapi/logger_test.go
```diff
@@ -0,0 +1,47 @@
...
+	ctx := e.NewContext(req, rec)
+	h := func(ctx echo.Context) error {
+		return ctx.String(http.StatusOK, "")
+	}
+	obs, logs := observer.New(zap.DebugLevel)
+	logger := zap.New(obs)
+	err := ZapLogger(logger)(h)(ctx)
+	c.Assert(err, check.IsNil)
+
+	logFields := logs.AllUntimed()[0].ContextMap()
```
lance6716, line 35: seems `All()` is enough

tests/_utils/test_prepare
lance6716: why we need temp not directly use pipe?

Ehco1996: i copied it from [here](https://github.com/pingcap/dm/blob/04c6778706e62074a225c5c345b933ac8c19ff87/tests/_utils/kill_dm_master#L4)

and I guess it's possible that if you do it directly with a pipe, the grep will always come up with a number greater than one： 
`ps aux | grep "dm-master.test"
ehco             10542   0.0  0.0 408112528   1408 s000  R+    3:12下午   0:00.00 grep --color=auto --exclude-dir=.bzr --exclude-dir=CVS --exclude-dir=.git --exclude-dir=.hg --exclude-dir=.svn --exclude-dir=.idea --exclude-dir=.tox dm-master.test`

dm/master/openapi.go
Ehco1996: Most cases here return an error from a failed request parameter check, so I think 4xx is more appropriat
5xx will be returned automatically when the program painc

dm/master/server.go
Ehco1996: it will be used from now

openapi/logger.go
Ehco1996: yes, we need, because L20 is actually call `terrorHTTPErrorHandler`, and we still need record ths http 5xx in dm-master's log

dm/master/server.go
Ehco1996: fixed in https://github.com/pingcap/dm/pull/1999/commits/a98c9a72795f4c6e7cc7f5fee1ad7a7a27c0d229
there is no need to clone a configuration yet, as `cfg.TargetDB` is not used again in the following logic

tests/_utils/test_prepare
lance6716: the `grep` itself is running, so it appears in `ps`

openapi/swaggerui.go
```diff
@@ -0,0 +1,115 @@
...
+	if err != nil {
+		return nil, err
+	}
+	uiHTML := buf.Bytes()
+	return func(next echo.HandlerFunc) echo.HandlerFunc {
+		return func(c echo.Context) error {
+			path := c.Request().URL.Path
+			switch {
+			case path == config.DocPath:
+				return c.HTML(200, string(uiHTML))
```
lance6716, line 108: if we returned here, how can chained `next` be applied?

lance6716 (commented):
rest lgtm

openapi/swaggerui.go
Ehco1996: this is by desigin😬  when the case `path == config.DocPath` is hit, I want to then return the rendered document page

In cases other than this case, the middleware will call `next(c)` to ensure that the other middleware is called in a chain in L112

lance6716: if a "counting for access" handler is added to chain, we will lose counting if don't apply them.

anyway, I think generally we should always propagate the `echo.Context` in the chain. "return the rendered document page(raw html)" doesn't interfere with `next(c)`

Ehco1996: > if a "counting for access" handler is added to chain, we will lose counting if don't apply them.

so user need to add this counter mw befroe doc mw, in other words, this mw is already at the end of the request lifecycle and all it needs to do is write the http reponse back to the client.

If we don't put it in the last middleware of the request, it could happen that the document-related html is written to the response, but this will be overwritten by the subsequent middleware, which I don't think is the state the user wants

lance6716: In some cases HTML should be overwritten/appended, like append a JS script. But I don't think there's a case that chaining should be broken other than we meet a critical error.

Ehco1996: I think a direct return makes sense here, as the scenario is more like an interceptor than other chain mw

To give an example of the official echo jwt [mw](https://echo.labstack.com/middleware/jwt/) 
In the logic of this jwt mw, when an illegal jwt token is encountered, errror=invalid token is returned directly

lance6716: I don't think there's a case that chaining should be broken **other than we meet a critical error**.

Ehco1996: or i can use new http handelr to serve doc?https://github.com/pingcap/dm/commit/62298cb106d211b9b459156d6a96b53e6bdd5600
but this code feels ugly to me

dm/master/openapi.go
```diff
@@ -0,0 +1,153 @@
...
+	swagger.Servers = nil
+	// use our validation middleware to check all requests against the OpenAPI schema.
+	e.Use(middleware.OapiRequestValidator(swagger))
+	openapi.RegisterHandlers(e, s)
+	s.echo = e
+}
+
+// redirectRequestToLeader is used to redirect the request to leader.
+// because the leader has some data in memory, only the leader can process the request.
+func (s *Server) redirectRequestToLeader(ctx context.Context) (needRedirect bool, host string, err error) {
```
lichunzhu, line 69: Will this function be used in next PR?

dm/master/server.go
```diff
@@ -182,10 +184,12 @@ func (s *Server) Start(ctx context.Context) (err error) {
 	// And curl or safari browser does trigger this problem.
 	// But I haven't figured it out.
 	// (maybe more requests are sent from chrome or its extensions).
+	s.InitOpenAPIHandles()
```
lichunzhu, line 187: What about this format?
```suggestion
	s.echo, err = InitOpenAPIHandles()
```

dm/master/openapi.go
```diff
@@ -0,0 +1,153 @@
...
+// DMAPIGetTaskStatus url is:(GET /api/v1/tasks/{task-name}/status).
+func (s *Server) DMAPIGetTaskStatus(ctx echo.Context, taskName string) error {
+	return nil
+}
+
+func terrorHTTPErrorHandler(err error, c echo.Context) {
+	var code int
+	var msg string
+	if tErr, ok := err.(*terror.Error); ok {
+		code = int(tErr.Code())
```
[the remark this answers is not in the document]
Ehco1996: yes, because we want to unify the return body when API occur error, all error will return in this format(ErrorWithMessage)

dm/master/openapi.go
Ehco1996: yes, it will be used like [this](https://github.com/Ehco1996/dm/blob/7c01c9cacf60d882548e0bbcc2486d444c9bace2/dm/master/openapi.go#L109)

dm/master/server.go
Ehco1996: i think `func (s *Server) InitOpenAPIHandles() error` is better, because in L68, i need to call(openapi.RegisterHandlers(e, s))

lichunzhu (approved):
LGTM

Issue #1997, opened 2021-08-16
Quality reinforcement (13): fix unstable test cases


lance6716:
previous one https://github.com/pingcap/dm/issues/1790

- [ ] ha_master https://github.com/pingcap/dm/issues/1762
- [ ] ~testForEtcd.TestWorkerKeepAlive DEAD LOOP [jenkins](https://ci.pingcap.net/blue/organizations/jenkins/dm_ghpr_new_test/detail/dm_ghpr_new_test/1293/pipeline)~
- [x] metrics & metrics causes [jenkins](https://ci.pingcap.net/blue/organizations/jenkins/dm_ghpr_new_test/detail/dm_ghpr_new_test/1681/pipeline/) [jenkins](https://ci.pingcap.net/blue/organizations/jenkins/dm_ghpr_new_test/detail/dm_ghpr_new_test/1745/pipeline/142) [jenkins](https://ci.pingcap.net/blue/organizations/jenkins/dm_ghpr_new_test/detail/dm_ghpr_new_test/1759/pipeline) [jenkins](https://ci.pingcap.net/blue/organizations/jenkins/dm_ghpr_new_test/detail/dm_ghpr_new_test/1773/pipeline) #2003
- [ ] ~syncer unit test [jenkins](https://ci.pingcap.net/blue/organizations/jenkins/dm_ghpr_new_test/detail/dm_ghpr_new_test/1309/pipeline)~
- [x] server_test.go:406 testMaster.TestQueryStatus [jenkins](https://ci.pingcap.net/blue/rest/organizations/jenkins/pipelines/dm_ghpr_new_test/runs/1510/nodes/140/steps/871/log/?start=0)
- [x] testMaster UT [jenkins](https://ci.pingcap.net/blue/organizations/jenkins/dm_ghpr_new_test/detail/dm_ghpr_new_test/1554/pipeline) [jenkins] https://ci.pingcap.net/blue/organizations/jenkins/dm_ghpr_new_test/detail/dm_ghpr_new_test/2025/pipeline
- [ ] testElectionSuite.TestFailToStartLeader
```
[2021-08-10T03:55:24.792Z] election_test.go:91:
[2021-08-10T03:55:24.792Z]     c.Assert(leaderID, check.Equals, cfg1.Name)
[2021-08-10T03:55:24.792Z] ... obtained string = "dm-master-2"
[2021-08-10T03:55:24.792Z] ... expected string = "dm-master-1"
```
[...]

PR #2003, opened 2021-08-17
test: fix unstable metrics test

lance6716:
part of https://github.com/pingcap/dm/issues/1997

- when a task is started some time is needed to `Init` units, only after `Init` the task goes to Running status, so wait more to pass `Init`
- worker is restarted, retry `query-status` to see if error is gone

### Check List 

Tests 

 - No code

Related changes


GMHDBJD (approved):
LGTM

Ehco1996 (approved):

PR #2008, opened 2021-08-17
scheduler: try bind free worker to source in start scheduler

### LINKED ISSUE #2002: The task cannot be transferred when dm-master leader and dm-worker down at the same time
## Bug Report

https://asktug.com/t/topic/123105

the task cannot be transferred to new free dm-worker

3. What did you see instead?
the task should be transferred after new leader elected success

4. Versions of the cluster
DM v2.0.6 and lower

glorv:
close #2002

If some worker are offline during master leader transfer, the offline event will be missing. So scheduler should try to bind free source and worker if any after initilized.

### Check List 

Tests 

 - Unit test
 
Code changes

Side effects

Related changes

### DIFF
--- a/dm/master/scheduler/scheduler.go
+++ b/dm/master/scheduler/scheduler.go
@@ -215,6 +215,19 @@ func (s *Scheduler) Start(pCtx context.Context, etcdCli *clientv3.Client) (err e
 		return err
 	}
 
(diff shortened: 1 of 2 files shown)


lance6716 (approved):

GMHDBJD (approved):
LGTM

PR #2014, opened 2021-08-18
syncer: add retry for binlog event streamer

glorv:
Also allow error retry for remote binlog reader

- For remote binlog puller, allow auto-retry pullEvent error but restrict to only once per specified interval (currently 1min). 

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

Related changes


lance6716 (approved):

syncer/streamer_controller.go
```diff
@@ -476,3 +497,33 @@ func isConnectionRefusedError(err error) bool {
...
+	CanRetry(error) bool
+}
+
+type alwaysRetryStrategy struct{}
+
+func (s alwaysRetryStrategy) CanRetry(error) bool {
+	return true
+}
+
+// maxIntervalRetryStrategy allows retry the duration since last retry is bigger than specified interval.
```
Ehco1996, line 512: maybe this comment is better?

`allow retry when the retry interval is greater than the set interval.`

syncer/streamer_controller.go
```diff
@@ -120,10 +124,30 @@ type StreamerController struct {
...
+	// let local binlog also return error to avoid infinity loop
+	failpoint.Inject("GetEventError", func() {
+		strategy = &maxIntervalRetryStrategy{
+			interval: minErrorRetryInterval,
+		}
+	})
 	streamerController := &StreamerController{
 		initBinlogType:    binlogType,
 		currentBinlogType: binlogType,
+		retryStrategy:     strategy,
```
Ehco1996, line 150: can we use func `NewStreamerController` to init streamerController here ?

tests/incremental_mode/run.sh
```diff
@@ -168,7 +169,8 @@ function run() {
 	check_port_offline $WORKER1_PORT 20
 	check_port_offline $WORKER2_PORT 20
 
-	export GO_FAILPOINTS="github.com/pingcap/dm/syncer/WaitUserCancel=return(8)"
+	# only mock pull binlog failed once
+	export GO_FAILPOINTS="github.com/pingcap/dm/syncer/WaitUserCancel=return(8);github.com/pingcap/dm/syncer/GetEventError=1*return"
```
Ehco1996, line 173: Can you please tell me what the usage of `1 * retrun` ?

Ehco1996 (commented):
restLGTM

syncer/streamer_controller.go
glorv: I don't get the point, this function is `NewStreamerController` 😕️

tests/incremental_mode/run.sh
glorv: Only trigger this error once. See: https://github.com/pingcap/failpoint#key-concepts

Ehco1996 (approved):

PR #2020, opened 2021-08-19
test: swap some test cases to make use of enabled task checker

lance6716:
part of #1997 

source.yaml in shardddl1 disabled task checker, while those in shardddl2 didn't. So swap some test cases to the test group they fit

### Check List 

Tests 

 - No code

Related changes


tests/shardddl1/run.sh
```diff
@@ -413,6 +413,11 @@ function DM_RestartMaster() {
 function restart_master_on_pos() {
 	if [ "$1" = "$2" ]; then
 		restart_master
+		# because we disabled auto-resume feature to satisfy other tests, we manually resume once to handle task was paused
```
Ehco1996, line 416: maybe we can add `rsume-task` to func `restart_master` or add a new func `restart_master_and_resume_task`?

this might avoid duplicate fixes😬

lance6716: some tests rely on auto-resume but some not, so we might check and split them, and these manually resuming can be removed.

cases using `restart_master` may not expected resuming, and `restart_master_and_resume_task` requires too much inspection, not worthy to me.

Ehco1996 (approved):

GMHDBJD (approved):
LGTM

PR #2024, opened 2021-08-19
syncer, loader: make warn log message more accurate

hey-hoho:
Add more accurate log messages to slow SQL.

When the SQL execution time exceeds 1 second, an accurate slow SQL warning will be prompted.

### Check List 

Tests 

 - No code

Related changes

### DIFF
--- a/loader/db.go
+++ b/loader/db.go
@@ -15,6 +15,7 @@ package loader
 
 import (
 	"database/sql"
+	"fmt"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


loader/db.go
```diff
@@ -87,9 +88,11 @@ func (conn *DBConn) querySQL(ctx *tcontext.Context, query string, args ...interf
...
 				}
 				cost := time.Since(startTime)
-				queryHistogram.WithLabelValues(conn.cfg.Name, conn.cfg.SourceID).Observe(cost.Seconds())
-				if cost.Seconds() > 1 {
-					ctx.L().Warn("query statement",
+				// duration seconds
+				ds := cost.Seconds()
+				queryHistogram.WithLabelValues(conn.cfg.Name, conn.cfg.SourceID).Observe(ds)
+				if ds > 1 {
+					ctx.L().Warn(fmt.Sprintf("query statement too slow for %vs", ds),
```
lance6716, line 95: zap field can be used to print the message, no need to introduce `fmt` package.

in fact `zap.Duration("cost time", cost)` has printed the duration.

lance6716 (commented):
rest lgtm

loader/db.go
hey-hoho: OK, I see.
But I think it would be more intuitive to put duration directly in message? Because there is a lot of information in the log, it is not easy to find the `cost time` field.
Or put `cost time`  in front of `query`?

lance6716: I prefer put cost time the first field, in front of query.

hey-hoho: Well, I'll revise it again.

lance6716 (approved):

Ehco1996 (approved):

PR #2036, opened 2021-08-24
test: try to write something when init()

lance6716:

record when we called `init()`. If time between `init()` and `main()` is long we know that some package has a slow `init()`.

### Check List 

Tests 

 - No code

Related changes


lance6716:
/run-unit-tests

seems TiDB with TLS failed

pkg/printinit/noimport.go
```diff
@@ -0,0 +1,9 @@
+package printinit
+
+// we expect this helper function to be called at very early stage of initialization, because this package has no
+// imports.
+// There are some failing CI indicating it takes at least 10 seconds to start a DM-master, we need to know if the OS
+// spawn processes too slowly or some init() are slow to finish.
+func init() {
+	println("init now")
```
lichunzhu, line 8: print a time.Now here？

lichunzhu (approved):
Rest LGTM

pkg/printinit/noimport.go
lance6716: that will introduce an import, which will lower the probability that this init is at early stage of initialization

glorv (approved):
I'm suspicious of the effect of PR. But, since there are no other side effect, we can first merge it and see if it can help investigate the unstable issue.

glorv:
I thinks we needn't merge this PR into release-2.0 branch

Issue #2034, opened 2021-08-24
the online-ddl-scheme value will be covered by online-ddl(false) after dm version upgrade to 2.0.6


Haaahei:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do? If possible, provide a recipe for reproducing the error.
see https://asktug.com/t/topic/123401

2. What did you expect to see?
the value of online-ddl should not be default (false), should depend on the value of online-ddl-scheme

3. What did you see instead?
the value of online-ddl is false, this cause online-ddl-scheme not work.

lance6716:
thanks for reporting, we will add an upgrading logic to adjust `online-ddl` in config in #2039 .

PR #2039, opened 2021-08-24
config: adjust OnlineDDL in subtask config

okJiang:
close https://github.com/pingcap/dm/issues/2034

add AdjustOnlineDDL in upgrade

### Check List 

Tests 

Related changes

### DIFF
--- a/dm/config/subtask.go
+++ b/dm/config/subtask.go
@@ -297,6 +297,9 @@ func (c *SubTaskConfig) Adjust(verifyDecryptPassword bool) error {
 
 	if c.OnlineDDLScheme != "" && c.OnlineDDLScheme != PT && c.OnlineDDLScheme != GHOST {
 		return terror.ErrConfigOnlineSchemeNotSupport.Generate(c.OnlineDDLScheme)
+	} else if c.OnlineDDLScheme == PT || c.OnlineDDLScheme == GHOST {
+		c.OnlineDDL = true
+		log.L().Warn("'online-ddl-scheme' will be deprecated soon. Recommend that use online-ddl instead of online-ddl-scheme.")
 	}
 
 	if c.MetaSchema == "" {
--- a/dm/config/subtask_test.go
+++ b/dm/config/subtask_test.go
@@ -20,11 +20,12 @@ import (
 
 func (t *testConfig) TestSubTask(c *C) {
 	cfg := &SubTaskConfig{
-		Name:       "test-task",
-		IsSharding: true,
-		ShardMode:  "optimistic",
-		SourceID:   "mysql-instance-01",
-		OnlineDDL:  true,
+		Name:            "test-task",
+		IsSharding:      true,
+		ShardMode:       "optimistic",
+		SourceID:        "mysql-instance-01",
+		OnlineDDL:       false,
+		OnlineDDLScheme: PT,
 		From: DBConfig{
 			Host:     "127.0.0.1",
 			Port:     3306,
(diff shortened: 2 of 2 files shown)


dm/config/subtask.go
```diff
@@ -429,3 +429,11 @@ func (c *SubTaskConfig) Clone() (*SubTaskConfig, error) {
 
 	return clone, nil
 }
+
+// AdjustOnlineDDL adjust OnlineDDL through OnlineDDLScheme.
+func (c *SubTaskConfig) AdjustOnlineDDL() {
+	if c.OnlineDDLScheme == GHOST || c.OnlineDDLScheme == PT {
+		c.OnlineDDL = true
+	}
+	c.OnlineDDLScheme = ""
```
lance6716, line 438: the original value should not be changed, to support downgrading

pkg/upgrade/upgrade.go
```diff
@@ -275,3 +276,30 @@ func upgradeToVer3(ctx context.Context, cli *clientv3.Client) error {
 func upgradeToVer4(cli *clientv3.Client, uctx Context) error {
 	return nil
 }
+
+// upgradeToVer5 does upgrade operations from Ver4 to Ver5 (v2.0.7) to update online-ddl from online-ddl-scheme in configure.
```
lance6716, line 280:
```suggestion
// upgradeToVer5 does upgrade operations from Ver4 to Ver5 (v2.1.0) to update online-ddl from online-ddl-scheme in configure.
```

OK to not change, we may change the release version in future 😂

pkg/upgrade/upgrade_test.go
```diff
@@ -162,3 +163,76 @@ func (t *testForEtcd) TestUpgradeToVer3(c *C) {
...
+			OnlineDDL:       true,
+			OnlineDDLScheme: "",
+		},
+		{
+			OnlineDDL:       false,
+			OnlineDDLScheme: "",
+		},
+	}
+	for i := range inputCfgs {
+		fmt.Println(i)
```
lance6716, line 219:
```suggestion
```

pkg/upgrade/upgrade.go
```diff
@@ -275,3 +276,30 @@ func upgradeToVer3(ctx context.Context, cli *clientv3.Client) error {
...
+	}
+
+	for task, m := range uctx.SubTaskConfigs {
+		for sourceID, subCfg := range m {
+			subCfg.AdjustOnlineDDL()
+			value, err := subCfg.Toml()
+			if err != nil {
+				return err
+			}
+			key := common.UpstreamSubTaskKeyAdapter.Encode(sourceID, task)
```
lance6716, line 297: I suddenlt realized that the upgrading actions can't make sure that DM-worker will see the new config. If DM-worker finish upgrading too quickly, at the time of its starting up, DM-master has not run here so task in etcd is not changed.

Maybe we should change the upgrading logic in TiUP 🤔

dm/config/subtask.go
glorv: +1

glorv:
I think it's better to do an `adjust` to the config struct after there are built so we can avoid doing the complex upgrading logic.

dm/config/subtask.go
okJiang: It does not affect the original configure...

pkg/upgrade/upgrade.go
lichunzhu: Some possible solutions:
1. Let worker do adjust job again.
2. If a master finds that its own version info is greater than version Info in etcd, transfer leader to itself immediately and boostrap before this dm-master provides service.
I think we can do this in another PR. It's hard and inappropriate for TiUP to handle DM's own upgrade logic.

lichunzhu: I have checked the logic in dm-worker. dm-worker will adjust subtask config again when dm-worker reads configuration from etcd. So maybe we don't have such problem?

lichunzhu (approved):
LGTM

glorv (approved):

PR #2047, opened 2021-08-25
syncer, relay: avoid read duplicate event when retry binlog streamer

glorv:
avoiding read duplicate event in syncer and relay log.

- Relay
  - discard the events that already written when reset binlog reader in GTID mode
  - return an error `ErrorMaybeDuplicateEvent` when the relay log is not complete(not end with RotateEvent). 
- Syncer 
  - recording the read event index and drop the first n events when reset streamer in GTID mode.
  - drop the first n events when binlog streamer returns the error `ErrorMaybeDuplicateEvent`

### Check List 

Tests 

 - Integration test

Code changes

Side effects

Related changes


lance6716:
seems an unstable test

pkg/streamer/reader.go
```diff
@@ -549,6 +553,9 @@ func (r *BinlogReader) parseFile(
 	if err != nil {
 		if possibleLast && isIgnorableParseError(err) {
 			r.tctx.L().Warn("fail to parse relay log file, meet some ignorable error", zap.String("file", fullPath), zap.Int64("offset", offset), zap.Error(err))
+			// the file is truncated, need to notify the caller
+			// TODO: should add a integration test for this
+			s.ech <- ErrorMaybeDuplicateEvent
```
lance6716, line 558: maybe we should assure the priority of `s.ech` in (*LocalStreamer).GetEvent

relay/relay.go
```diff
@@ -496,13 +541,15 @@ func (r *Relay) handleEvents(ctx context.Context, reader2 reader.Reader, transfo
 
 		if _, ok := e.Event.(*replication.RotateEvent); ok && utils.IsFakeRotateEvent(e.Header) {
 			isNew, err2 := isNewServer(ctx, r.meta.UUID(), r.db.DB, r.cfg.Flavor)
+			// should start from the transaction beginning when switch to a new server
 			if err2 != nil {
-				return err2
+				return 0, err2
```
lance6716, line 546: maybe also return `eventIndex` so we could skip the events?

glorv:
> seems an unstable test

fixed 😅

syncer/syncer.go
```diff
@@ -1737,6 +1737,12 @@ func (s *Syncer) Run(ctx context.Context) (err error) {
 		return nil
 	}
 
+	// eventIndex is the rows event index in this transaction, it's used to avoiding read duplicate event in gtid mode
+	eventIndex := 0
+	// the relay log file may be truncated(not end with an RotateEvent), int this situation, we may read some rows events
```
Ehco1996, line 1742:
```suggestion
	// the relay log file may be truncated(not end with an RotateEvent), in this situation, we may read some rows events
```

syncer/syncer.go
```diff
@@ -1803,11 +1822,27 @@ func (s *Syncer) Run(ctx context.Context) (err error) {
...
-				log.L().Info("reset replication binlog puller")
+				log.L().Info("reset replication binlog puller", zap.Any("pos", s.checkpoint.GlobalPoint()))
+				if s.cfg.EnableGTID {
+					for i := 0; i < eventIndex; {
+						e, err = s.getEvent(tctx, currentLocation)
+						if err != nil {
+							return err
+						}
+						if _, ok := e.Event.(*replication.RowsEvent); ok {
+							i++
```
Ehco1996, line 1838: pelase tell me why we can drop the first n events here? 
will this result in data loss from last saved checkpoint until the `eventIndex` event

syncer/syncer.go
```diff
@@ -1737,6 +1737,12 @@ func (s *Syncer) Run(ctx context.Context) (err error) {
 		return nil
 	}
 
+	// eventIndex is the rows event index in this transaction, it's used to avoiding read duplicate event in gtid mode
+	eventIndex := 0
+	// the relay log file may be truncated(not end with an RotateEvent), int this situation, we may read some rows events
+	// and then read from the gtid again, so we force enter safe-mode for one more transaction to avoid failure due to
+	// conflict
+	safeMode := false
```
lance6716, line 1745: I think when we meet ErrorMaybeDuplicateEvent, we know that the first half part of transaction is in the previous relay log file, so we could also use `eventIndex` to skip them.

syncer/syncer.go
Ehco1996: i think we should alway into safemode for one trx in this situation

tests/duplicate_event/run.sh
```diff
@@ -0,0 +1,95 @@
+#!/bin/bash
+
+set -eu
+
+cur=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
+source $cur/../_utils/test_prepare
+WORK_DIR=$TEST_DIR/$TEST_NAME
+TASK_NAME="test"
```
Ehco1996, line 8: seems `TASK_NAME, SQL_RESULT_FILE ,API_VERSION ` is not used in this test

tests/duplicate_event/run.sh
```diff
@@ -0,0 +1,95 @@
...
+SQL_RESULT_FILE="$TEST_DIR/sql_res.$TEST_NAME.txt"
+
+API_VERSION="v1alpha1"
+
+function run() {
+	# 1. test sync fetch binlog met error and reset binlog streamer with remote binlog
```
Ehco1996, line 14:
```suggestion
	# 1. test sync fetch binlog meet error and reset binlog streamer with remote binlog
```

syncer/syncer.go
```diff
@@ -1808,6 +1814,14 @@ func (s *Syncer) Run(ctx context.Context) (err error) {
 				err = errors.New("connect: connection refused")
 			}
 		})
+		failpoint.Inject("GetEventErrorInTxn", func(val failpoint.Value) {
+			if intVal, ok := val.(int); ok && intVal == eventIndex {
+				err = errors.New("fail point triggered")
```
Ehco1996, line 1819:
```suggestion
				err = errors.New("failpoint triggered")
```

tests/duplicate_event/run.sh
```diff
@@ -0,0 +1,95 @@
...
+		"start-task $cur/conf/dm-task-relay.yaml --remove-meta"
+	check_metric $WORKER2_PORT "dm_worker_task_state{source_id=\"mysql-replica-02\",task=\"test_relay\",worker=\"worker2\"}" 10 1 3
+
+	check_sync_diff $WORK_DIR $cur/conf/diff_relay_config.toml
+
+	run_sql_source2 "flush logs;"
+	run_dm_ctl_with_retry $WORK_DIR "127.0.0.1:$MASTER_PORT" \
+		"start-relay -s $SOURCE_ID2 worker2" \
+		"\"result\": true" 1
+	sleep 5
```
Ehco1996, line 70: why sleep here?

relay/relay.go
```diff
@@ -319,6 +320,33 @@ func (r *Relay) process(ctx context.Context) error {
...
+				goto checkError
+			}
+			if !isNew {
+				for i := 0; i < eventInx; {
+					res, err2 := reader2.GetEvent(ctx)
+					if err2 != nil {
+						err = err2
+						goto checkError
+					}
+					tResult := transformer2.Transform(res.Event)
```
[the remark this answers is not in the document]
glorv: Need to depend on this transform to ensure we don't count useless event such as `HEARTBEAT_EVENT`

relay/relay.go
glorv: No. We need to ensure the new file start with a complete transaction because we need to include the `PreviousGTIDSET` event at the file beginning.

syncer/syncer.go
glorv: In GTID mode, the reader will start for the first event of the transaction after the `streamerController` is reset. So we should skip the first N rowsEevets because they have been processed. Though enter safe mode is another way to do so, but just skip should be a more straight forward way to avoid duplication. Maybe you need to take a deeper look at the changes.😅

tests/duplicate_event/run.sh
glorv: To Ensure the relay unit has catch up with the upstream.

Ehco1996: maybe use `run_dmctl_with_retry query-status -s source1 "relayCatchUpMaster": true` ?

syncer/syncer.go
Ehco1996: > first N rowsEevets because they have been processed

i don't think our code ensure this, for example one tx have 3 events  and getEvent met error at the second event,
in this case, all we can guarantee is that event1's sql is already in the queue, not that it has been executed downstream, i am sure if this will result in data loss🤔, so i perfer to alway into safemode. in addition the connection-level retry already serves our purpose of not interrupting the task

but ignore first N event in relay unit is ok to me , because we can make sure all event is write to file

tests/duplicate_event/run.sh
```diff
@@ -0,0 +1,95 @@
...
+SQL_RESULT_FILE="$TEST_DIR/sql_res.$TEST_NAME.txt"
+
+API_VERSION="v1alpha1"
+
+function run() {
+	# 1. test sync fetch binlog met error and reset binlog streamer with remote binlog
+
+	# with a 5 rows insert txn: 1 * GTID + 1 * BEGIN + 5 * (Table_map + Write_rows) + 1 * XID
+	# here we fail at the third write rows event, sync should retry and auto recover without any duplicate event
+	export GO_FAILPOINTS="github.com/pingcap/dm/syncer/GetEventErrorInTxn=13*return(3)"
```
lance6716, line 18: there should be format_description event and previoud_gtid event so it's 15, so maybe we should check `failed to get event` in log.
But seems 13 also triggered "discard event already consumed" 🤔

tests/duplicate_event/run.sh
```diff
@@ -0,0 +1,95 @@
...
+	check_contains 'Query OK, 2 rows affected'
+
+	run_dm_master $WORK_DIR/master $MASTER_PORT $cur/conf/dm-master.toml
+	check_rpc_alive $cur/../bin/check_master_online 127.0.0.1:$MASTER_PORT
+	check_metric $MASTER_PORT 'start_leader_counter' 3 0 2
+	run_dm_worker $WORK_DIR/worker1 $WORKER1_PORT $cur/conf/dm-worker1.toml
+	check_rpc_alive $cur/../bin/check_worker_online 127.0.0.1:$WORKER1_PORT
+
+	cp $cur/conf/source1.yaml $WORK_DIR/source1.yaml
+	dmctl_operate_source create $WORK_DIR/source1.yaml $SOURCE_ID1
```
lance6716, line 30:
```suggestion
	dmctl_operate_source create $cur/conf/source1.yaml $SOURCE_ID1
```

if we don't modify $cur/conf/source1.yaml

and for source2

lance6716 (approved):
weakly want to change the safe-mode behaviour

https://github.com/pingcap/dm/pull/2047/files#r699233692

tests/duplicate_event/run.sh
glorv: The failpoint will trigger the error at the third WriteRowsEvent, so 13 is a upper bound the ensure that it won't impact the next txn

lance6716: 1. format_description
2. previous_gtid
3. gtid
4. query (begin)
5 ~ 8
9. table_map 3
10. write_rows 3
11 ~ 14
15. xid

now I see 13 times is enough

syncer/syncer.go
glorv: I don't get the point in what condition in your example will it cause a data loss, can you explain it in detail @Ehco1996

tests/duplicate_event/run.sh
glorv: https://github.com/pingcap/dm/pull/2047/files#r699233692 I think in this comment, @lance6716 also prefer to just skip the duplicate event instead of relay on safe-mode.

syncer/syncer.go
glorv: https://github.com/pingcap/dm/pull/2047/files#r699233692 I think in this comment, @lance6716 also prefer to just skip the duplicate event instead of relay on safe-mode.

glorv:
@lance6716 I change the `safe-mode` logic to also directly skip N event, PTAL again.

lance6716:
lgtm

Ehco1996 (approved):
LGTM, please remember to fix lint error

syncer/syncer.go
glorv: The former logic also directly retry some error instead of enter Pause&Resume cycle when meets some error in GetEvent, this PR just resolve the issue that the former logic may cause duplicate error when GTID is enabled.😅

PR #2051, opened 2021-08-26
*: support lightning

Little-Wallace:
Signed-off-by: Little-Wallace <bupt2013211450@gmail.com>

I will use lightning import the full data produced by dumpling instead of loader. Lightning could import data much more quickly than loader.

User only need to set configure `tidb.backend` to `local` or `tidb`. Then DM-master will start a task in DM-worker with lightning.

### Check List 

Tests 

 - Unit test

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes


dm/config/subtask.go
```diff
@@ -147,6 +149,37 @@ func GetDBConfigFromEnv() DBConfig {
 	}
 }
 
+// TiDBExtraConfig is the extra DB configuration only for TiDb.
```
lance6716, line 152:
```suggestion
// TiDBExtraConfig is the extra DB configuration only for TiDB.
```

dm/master/server.go
```diff
@@ -1350,6 +1350,11 @@ func (s *Server) generateSubTask(ctx context.Context, task string, errCnt, warnC
 	sourceCfgs := s.getSourceConfigs(cfg.MySQLInstances)
 
 	stCfgs, err := cfg.SubTaskConfigs(sourceCfgs)
+	if cfg.TiDB != nil {
```
lance6716, line 1353: seems this logic should be put inside `SubTaskConfigs`

dm/worker/worker.go
```diff
@@ -403,6 +405,11 @@ func (w *Worker) StartSubTask(cfg *config.SubTaskConfig, expectStage pb.Stage, n
 
 	// directly put cfg into subTaskHolder
 	// the unique of subtask should be assured by etcd
+	if w.logConfig != nil && cfg.LogFile == "" {
+		cfg.LogLevel = w.logConfig.Level
```
lance6716, line 409: can lightning use external logger? so we won't bother adding a logConfig

dumpling's example https://github.com/pingcap/dm/blob/6ad2f890822495986e7d98a55d5202b72a8442b9/dumpling/dumpling.go#L234

dm/worker/subtask.go
```diff
@@ -59,7 +59,11 @@ func createRealUnits(cfg *config.SubTaskConfig, etcdClient *clientv3.Client, wor
 	switch cfg.Mode {
 	case config.ModeAll:
 		us = append(us, dumpling.NewDumpling(cfg))
-		us = append(us, loader.NewLoader(cfg, etcdClient, workerName))
+		if cfg.TiDB.Backend == "" {
+			us = append(us, loader.NewLoader(cfg, etcdClient, workerName))
+		} else {
+			us = append(us, loader.NewLightningLoader(cfg, etcdClient, workerName))
```
lance6716, line 65: I think `NewLightning` is enough, just like `NewDumpling`

loader/lightning.go
```diff
@@ -0,0 +1,318 @@
...
+// Process implements Unit.Process.
+func (l *LightningLoader) Process(ctx context.Context, pr chan pb.ProcessResult) {
+	l.logger.Info("lightning load start")
+	errs := make([]*pb.ProcessError, 0, 1)
+	if err := l.restore(ctx); err != nil {
+		errs = append(errs, unit.NewProcessError(err))
+	}
+	l.logger.Info("lightning load end")
+	pr <- pb.ProcessResult{
+		IsCanceled: false,
```
lance6716, line 185: this field is used to distinguish user canceling context, in current implementation it should be set to `true` if context is canceled (seems we can move this logic to caller 🤔 )

https://github.com/pingcap/dm/blob/6ad2f890822495986e7d98a55d5202b72a8442b9/loader/loader.go#L616-L628

dm/worker/subtask.go
Little-Wallace: OK

loader/loader.go
```diff
@@ -1560,31 +1560,31 @@ func (l *Loader) cleanDumpFiles() {
...
-			l.logger.Warn("show last error when remove loaded dump sql files", zap.String("data folder", l.cfg.Dir), zap.Error(lastErr))
+			log.L().Warn("show last error when remove loaded dump sql files", zap.String("data folder", cfg.Dir), zap.Error(lastErr))
 		}
 	}
 }
 
 // putLoadTask is called when start restoring data, to put load worker in etcd.
-func (l *Loader) putLoadTask() error {
-	_, err := ha.PutLoadTask(l.cli, l.cfg.Name, l.cfg.SourceID, l.workerName)
+func putLoadTask(cli *clientv3.Client, cfg *config.SubTaskConfig, workerName string) error {
```
lance6716, line 1573: slightly recommend move those helper function into loader/util.go

loader/lightning.go
```diff
@@ -0,0 +1,273 @@
...
+		lightningCfg.Security.KeyPath = cfg.To.Security.SSLKey
+	}
+	lightningCfg.TiDB.Host = cfg.To.Host
+	lightningCfg.TiDB.Psw = cfg.To.Password
+	lightningCfg.TiDB.User = cfg.To.User
+	lightningCfg.TiDB.Port = cfg.To.Port
+	lightningCfg.TiDB.StatusPort = cfg.TiDB.StatusPort
+	lightningCfg.TiDB.PdAddr = cfg.TiDB.PdAddr
+	lightningCfg.TikvImporter.Backend = cfg.TiDB.Backend
+	lightningCfg.PostRestore.Checksum = lcfg.OpLevelOff
```
lance6716, line 92: will we need it for local backend?

loader/lightning.go
```diff
@@ -0,0 +1,273 @@
...
+// Type implements Unit.Type.
+func (l *LightningLoader) Type() pb.UnitType {
+	return pb.UnitType_Load
+}
+
+// Init initializes loader for a load task, but not start Process.
+// if fail, it should not call l.Close.
+func (l *LightningLoader) Init(ctx context.Context) (err error) {
+	tctx := tcontext.NewContext(ctx, l.logger)
+	checkpoint, err := newRemoteCheckPoint(tctx, l.cfg, l.checkpointID())
```
lance6716, line 109: lightning will use file checkpoint by default, and this checkpoint is only a dummy entry? 
How about we use a mysql checkpoint, even if task is failed, user still can query the progress from checkpoint.

we need to remove file checkpoint when enter https://github.com/pingcap/dm/blob/487a3d10f0bb1548ce1c8a0207a1ad4a95fd9f6e/dm/master/server.go#L467

loader/lightning.go
```diff
@@ -0,0 +1,273 @@
...
+
+// Status returns the unit's current status.
+func (l *LightningLoader) Status(_ *binlog.SourceStatus) interface{} {
+	finished, total := l.core.Status()
+	progress := percent(finished, total, l.finish.Load())
+	s := &pb.LoadStatus{
+		FinishedBytes:  finished,
+		TotalBytes:     total,
+		Progress:       progress,
+		MetaBinlog:     "0",
```
lance6716, line 256: seems we need `(*Loader).getMydumpMetadata`

lance6716 (commented):
and will we add tests in this or next PR?

loader/lightning.go
Little-Wallace: But dm may load multiple sharding to tidb, and we can not guarentee that no other client try to write data to tidb during checksum

tests/lightning_mode/conf/dm-task.yaml
```diff
@@ -0,0 +1,63 @@
...
+
+mydumpers:
+  global:
+    threads: 4
+    chunk-filesize: 64
+    skip-tz-utc: true
+    extra-args: ""
+
+loaders:
+  global:
```
lance6716, line 55: didn't set lightning backend. and I think we can use one of below ways to test

- add a new case of `lightning_mode`, the database name of this test should be `lightning_mode` not `all_mode`, and `lightning_mode` should be added into `other_integration_1.txt` (or 2)
- just change an existing case (like all_mode), modify the task configuration to switch to lightning loader

loader/lightning.go
```diff
@@ -0,0 +1,306 @@
+// Copyright 2019 PingCAP, Inc.
```
lance6716, line 1:
```suggestion
// Copyright 2021 PingCAP, Inc.
```

loader/lightning.go
```diff
@@ -0,0 +1,306 @@
...
+	"github.com/pingcap/dm/dm/config"
+	"github.com/pingcap/dm/dm/pb"
+	"github.com/pingcap/dm/dm/unit"
+	"github.com/pingcap/dm/pkg/binlog"
+	"github.com/pingcap/dm/pkg/conn"
+	tcontext "github.com/pingcap/dm/pkg/context"
+	"github.com/pingcap/dm/pkg/log"
+	"github.com/pingcap/dm/pkg/utils"
+
+	"github.com/pingcap/failpoint"
```
lance6716, line 34: need to group the 3-rd packages

loader/lightning.go
```diff
@@ -0,0 +1,306 @@
...
+			l.finish.Store(true)
+			offsetSQL := l.checkPoint.GenSQL(lightningCheckpointFile, 1)
+			err = l.toDBConns[0].executeSQL(tctx, []string{offsetSQL})
+			_ = l.checkPoint.UpdateOffset(lightningCheckpointFile, 1)
+		}
+	} else {
+		l.finish.Store(true)
+	}
+	if l.cfg.Mode == config.ModeFull {
+		_ = delLoadTask(l.cli, l.cfg, l.workerName)
```
lance6716, line 182: I think this is the only chance to `delLoadTask`, so we'd better let the task fail and retry to delete the marker in etcd. Otherwise the source is always pinned to this worker.

loader/lightning.go
```diff
@@ -0,0 +1,306 @@
...
+		TotalBytes:     total,
+		Progress:       progress,
+		MetaBinlog:     l.metaBinlog.Load(),
+		MetaBinlogGTID: l.metaBinlogGTID.Load(),
+	}
+	return s
+}
+
+// checkpointID returns ID which used for checkpoint table.
+func (l *LightningLoader) checkpointID() string {
```
lance6716, line 296: seems also a shared function

lance6716 (commented):
rest lgtm

loader/lightning.go
glorv: Actually tidb-backend doesn't support any kind of `checksum`😅

loader/lightning.go
```diff
@@ -0,0 +1,307 @@
...
+	l.Process(ctx, pr)
+}
+
+// Update implements Unit.Update
+// now, only support to update config for routes, filters, column-mappings, block-allow-list
+// now no config diff implemented, so simply re-init use new config
+// no binlog filter for loader need to update.
+func (l *LightningLoader) Update(cfg *config.SubTaskConfig) error {
+	// update l.cfg
+	l.cfg.BAList = cfg.BAList
```
glorv, line 276: How does these configs take effect. After table checkpoint has been inited, these config change won't take effect, does this behavior in line with expectation?

Issue #524, opened 2020-03-09
Validate CommonName attribute for TLS certificate


tennix:
## Feature Request

**Is your feature request related to a problem? Please describe:**

If we follow current TLS in TiDB cluster, any TLS certificates issued by the same CA can access TiDB cluster component. This is insecure as the same CA issued certificates are pretty common.

**Describe the feature you'd like:**

Allow DM add TLS certificate CN validation, this is the issue in PD, TiKV and TiDB pingcap/pd#2209 tikv/tikv#6982 pingcap/tidb#15137

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

WangXiangUSTC:
finish in https://github.com/pingcap/dm/pull/569

PR #2055, opened 2021-08-27
openapi: add source releated API

Ehco1996:
#1982

Part3 - Source Related APIs

as title

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

 - Increased code complexity

Related changes


dm/master/openapi.go
```diff
@@ -154,3 +203,36 @@ func sendHTTPErrorResp(ctx echo.Context, code int, message string) error {
 	err := openapi.ErrorWithMessage{ErrorMsg: message, ErrorCode: code}
 	return ctx.JSON(http.StatusBadRequest, err)
 }
+
+func sourceCfgToModel(cfg config.SourceConfig) openapi.Source {
+	// NOTE we don't return SSL cert here, because we don't want to expose it to the user.
```
Ehco1996, line 208: security

dm/master/openapi_test.go
```diff
@@ -80,3 +127,78 @@ func (t *openAPISuite) TestRedirectRequestToLeader(c *check.C) {
...
+	source2 := source1
+	result2 := testutil.NewRequest().Post(baseURL).WithJsonBody(source2).Go(t.testT, s.echo)
+	// check http status code
+	c.Assert(result2.Code(), check.Equals, http.StatusBadRequest)
+	var errResp openapi.ErrorWithMessage
+	err = result2.UnmarshalBodyToObject(&errResp)
+	c.Assert(err, check.IsNil)
+	c.Assert(errResp.ErrorCode, check.Equals, int(terror.ErrSchedulerSourceCfgExist.Code()))
+
+	// list sources
```
Ehco1996, line 171: source

tests/openapi/client/source.py
```diff
@@ -0,0 +1,94 @@
...
+    resp = requests.post(url=API_ENDPOINT, json=req)
+    assert resp.status_code == 201
+    print("create_source1_success resp=", resp.json())
+
+
+def list_source_success(source_count):
+    resp = requests.get(url=API_ENDPOINT)
+    assert resp.status_code == 200
+    data = resp.json()
+    # only create one source
```
Ehco1996, line 54: remove this

tests/openapi/conf/dm-master1.toml
```diff
@@ -0,0 +1,6 @@
+# Master Configuration.
+advertise-addr = "127.0.0.1:8261"
+peer-urls = "127.0.0.1:8291"
+initial-cluster = "master1=http://127.0.0.1:8291"
+master-addr = ":8261"
+name = "master1"
```
Ehco1996, line 6: end of line

dm/config/source_config_test.go
```diff
@@ -312,3 +314,24 @@ func getMockServerIDs(ctx context.Context, db *sql.DB) (map[uint32]struct{}, err
...
+
+	db, mock, err := sqlmock.New()
+	c.Assert(err, IsNil)
+	mock.ExpectQuery("SELECT @@lower_case_table_names;").
+		WillReturnRows(sqlmock.NewRows([]string{"@@lower_case_table_names"}).AddRow(utils.LCTableNamesMixed))
+	mock.ExpectClose()
+	c.Assert(cfg.AdjustCaseSensitive(context.Background(), db), IsNil)
+	c.Assert(cfg.CaseSensitive, Equals, false)
+
+	db, mock, err = sqlmock.New()
```
lance6716, line 330: could reuse previous mock

dm/config/source_config_test.go
```diff
@@ -312,3 +314,24 @@ func getMockServerIDs(ctx context.Context, db *sql.DB) (map[uint32]struct{}, err
...
+	c.Assert(err, IsNil)
+	mock.ExpectQuery("SELECT @@lower_case_table_names;").
+		WillReturnRows(sqlmock.NewRows([]string{"@@lower_case_table_names"}).AddRow(utils.LCTableNamesSensitive))
+	mock.ExpectClose()
+	c.Assert(cfg.AdjustCaseSensitive(context.Background(), db), IsNil)
+	c.Assert(cfg.CaseSensitive, Equals, true)
```
lance6716, line 336: and check ExpectationAllMeet

dm/master/openapi.go
```diff
@@ -88,17 +90,64 @@ func (s *Server) GetDocHTML(ctx echo.Context) error {
 
 // DMAPICreateSource url is:(POST /api/v1/sources).
 func (s *Server) DMAPICreateSource(ctx echo.Context) error {
```
lance6716, line 92: will this method print some logs to audit?

dm/config/source_config.go
```diff
@@ -271,6 +271,16 @@ func (c *SourceConfig) Adjust(ctx context.Context, db *sql.DB) (err error) {
...
+	return c.AdjustCaseSensitive(ctx2, db)
+}
+
+// AdjustCaseSensitive adjust CaseSensitive from DB.
+func (c *SourceConfig) AdjustCaseSensitive(ctx context.Context, db *sql.DB) (err error) {
+	caseSensitive, err2 := utils.GetDBCaseSensitive(ctx, db)
+	if err2 != nil {
+		return err2
+	}
+	c.CaseSensitive = caseSensitive
```
lance6716, line 283: Is this field deprecated? @glorv

dm/master/openapi.go
Ehco1996: there is a access log in openapi logger. 
it looks like this: `
[2021/08/30 17:50:27.642 +08:00] [INFO] [logger.go:47] [Success] [component=openapi] [size=128] [host=127.0.0.1:8261] [status=201] [method=POST] [protocol=HTTP/1.1] [remote_ip=127.0.0.1] [user_agent=python-requests/2.22.0] [duration=171.443871ms] [request="POST /api/v1/sources"]`

dm/master/openapi.go
```diff
@@ -88,17 +90,64 @@ func (s *Server) GetDocHTML(ctx echo.Context) error {
...
-	return nil
+	needRedirect, host, err := s.redirectRequestToLeader(ctx.Request().Context())
```
lance6716, line 93: maybe redirect is a good place to use middleware?

dm/master/openapi.go
```diff
@@ -88,17 +90,64 @@ func (s *Server) GetDocHTML(ctx echo.Context) error {
...
+		return err
+	}
+	cfg := modelToSourceCfg(createSourceReq)
+	if err := checkAndAdjustSourceConfig(ctx.Request().Context(), cfg); err != nil {
+		return err
+	}
+	if err := s.scheduler.AddSourceCfg(cfg); err != nil {
+		return err
+	}
+	return ctx.JSON(http.StatusCreated, createSourceReq)
```
lance6716, line 112: `OperateSource` returns the binding relation, I think that's more helpful than copying user's input.

dm/master/openapi.go
```diff
@@ -88,17 +90,64 @@ func (s *Server) GetDocHTML(ctx echo.Context) error {
...
+		return err
+	}
+	if needRedirect {
+		return ctx.Redirect(http.StatusTemporaryRedirect, fmt.Sprintf("http://%s%s", host, ctx.Request().RequestURI))
+	}
+
+	sourceIDS := s.scheduler.GetSourceCfgIDs()
+	sourceCfgList := make([]*config.SourceConfig, len(sourceIDS))
+	for idx, sourceID := range sourceIDS {
+		sourceCfgList[idx] = s.scheduler.GetSourceCfgByID(sourceID)
```
lance6716, line 128: `GetSourceCfgs`

dm/master/openapi.go
```diff
@@ -154,3 +203,36 @@ func sendHTTPErrorResp(ctx echo.Context, code int, message string) error {
 	err := openapi.ErrorWithMessage{ErrorMsg: message, ErrorCode: code}
 	return ctx.JSON(http.StatusBadRequest, err)
 }
+
+func sourceCfgToModel(cfg config.SourceConfig) openapi.Source {
+	// NOTE we don't return security content here, because we don't want to expose it to the user.
```
lance6716, line 208: but decrypted password is returned

dm/master/openapi.go
```diff
@@ -88,17 +90,64 @@ func (s *Server) GetDocHTML(ctx echo.Context) error {
...
+	if err := ctx.Bind(&createSourceReq); err != nil {
+		return err
+	}
+	cfg := modelToSourceCfg(createSourceReq)
+	if err := checkAndAdjustSourceConfig(ctx.Request().Context(), cfg); err != nil {
+		return err
+	}
+	if err := s.scheduler.AddSourceCfg(cfg); err != nil {
+		return err
+	}
```
lance6716, line 111: `OperateSource` has waited for that worker has responsed to the action

tests/openapi/run.sh
```diff
@@ -0,0 +1,75 @@
...
+}
+
+function test_source() {
+	echo ">>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>START TEST OPENAPI: SOURCE"
+	prepare_database
+	# create source succesfully
+	python $cur/client/source.py "create_source1_success"
+
+	# recreate source will failed
+	python $cur/client/source.py "create_source_failed"
```
lance6716, line 25: use `source_check "create_source_failed"` like other commands

dm/master/openapi.go
```diff
@@ -35,6 +34,27 @@ const (
 	docJSONBasePath = "/api/v1/dm.json"
 )
 
+// redirectRequestToLeaderMW a middleware auto redirect request to leader.
+// because the leader has some data in memory, only the leader can process the request.
+func (s *Server) redirectRequestToLeaderMW() echo.MiddlewareFunc {
+	return func(next echo.HandlerFunc) echo.HandlerFunc {
+		return func(ctx echo.Context) error {
+			ctx2 := ctx.Request().Context()
+
```
Ehco1996, line 43: remove this line

dm/master/openapi.go
Ehco1996: I'm not quite sure if I need to hide the password here, because the API usage scenario is a bit different from ctl
When the user uses ctl, he has a copy of the yaml archive, so the yaml has a record of the password, but if the API is used, the user is uploading the plain text and may don't have raw config backup, so it would be a bit strange if we encrypted it and sent it back🤔

dm/master/openapi.go
Ehco1996: At the API level, I don't think it's necessary to return whether the source is successfully bound or not, the user can wheel through the `source status`(in next PR) api to get the result

Essentially because the DM implementation is declarative and asynchronous, the success of the API is simply an indication that dm-master has accepted the request and started processing it

dm/master/openapi.go
lance6716: I remember scheduler stores plain password, so user has uploaded an encrypted password but this API will decrypt it

dm/master/openapi.go
lance6716: OK, if TiCDC also treat the API as not declarative, we can also do so.

dm/master/openapi.go
Ehco1996: 😂 yes it's, but  there is no way to know if the user upload is encrypted or not (because the scheduler does not store such information)，however, there is no indication in the API documentation that the change field can be encrypted by dmctl,so maybe it's not a problem at now

dm/master/openapi.go
Ehco1996: after read cdc's code,  it did the same https://github.com/pingcap/ticdc/blob/8abf12a5f29143d0379103847eb2b08d9a033524/cdc/capture/http_handler.go#L234

tests/openapi/run.sh
```diff
@@ -0,0 +1,75 @@
...
+
+	run_sql 'DROP DATABASE if exists openapi;' $MYSQL_PORT2 $MYSQL_PASSWORD2
+	run_sql 'CREATE DATABASE openapi;' $MYSQL_PORT2 $MYSQL_PASSWORD2
+}
+
+function test_source() {
+	echo ">>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>START TEST OPENAPI: SOURCE"
+	prepare_database
+	# create source succesfully
+	$cur/client/source.py "create_source1_success"
```
lance6716, line 22:
```suggestion
	openapi_check "create_source1_success"
```

like tests/_utils/check_ticker_interval

tests/openapi/run.sh
```diff
@@ -0,0 +1,75 @@
...
+	$cur/client/source.py "list_source_with_redirect" 0
+
+	echo ">>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>TEST OPENAPI: SOURCE SUCCESS"
+}
+
+function run() {
+	if [[ $(pip list) == *"requests"* ]]; then
+		echo "requests already installed"
+	else
+		sudo pip install requests # in ci need sudo to install requests
```
lance6716, line 49: we should install the requirement in Makefile

lance6716 (commented):
rest lgtm

dm/master/openapi.go
sunzhaoyang: “but if the API is used, the user is uploading the plain text and may don't have raw config backup” 
There is no need to consider this case, the user needs to save the password himself in a secure way, and we should ensure that the received plaintext password is always saved and transmitted in an encrypted way @Ehco1996 @lance6716

tests/openapi/run.sh
Ehco1996: fixed in https://github.com/pingcap/dm/pull/2055/commits/1d68f49c3873e6c62b1e4ed7bab8832893f73034

tests/openapi/run.sh
Ehco1996: At the moment only the integration tests for openapi need this dependency, I think it would be easier to read in a separate test file?😬

dm/master/openapi.go
```diff
@@ -154,3 +184,36 @@ func sendHTTPErrorResp(ctx echo.Context, code int, message string) error {
...
+func modelToSourceCfg(source openapi.Source) *config.SourceConfig {
+	cfg := config.NewSourceConfig()
+	from := config.DBConfig{
+		Host:     source.Host,
+		Port:     source.Port,
+		User:     source.User,
+		Password: source.Password,
+	}
+	if source.Security != nil {
+		from.Security = &config.Security{
```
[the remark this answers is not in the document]
Ehco1996: seems we don't need this now

dm/master/openapi.go
glorv: There is no extra information (e.g. the resource_id created for a web service) we need to return, so here I think we just return an empty response is ok.

tests/openapi/run.sh
lance6716: you can call `make install_python_dep` in test scripts 🤔

dm/master/openapi.go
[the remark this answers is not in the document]
lance6716: https://github.com/pingcap/dm/issues/524 the requirement is here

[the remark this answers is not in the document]
Ehco1996: added in https://github.com/pingcap/dm/pull/2055/commits/b07ee1cdee6e28a200a7233b0557c4d888acdb07

openapi/spec/dm.yaml
```diff
@@ -297,6 +297,11 @@ components:
           type: string
           example: ""
           description: "Content of the private key file in X509 format"
+        cert_allowed_cn:
+          type: array
+          description: "Common Name of ssl certs"
```
lance6716, line 302: SSL

tests/openapi/run.sh
```diff
@@ -43,11 +43,7 @@ function test_source() {
 }
 
 function run() {
-	if [[ $(pip list) == *"requests"* ]]; then
-		echo "requests already installed"
-	else
-		sudo pip install requests # in ci need sudo to install requests
-	fi
+	make install_test_python_dep
```
lance6716, line 46: not sure if we should pass path of Makefile to `make`, because it looks for current directory

dm/master/openapi.go
```diff
@@ -186,16 +186,24 @@ func sendHTTPErrorResp(ctx echo.Context, code int, message string) error {
...
 		EnableGtid: cfg.EnableGTID,
 		Host:       cfg.From.Host,
 		Password:   "******",
 		Port:       cfg.From.Port,
 		SourceName: cfg.SourceID,
 		User:       cfg.From.User,
 	}
+	if cfg.From.Security != nil {
+		// NOTE we don't return security content here, because we don't want to expose it to the user.
+		certAllowedCn := []string{}
```
lance6716, line 200: maybe initialized as `nil`?

var certAllowedCn []string

lance6716 (commented):
rest lgtm

tests/openapi/run.sh
Ehco1996: in `make integration_test CASE="xxx"`, current  directory is the project's root directory , so it's ok to use `make xxx` here

openapi/spec/dm.yaml
Ehco1996: fixed in https://github.com/pingcap/dm/pull/2055/commits/dae765d7920c5b76f2b1873e1f6a70fca818dbfb

lance6716 (approved):

glorv (approved):

PR #2063, opened 2021-08-30
master: return correct "result" for commands even error exists

lance6716:
if we want to pause a task and the task has already meet error, before this PR it returns `result: false` which is not reasonable because the task is indeed paused

separate generating result boolean and error message logic

### Check List 

Tests 

 - Unit test
 - Integration test

Side effects

Related changes


dm/master/server.go
```diff
@@ -1530,73 +1533,99 @@ func (s *Server) waitOperationOk(ctx context.Context, cli *scheduler.Worker, tas
...
-							if st, ok := subtaskStatus.Status.(*pb.SubTaskStatus_Msg); ok && st.Msg == fmt.Sprintf("no sub task with name %s has started", taskName) {
-								return queryResp, nil
+							if st, ok2 := subtaskStatus.Status.(*pb.SubTaskStatus_Msg); ok2 && st.Msg == common2.NoSubTaskMsg(taskName) {
+								ok = true
 							}
 						} else if subtaskStatus.Name == taskName && (subtaskStatus.Stage == expect || subtaskStatus.Stage == finished) {
-							return queryResp, nil
+							ok = true
+						}
+						if ok || msg != "" {
```
Ehco1996, line 1605: perfer to only check `ok == true` here

dm/master/server.go
```diff
@@ -1530,73 +1533,99 @@ func (s *Server) waitOperationOk(ctx context.Context, cli *scheduler.Worker, tas
...
+				}
+				if len(queryResp.SubTaskStatus) == 1 {
 					if subtaskStatus := queryResp.SubTaskStatus[0]; subtaskStatus != nil {
-						if err := extractWorkerError(subtaskStatus.Result); err != nil {
-							return queryResp, err
+						msg := ""
+						if err2 := extractWorkerError(subtaskStatus.Result); err2 != nil {
+							msg = err2.Error()
 						}
+						ok := false
```
Ehco1996, line 1592: maybe use `ok := msg == ""`, so that we don't need to check msg!="" before return

dm/master/server.go
```diff
@@ -69,7 +70,7 @@ const (
 
 var (
 	// the retry times for dm-master to confirm the dm-workers status is expected.
-	maxRetryNum = 30
+	maxRetryNum = 10
```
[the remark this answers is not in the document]
lance6716: a customer complaints about the command never returned

dm/master/server.go
lance6716: the two variables are independent

dm/master/server.go
```diff
@@ -37,6 +37,7 @@ import (
 	"google.golang.org/grpc"
 
 	"github.com/pingcap/dm/checker"
+	common2 "github.com/pingcap/dm/dm/common"
```
lonng, line 40: Prefer to use a more clear package alias?

dm/master/server.go
```diff
@@ -1461,27 +1462,29 @@ func extractWorkerError(result *pb.ProcessResult) error {
...
+//     * pause: related relay status is paused
+//     * resume: related relay status is running
+// returns OK, error message of QueryStatusResponse, raw QueryStatusResponse, error that not from QueryStatusResponse.
+func (s *Server) waitOperationOk(
+	ctx context.Context,
+	cli *scheduler.Worker,
+	taskName string,
+	sourceID string,
+	masterReq interface{},
+) (bool, string, *pb.QueryStatusResponse, error) {
```
okJiang, line 1487: Doesn't `*pb.QueryStatusResponse` contain `result` and `msg`? Can you put `OK` and `error message of QueryStatusResponse` in the `pb.QueryStatusResponse`?

https://github.com/pingcap/dm/blob/master/dm/pb/dmworker.pb.go#L407-L414
```
// QueryStatusResponse represents status response for query on a dm-worker
// status: dm-worker's current sub tasks' status
type QueryStatusResponse struct {
	Result        bool             `protobuf:"varint,1,opt,name=result,proto3" json:"result,omitempty"`
	Msg           string           `protobuf:"bytes,2,opt,name=msg,proto3" json:"msg,omitempty"`
	SourceStatus  *SourceStatus    `protobuf:"bytes,3,opt,name=sourceStatus,proto3" json:"sourceStatus,omitempty"`
	SubTaskStatus []*SubTaskStatus `protobuf:"bytes,4,rep,name=subTaskStatus,proto3" json:"subTaskStatus,omitempty"`
}
```

dm/worker/status.go
```diff
@@ -71,8 +71,9 @@ func (w *Worker) Status(stName string) []*pb.SubTaskStatus {
 		var stStatus pb.SubTaskStatus
 		if !ok {
 			stStatus = pb.SubTaskStatus{
-				Name:   name,
-				Status: &pb.SubTaskStatus_Msg{Msg: fmt.Sprintf("no sub task with name %s has started", name)},
+				Name: name,
+
```
okJiang, line 75: > 
```suggestion
```

dm/master/server.go
lance6716: the request is not QueryStatus, we use QueryStatus to implement declarative API for other commands, like pause-task (OperateTask)

okJiang: How about directly returning `*pb.CommonWorkerResponse`?

lance6716: that's what caller did in my memory. will check later

dm/master/server.go
Ehco1996: i got it, there is a situation that: ok=false but msg contains some message need to return

Ehco1996 (approved):

dm/master/server.go
Ehco1996: i think it is ok to return `QueryStatusResponse ` here because what `waitOperationOk` does is send `query-status` request to worker

lichunzhu (approved):
LGTM

PR #2065, opened 2021-08-30
tracker: disable background statistic job

lance6716:
reduce CPU consumption

as title, ref https://docs.pingcap.com/zh/tidb/stable/tidb-configuration-file#stats-lease

### Check List 

Tests 

 - No code

Related changes

### DIFF
--- a/pkg/schema/tracker.go
+++ b/pkg/schema/tracker.go
@@ -66,13 +66,11 @@ type Tracker struct {
 func NewTracker(ctx context.Context, task string, sessionCfg map[string]string, tidbConn *conn.BaseConn) (*Tracker, error) {
 	// NOTE: tidb uses a **global** config so can't isolate tracker's config from each other. If that isolation is needed,
 	// we might SetGlobalConfig before every call to tracker, or use some patch like https://github.com/bouk/monkey
-	toSet := tidbConfig.NewConfig()
-	// bypass wait time of https://github.com/pingcap/tidb/pull/20550
-	toSet.TiKVClient.AsyncCommit.SafeWindow = 0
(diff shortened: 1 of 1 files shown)


pkg/schema/tracker.go
```diff
@@ -70,6 +70,8 @@ func NewTracker(ctx context.Context, task string, sessionCfg map[string]string,
 	// bypass wait time of https://github.com/pingcap/tidb/pull/20550
 	toSet.TiKVClient.AsyncCommit.SafeWindow = 0
 	toSet.TiKVClient.AsyncCommit.AllowedClockDrift = 0
+	// set a negative lease will disable the background statistic jobs.
```
Ehco1996, line 73: Maybe add a link to tidb to explain how this change is work

https://github.com/pingcap/tidb/blob/7755d25aba5120dafde98fff11ab3b98ca4d192f/domain/domain.go#L1176

Ehco1996: or we can upgrade `tidb` version to master, it provides a func `SetStatsLease` which can help us do this config

https://github.com/pingcap/tidb/blob/7755d25aba5120dafde98fff11ab3b98ca4d192f/session/tidb.go#L156

Ehco1996 (commented):
restLGTM

pkg/schema/tracker.go
lance6716: The code are changing so we'll give an inactive reference soon.

And I don't the the right time to call `SetStatsLease`

lance6716: changed to function call in https://github.com/pingcap/dm/pull/2065/commits/7ed00726787fe527e07930cb27c4fdde0b0c1771

Ehco1996 (approved):

Issue #2075, opened 2021-09-01
optimizate countBinlogSize


GMHDBJD:
## Feature Request

**Is your feature request related to a problem? Please describe:**

Currently, DM use `show binary logs` to count remain binlog size every 30s per task. When user start multiple tasks in one source, the query may execute frequently in the upstream source and become slow queries

**Describe the feature you'd like:**

count remain binlog size per source or increase the interval time

PR #2076, opened 2021-09-01
worker: rename to source worker and provide source-level status

lance6716:
close https://github.com/pingcap/dm/issues/2075

- rename worker/Worker to worker/SourceWorker
- now SourceWorker get source status, passing them to `Unit.Status(*SourceStatus)` to unit can calculate fields like "catchup with master"
- now after invoking `Unit.Status(*SourceStatus)`, unit will print status log. So if SourceWorker periodically invoke `Unit.Status(*SourceStatus)` the status log can be printed periodically like before

### Check List 

Tests 

 - Unit test

Code changes

Related changes

 - Need to update release note


dm/worker/source_worker.go
```diff
@@ -186,11 +207,54 @@ func (w *Worker) Close() {
...
+
+	w.sourceDB.Close()
+	w.sourceDB = nil
+
 	w.l.Info("Stop worker")
 }
 
+// updateSourceStatus updates w.sourceStatus.
+func (w *SourceWorker) updateSourceStatus(ctx context.Context) error {
+	w.sourceDBMu.Lock()
```
lichunzhu, line 219: This lock doesn't actually work because we directly use `w.sourceDB.DB` in the following part.

loader/loader.go
```diff
@@ -729,12 +730,6 @@ func (l *Loader) Restore(ctx context.Context) error {
 		return err2
 	}
 
-	l.wg.Add(1)
-	go func() {
-		defer l.wg.Done()
-		l.PrintStatus(ctx)
-	}()
-
```
lichunzhu, line 737: I think we can have a goroutine keep printing loader's status to users know loader work well now.

pkg/utils/db.go
```diff
@@ -50,6 +50,20 @@ const (
...
+func init() {
+	failpoint.Inject("PrintStatusCheckSeconds", func(val failpoint.Value) {
+		if seconds, ok := val.(int); ok {
+			PrintTaskInterval = time.Duration(seconds) * time.Second
+			log.L().Info("set printStatusInterval",
+				zap.String("failpoint", "PrintStatusCheckSeconds"),
+				zap.Int("value", seconds))
+		}
+	})
+}
```
lichunzhu, line 65: I think we can set `PrintStatusCheckSeconds` just in `source_worker.go:Start`. Besides, why do we need to set this variable in `utils.go`?

syncer/status.go
```diff
@@ -22,27 +22,50 @@ import (
...
 	st := &pb.SyncStatus{
-		TotalEvents:  total,
-		TotalTps:     totalTps,
-		RecentTps:    tps,
-		SyncerBinlog: syncerLocation.Position.String(),
+		TotalEvents:         total,
+		TotalTps:            totalTps,
+		RecentTps:           tps,
+		SyncerBinlog:        syncerLocation.Position.String(),
+		SecondsBehindMaster: lag,
```
lichunzhu, line 44: It seems that `lag` is only used here
```suggestion
		SecondsBehindMaster: s.secondsBehindMaster.Load(),
```

dm/worker/source_worker.go
lance6716: sql.DB is safe for concurrent use, I only want to protect the assignment to `w.sourceDB`

loader/loader.go
lance6716: Now the SourceWorker will periodically call `Unit.Status`, and `Unit.Status` will print unit's status to log

dm/worker/source_worker.go
lichunzhu: If one routine assign value to w.sourceDB in L219 and another read `w.sourceDB` at L233, we still get a data race.

lichunzhu: Besides, we will unlock `w.sourceDBMu` twice if we successfully apply a new DB.

loader/loader.go
lichunzhu: I see

dm/worker/source_worker.go
lance6716: if the goroutine is going to read `w.sourceDB` at L244 (updated line numbers 🤣) it must pass the nil check in 231 or itself initiated `w.sourceDB`. For the former case, since it has been passed nil check, the assignment is already done. For the latter case, this goroutine finished the assignment and release the lock, so other goroutine which acquired the lock will pass the nil check and not go to line 233 again

lichunzhu: I see. What if we `QueryStatus` and `worker.Close` at the same time? Will L244 cause a data race with L223?

lance6716: currently we can't assure that, will update later

dm/worker/source_worker.go
```diff
@@ -143,13 +163,30 @@ func (w *Worker) Start() {
 			w.l.Info("status print process exits!")
 			return
 		case <-ticker.C:
-			w.l.Debug("runtime status", zap.String("status", w.StatusJSON("")))
+			old := w.sourceStatus.Load()
+			if old != nil {
+				status := old.(*binlog.SourceStatus)
+				if time.Since(status.UpdateTime) < printTaskInterval/2 {
```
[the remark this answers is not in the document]
lance6716: I think there's no need to print status to log too frequently, if we just printed in 15 seconds (may be invoked by query-status) just skip the periodically printing.

dm/worker/source_worker.go
```diff
@@ -186,11 +218,54 @@ func (w *Worker) Close() {
 	}
 
 	w.closed.Store(true)
+
+	w.sourceDB.Close()
```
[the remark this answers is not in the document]
lance6716: `w.closed.Store(true)` is to make sure happened-before relationship (though it failed to protect the concurrent accessing to `w.sourceDB` for now)

dm/worker/source_worker.go
```diff
@@ -122,19 +126,35 @@ func NewWorker(cfg *config.SourceConfig, etcdClient *clientv3.Client, name strin
...
+// Start starts working, but the functionalities should be turned on separately.
+func (w *SourceWorker) Start() {
 	// start task status checker
 	if w.cfg.Checker.CheckEnable {
 		w.taskStatusChecker.Start()
 	}
 
+	var err error
+	w.sourceDB, err = conn.DefaultDBProvider.Apply(w.cfg.DecryptPassword().From)
+	if err != nil {
```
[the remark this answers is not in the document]
lance6716: temporary network problem

loader/status.go
```diff
@@ -37,56 +34,34 @@ func (l *Loader) Status() interface{} {
 		MetaBinlog:     l.metaBinlog.Load(),
 		MetaBinlogGTID: l.metaBinlogGTID.Load(),
 	}
+	go l.printStatus()
```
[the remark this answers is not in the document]
lance6716: just want to reduce blocking time, but seems it's not that slowly 😂 wait for another reviewer

dm/worker/source_worker.go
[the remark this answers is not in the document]
Ehco1996: maybe add comment to ponit out this `invoked by query-status`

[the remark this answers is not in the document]
lance6716: it's not only invoked by `query-status`, many commands will internally call QueryStatus. I think the logging message one line below is enough to explain this behaviour.

dm/worker/source_worker.go
lance6716: I found `SourceWorker.RWMutex` has protected this.

dm/worker/source_worker.go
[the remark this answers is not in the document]
lance6716: above comment is wrong because I missed `SourceWorker.RWMutex`

https://github.com/pingcap/dm/pull/2076/commits/b3670c92d99450a20ee37e826324f9dd6d086ff0

loader/loader.go
Ehco1996: > Now the SourceWorker will periodically call Unit.Status, and Unit.Status will print unit's status to log

this hide in `StatusJSON` ....  it's too hard to find, maybe change `StatusJSON` to `GetUnitAndSourceStatusJSON` or some other easy-for-read name  is better 😂

lance6716: https://github.com/pingcap/dm/pull/2076/commits/7dc8fc3b53064d9fe63b52e05aac60e8287dc670

dm/worker/source_worker.go
lichunzhu: Will this cause a problem?
```
QueryStatus ───► closed(false) ───► utils.GetMasterStatus

worker.Close ─────────────────────► w.sourceDB = nil
```

dm/worker/source_worker.go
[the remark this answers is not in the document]
lichunzhu: What if `w.sourceDB` is nil?

dm/worker/source_worker.go
lance6716: Close will call w.Lock() and QueryStatus will call w.RLock()

dm/worker/source_worker.go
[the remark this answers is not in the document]
lance6716: `(*BaseDB).Close()` will check that

lichunzhu (approved):
LGTM

Ehco1996 (approved):

Issue #2078, opened 2021-09-01
syncer: replace `{schema, table}` with `Table`


okJiang:
## Feature Request

**Is your feature request related to a problem? Please describe:**

Part of https://github.com/pingcap/dm/issues/2064.

- A large number of `{schema, table}` affected code readability. Like https://github.com/pingcap/dm/blob/master/syncer/syncer.go#L2831 and https://github.com/pingcap/dm/blob/master/syncer/syncer.go#L2852.
- Repeated logic like https://github.com/pingcap/dm/blob/2b6279d2b9b1633f3062ab6b819d1f190d5a0ca7/pkg/utils/common.go#L251 and https://github.com/pingcap/tidb-tools/blob/288ab02f1c79faaff0098bfe66d1d9da48607d23/pkg/table-filter/compat.go#L36-L41

**Describe the feature you'd like:**

As title.

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

Issue #2064, opened 2021-08-30
refactor: handleQueryEvent


okJiang:
## Feature Request

**Is your feature request related to a problem? Please describe:**

Many functions in `handleQueryEvent` is mixed. There are lots of question:
- One function contains too many actions but its function name does not contain. For example, 
  - we can see `parseDDLSQL` which can parse ddl from its name, but it also can filter sql and check if it is DML and DDL.
  - `splitAndFilter` also handles online ddl.
- Function contains some features can exclude. For example,
  - `onlineDDL.Apply` returns {realSchema, realTable} only in two situation. Compared to all situations, we can create a new function to implement it.
- Two function have lots of similar logic. For example `skipQuery` and `skipDMLEvent`.
- A large number of parameter income/return in the same time. For example, {schema, table}.

**Describe the feature you'd like:**

- [x] Instead all `{schema string, table string}` of `{Table}`. https://github.com/pingcap/dm/issues/2078
- [x] Delete `GetRealTable` in `OnlinePlugin.Apply`. https://github.com/pingcap/dm/issues/2069
- [x] Refactor `skipQuery/skipDMLEvent`: https://github.com/pingcap/dm/pull/2125 https://github.com/pingcap/dm/pull/2147
  - Extract `filterSQL` from `skipQuery`;
  - Extract `filterOneEvent` from `skipQuery/skipDMLEvent`.
- [x] Refactor `ddl.go`: https://github.com/pingcap/dm/pull/2090 
  - `parseDDLSQL` only parse ddl sql;
  - change `splitAndFilterDDL` to `processSplitedDDL`;
  - change `handleOnlineDDL` to `renameOnlineDDLTable`.
- [x] Add a new struct `queryEventContext` for better income/return parameters. https://github.com/pingcap/dm/pull/2112 
- [x] Make some variable names more meaningful: https://github.com/pingcap/dm/pull/2152

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

okJiang:
I have a new idea about `refactor ddl.go`.

In `onlineDDL`, we can save `ddlInfo` rather than `ddl` in `onlineddl.storage` where https://github.com/pingcap/dm/blob/master/syncer/online-ddl-tools/online_ddl.go#L225-L226.

`ddlInfo` is renamed from `trackDDL` by https://github.com/pingcap/dm/pull/2152 (waiting for review), which represents all used possibly informations about single ddl.

```go
[...]

PR #2079, opened 2021-09-01
syncer: replace `{schema, table}` with `schemapkg.Table` Part 1

okJiang:

Part of https://github.com/pingcap/dm/issues/2078 and https://github.com/pingcap/dm/issues/2064

- Add a alias `Table` for `filter.Table`.
- Replace `{schema, table}` with `Table

### Check List 

Tests 

Code changes


syncer/checkpoint.go
```diff
@@ -32,7 +32,7 @@ import (
 	"github.com/pingcap/dm/pkg/dumpling"
 	"github.com/pingcap/dm/pkg/gtid"
 	"github.com/pingcap/dm/pkg/log"
-	"github.com/pingcap/dm/pkg/schema"
+	schemapkg "github.com/pingcap/dm/pkg/schema"
```
lonng, line 35: Why use package alias?

okJiang: Because we occasionally use `schema` as the variable name.

Issue #2085, opened 2021-09-02
remove mysql dependencies when running unit tests


Ehco1996:
## Feature Request

**Is your feature request related to a problem? Please describe:**

Currently in the DM unit tests we are relying on a real mysql database, which is not really good way in unit test

**Describe the feature you'd like:**

As title，use mock instead of real mysql in unit-test

- [x] pkg: dm
- [x] pkg: dumpling
- [x] pkg: loader
- [x] pkg: relay
- [ ] pkg: syncer
- [x] pkg: others 

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

PR #2096, opened 2021-09-06
test: remove mysql in pkg ut

Ehco1996:
part of #2085 

* pkg others

### Check List 

Tests 

 - Unit test

Code changes

Side effects

 - Increased code complexity

Related changes

### DIFF
--- a/pkg/v1dbschema/schema.go
+++ b/pkg/v1dbschema/schema.go
@@ -152,6 +152,16 @@ func updateSyncerOnlineDDLMeta(tctx *tcontext.Context, dbConn *conn.BaseConn, ta
 
 // getGlobalPos tries to get the global checkpoint position.
 func getGlobalPos(tctx *tcontext.Context, dbConn *conn.BaseConn, tableName, sourceID string) (gmysql.Position, error) {
+	failpoint.Inject("MockGetGlobalPos", func(val failpoint.Value) {
+		posName := val.(string)
+		pos := gmysql.Position{
+			Name: posName,
+			Pos:  uint32(0),
+		}
+		tctx.L().Info("MockGetGlobalPos", zap.String("failpoint", "MockGetGlobalPos"), zap.Any("pos", pos))
+		failpoint.Return(pos, nil)
+	})
+
 	query := fmt.Sprintf(`SELECT binlog_name, binlog_pos FROM %s WHERE id=? AND is_global=? LIMIT 1`, tableName)
 	args := []interface{}{sourceID, true}
 	rows, err := dbConn.QuerySQL(tctx, query, args...)
--- a/pkg/v1dbschema/schema_test.go
+++ b/pkg/v1dbschema/schema_test.go
@@ -24,6 +24,7 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/DATA-DOG/go-sqlmock"
 	gmysql "github.com/go-mysql-org/go-mysql/mysql"
 	. "github.com/pingcap/check"
 	"github.com/pingcap/failpoint"
@@ -46,6 +47,7 @@ type testSchema struct {
 	user     string
 	password string
 	db       *conn.BaseDB
+	mockDB   sqlmock.Sqlmock
 }
 
 var _ = Suite(&testSchema{})
@@ -83,6 +85,7 @@ func (t *testSchema) setUpDBConn(c *C) {
 	cfg.Adjust()
 
 	var err error
+	t.mockDB = conn.InitMockDB(c)
 	t.db, err = conn.DefaultDBProvider.Apply(cfg)
 	c.Assert(err, IsNil)
 }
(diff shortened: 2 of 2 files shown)


pkg/v1dbschema/schema_test.go
```diff
@@ -129,50 +132,70 @@ func (t *testSchema) TestSchemaV106ToV20x(c *C) {
...
 
 	// create v1.0.6 checkpoint table.
 	createCpV106, err := ioutil.ReadFile(filepath.Join(v1DataDir, "v106_syncer_checkpoint-schema.sql"))
 	c.Assert(err, IsNil)
-	_, err = dbConn.ExecuteSQL(tctx, nil, cfg.Name, []string{
-		string(createCpV106),
-	})
+	createCpV106SQL := string(createCpV106)
+	t.mockDB.ExpectBegin()
+	t.mockDB.ExpectExec(createCpV106SQL[:10] + ".*").WithArgs().WillReturnResult(sqlmock.NewErrorResult(nil))
```
Ehco1996, line 147: use this `createCpV106SQL[:10] + ".*"` becasue if i pass `createCpV106SQL` to mockdb, it will return a not matched error 🤔, i don't konw why, need some help here

pkg/v1dbschema/schema_test.go
```diff
@@ -208,7 +252,9 @@ func (t *testSchema) TestSchemaV106ToV20x(c *C) {
 	defer rows.Close()
 	c.Assert(gs.String, Equals, endGS.String())
 
-	rows, err = dbConn.QuerySQL(tctx, fmt.Sprintf(`SELECT binlog_gtid FROM %s WHERE is_global!=1`, dbutil.TableName(cfg.MetaSchema, cputil.SyncerCheckpoint(cfg.Name))))
+	selectSQL = fmt.Sprintf(`SELECT binlog_gtid FROM %s WHERE is_global!=1`, dbutil.TableName(cfg.MetaSchema, cputil.SyncerCheckpoint(cfg.Name)))
+	t.mockDB.ExpectQuery(selectSQL).WithArgs().WillReturnRows(sqlmock.NewRows([]string{"binlog_gtid"}).AddRow(nil))
+	rows, err = dbConn.QuerySQL(tctx, selectSQL)
```
Ehco1996, line 257: need check mock all meet

pkg/v1dbschema/schema_test.go
lance6716: maybe the regular expression engine doesn't support multi-lines matching.

pkg/v1dbschema/schema_test.go
```diff
@@ -129,50 +132,70 @@ func (t *testSchema) TestSchemaV106ToV20x(c *C) {
 	}()
 
 	// create metadata schema.
-	_, err = dbConn.ExecuteSQL(tctx, nil, cfg.Name, []string{
-		`CREATE DATABASE IF NOT EXISTS ` + cfg.MetaSchema,
-	})
+	createSchemaSQL := `CREATE DATABASE IF NOT EXISTS ` + cfg.MetaSchema
```
lance6716, line 135: can be removed because it just tested itself

and many rest of them

lance6716: I think so. the original logic is to do prepare jobs so the `UpdateSchema` can work.

lance6716 (approved):

pkg/v1dbschema/schema.go
```diff
@@ -152,6 +152,16 @@ func updateSyncerOnlineDDLMeta(tctx *tcontext.Context, dbConn *conn.BaseConn, ta
 
 // getGlobalPos tries to get the global checkpoint position.
 func getGlobalPos(tctx *tcontext.Context, dbConn *conn.BaseConn, tableName, sourceID string) (gmysql.Position, error) {
+	failpoint.Inject("MockGetGlobalPos", func(val failpoint.Value) {
```
glorv, line 155: For this kind of function, is it easier and more flexible to depend on sqlmok

Ehco1996: fixed in  https://github.com/pingcap/dm/pull/2096/commits/60f56dbc3048c5407570aeb8e548b38c400053fd 😬

pkg/binlog/reader/util_test.go
```diff
@@ -1,93 +0,0 @@
...
-
-	gmysql "github.com/go-mysql-org/go-mysql/mysql"
-	"github.com/go-mysql-org/go-mysql/replication"
-	. "github.com/pingcap/check"
-
-	"github.com/pingcap/dm/pkg/utils"
-)
-
-// added to testTCPReaderSuite to re-use DB connection.
-func (t *testTCPReaderSuite) TestGetGTIDsForPos(c *C) {
```
Ehco1996, line 27: seems we can not test this func without a connection to mysql  and we actually tested these functions in many integration tests, so I just removed these two test cases

pkg/binlog/reader/tcp_test.go
```diff
@@ -171,55 +95,11 @@ func (t *testTCPReaderSuite) TestSyncPos(c *C) {
 	// already closed
 	err = r.Close()
 	c.Assert(err, NotNil)
-
-	// invalid startup position
-	pos = gmysql.Position{
-		Name: "mysql-bin.888888",
```
Ehco1996, line 177: delete some tests lines beacause we just test wheather `TCPreader` status is  right, some test about  binlog events should be tested in it test

Ehco1996:
@glorv  PTAL😉

glorv (approved):

glorv:
I'm a bit concerned that our current test implement is heavily depending on failpoint, but maybe is better to define the outer depndencies as interface and create a implementation in unit test. So we can fully test all the logic without the restriction of failpoint. We may need to head in this way in the future.

PR #2115, opened 2021-09-09
syncer: fix panic when concurrent call Status

lance6716:
before https://github.com/pingcap/dm/pull/2076, the first print status is happened after a tick (30s), so s.start is very likely initialized. After that PR it's more possible that race happened and further cause a panic.

use atomic type and check zero before division.

### Check List 

Tests 

 - Unit test

Related changes


syncer/status_test.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+func (*mockCheckpoint) FlushedGlobalPoint() binlog.Location {
+	return binlog.Location{
+		Position: mysql.Position{
+			Name: "mysql-bin.000123",
+			Pos: 123,
+		},
+	}
+}
```
Ehco1996, line 64: end line

syncer/status_test.go
```diff
@@ -0,0 +1,64 @@
...
+	wg := sync.WaitGroup{}
+	for i := 0; i < 10; i++ {
+		wg.Add(1)
+		go func() {
+			ret := s.Status(sourceStatus)
+			status := ret.(*pb.SyncStatus)
+			c.Assert(status.MasterBinlog, Equals, "(mysql-bin.000123, 223)")
+			c.Assert(status.SyncerBinlog, Equals, "(mysql-bin.000123, 123)")
+		}()
+	}
```
Ehco1996, line 50: do we need wait wg?

Ehco1996 (approved):

GMHDBJD (approved):
LGTM

PR #2123, opened 2021-09-13
onlineddl: support regular expressions rules for PT/GH-OST

wentaojin:
Solve the problem of custom PT/GH-OST format

This feature is added because some users may modify the pt/gh-ost code because they are worried about the business mistakenly deleting the table, which may cause the format to be inconsistent with the Dm, which affects synchronization，the feature can solve it

Add DM config，if values is null, dm will keep default , if values is not null , match according to specific value

shadow-table-rule: []           # 支持正则，默认 gh-ost(_gho) pt(_new)，匹配多个需要加 (.*)
trash-table-rule: ["_old(.*)"]    # 支持正则, 默认 gh-ost(_ghc,_del) pt(_old)，匹配多个需要加 (.*)


Makefile
```diff
@@ -5,7 +5,7 @@ LDFLAGS += -X "github.com/pingcap/dm/pkg/utils.GitBranch=$(shell git rev-parse -
 LDFLAGS += -X "github.com/pingcap/dm/pkg/utils.GoVersion=$(shell go version)"
 
 CURDIR   := $(shell pwd)
-GO       := GO111MODULE=on go
+GO       := GO111MODULE=on GOOS=linux GOARCH=amd64 go
```
Ehco1996, line 8: i think we should't add `GOOS=linux GOARCH=amd64` here, because many developer also use mac and windows

Makefile
```diff
@@ -84,21 +84,29 @@ tools_setup:
 	@echo "setup tools"
 	@cd tools && make
 
+install_test_python_dep:
+	@echo "install python requirments for test"
```
Ehco1996, line 88: seems not need this cmd🤔

Makefile
```diff
@@ -84,21 +84,29 @@ tools_setup:
...
+	sudo pip install -r tests/requirements.txt
+
 generate_proto: tools_setup
 	./generate-dm.sh
 
 generate_mock: tools_setup
 	./tests/generate-mock.sh
 
+generate_openapi: tools_setup
```
Ehco1996, line 97: ditto

syncer/online-ddl-tools/online_ddl.go
```diff
@@ -488,13 +489,29 @@ func (r *RealOnlinePlugin) Finish(tctx *tcontext.Context, schema, table string)
...
-			return TrashTable
+		trashSuffix := strings.Split(r.storage.cfg.TrashTableRule, ";")
+		for _, trash := range trashSuffix {
+			regex, _ := regexp.Compile(fmt.Sprintf(`%s$`, trash))
+			if regex.MatchString(table) {
+				return TrashTable
+			}
 		}
+
+		//if strings.HasSuffix(table, "_gho") || strings.HasSuffix(table, "_new") {
```
Ehco1996, line 508: please just delete the code you find useless instead of using comments

syncer/online-ddl-tools/online_ddl.go
```diff
@@ -488,13 +489,29 @@ func (r *RealOnlinePlugin) Finish(tctx *tcontext.Context, schema, table string)
...
+			if regex.MatchString(table) {
+				return GhostTable
+			}
 		}
 
-		if strings.HasSuffix(table, "_ghc") || strings.HasSuffix(table, "_del") || strings.HasSuffix(table, "_old") {
-			return TrashTable
+		trashSuffix := strings.Split(r.storage.cfg.TrashTableRule, ";")
+		for _, trash := range trashSuffix {
+			regex, _ := regexp.Compile(fmt.Sprintf(`%s$`, trash))
```
Ehco1996, line 502: we can add a `regexp.Compile` cache in struct `RealOnlinePlugin ` for better performance

Ehco1996 (commented):
Is there a description document for the customer's problem? I think it would help to review PR

dm/config/subtask.go
```diff
@@ -137,6 +137,10 @@ type SubTaskConfig struct {
 	// deprecated
 	OnlineDDLScheme string `toml:"online-ddl-scheme" json:"online-ddl-scheme"`
 
+	// pt/gh-ost name rule,support regex
+	ShadowTableRule string `yaml:"shadow-table-rule" toml:"shadow-table-rule" json:"shadow-table-rule"`
```
Ehco1996, line 141: maybe not export `ShadowTableRule` and `TrashTableRule`  to user is bettrer for now 

we can tell if these temporary tables are from `pt` or `ghc`>

Makefile
wentaojin: yes，because build binary testing, so i modify

Makefile
wentaojin: em... strange, I didn’t change this part

wentaojin:
> Is there a description document for the customer's problem? I think it would help to review PR

This feature is added because some users may modify the pt/gh-ost code because they are worried about the business mistakenly deleting the table, which may cause the format to be inconsistent with the Dm, which affects synchronization，the feature can solve it

webank customer

dm/config/subtask.go
```diff
@@ -274,6 +278,14 @@ func (c *SubTaskConfig) Adjust(verifyDecryptPassword bool) error {
 		return terror.ErrConfigOnlineSchemeNotSupport.Generate(c.OnlineDDLScheme)
 	}
 
+	if len(c.ShadowTableRule) == 0 {
+		c.ShadowTableRule = []string{"_gho", "_new"}
```
glorv, line 282:
```suggestion
		c.ShadowTableRule = []string{"^_(.+)_(?:new|gho)$"}
```
The pattern should contains exact one match group so we can use something like  `pattern.FindStringSubmatch` to extra the original table name from the pattern matching

dm/config/subtask.go
```diff
@@ -274,6 +278,14 @@ func (c *SubTaskConfig) Adjust(verifyDecryptPassword bool) error {
...
+	}
+
+	if len(c.TrashTableRule) == 0 {
+		c.TrashTableRule = []string{"_ghc", "_del", "_old"}
```
glorv, line 286:
```suggestion
		c.TrashTableRule = []string{"^_(.+)_(?:ghc|del|old)$"}
```

syncer/online-ddl-tools/online_ddl.go
```diff
@@ -488,11 +493,11 @@ func (r *RealOnlinePlugin) Finish(tctx *tcontext.Context, schema, table string)
 func (r *RealOnlinePlugin) TableType(table string) TableType {
 	// 5 is _ _gho/ghc/del or _ _old/new
 	if len(table) > 5 && strings.HasPrefix(table, "_") {
-		if strings.HasSuffix(table, "_gho") || strings.HasSuffix(table, "_new") {
+		if r.shadowReg.MatchString(table) {
```
glorv, line 496: Remove the useless `if len(table) > 5 && strings.HasPrefix(table, "_")`, we should only depend on the pattern matching result here

syncer/online-ddl-tools/online_ddl.go
```diff
@@ -488,11 +493,11 @@ func (r *RealOnlinePlugin) Finish(tctx *tcontext.Context, schema, table string)
...
 			return GhostTable
 		}
 
-		if strings.HasSuffix(table, "_ghc") || strings.HasSuffix(table, "_del") || strings.HasSuffix(table, "_old") {
+		if r.trashReg.MatchString(table) {
```
glorv, line 500: Please rewrite the `RealName` function with something like:
```
// RealName implements interface.
func (r *RealOnlinePlugin) RealName(table string) string {
	if res := r.shadowReg.FindStringSubmatch(table) > 1 {
		return res[1]
	} else if res := r.trashReg.FindStringSubmatch(table) > 1 {
		return res[1]
	}
	return table
}
```

syncer/syncer.go
```diff
@@ -333,7 +334,17 @@ func (s *Syncer) Init(ctx context.Context) (err error) {
 	}
 
 	if s.cfg.OnlineDDL {
-		s.onlineDDL, err = onlineddl.NewRealOnlinePlugin(tctx, s.cfg)
+		shadowReg, err := regexp.Compile(fmt.Sprintf(`%s%s`,
```
glorv, line 337: I think we need to support multiple expressions here to extract the right real table name

dm/config/subtask.go
wentaojin: ^_(.+)_(?:new|gho)$ , The user configuration feels not very friendly, Similar configuration,Meet three requirements at the same time，unless user set ["new","gho(.*)"], Background processing into ^_(.+)_(?:new|gho)$ 

there are three situations here:
1. Default, such as: _old
2. User-specified format, such as: _old_9999
3. User fuzzy matching format, such as: _old_9999(.*)

wentaojin: i think  ["_new","_gho","_old(.*)"] is friendly

dm/config/subtask.go
```diff
@@ -274,6 +278,14 @@ func (c *SubTaskConfig) Adjust(verifyDecryptPassword bool) error {
 		return terror.ErrConfigOnlineSchemeNotSupport.Generate(c.OnlineDDLScheme)
 	}
 
+	if len(c.ShadowTableRule) == 0 {
```
glorv, line 281: I think we should also add this adjust in `TaskConfig.adjust`. By the way, plase add following check to ensure that each regexp pattern contains exactly one submatch:
```
for _, r := range c.ShadowTableRule {
  p, err := regexp.Compile(r)
  if err != nil {
    ...
  }
  if p.NumSubexp() != 1 {
    return errors.New(...)
  }
}
```

PR #2139, opened 2021-09-15
onlineddl: support regular expressions rules for PT/GH-OST 

wentaojin:
## What problem does this PR solve?
Solve the problem of custom PT/GH-OST format

This feature is added because some users may modify the pt/gh-ost code because they are worried about the business mistakenly deleting the table, which may cause the format to be inconsistent with the Dm, which affects synchronization，the feature can solve it

close https://github.com/pingcap/dm/pull/2123

## What is changed and how it works?
Add DM config，if values is null, dm will keep default , if values is not null , match according to specific value

shadow-table-rule: []    # 支持正则，默认 `^_(.+)_(?:new|gho)$` 匹配多个具体值后加 `(.*)`
trash-table-rule: [] # 支持正则, 默认 `^_(.+)_(?:ghc|del|old)$` 匹配多个具体值后加 `(.*)`


syncer/online-ddl-tools/online_ddl.go
```diff
@@ -500,9 +506,18 @@ func (r *RealOnlinePlugin) TableType(table string) TableType {
...
+		if len(shadowRes) > 1 {
+			return shadowRes[1]
+		}
+	}
+
+	for _, trashReg := range r.trashRegs {
+		trashRes := trashReg.FindStringSubmatch(table)
+		if len(trashRes) > 1 {
+			return trashRes[1]
+		}
```
lichunzhu, line 520: Is that possible for us to add some unit tests for this part?

dm/config/subtask.go
```diff
@@ -338,6 +344,54 @@ func (c *SubTaskConfig) Adjust(verifyDecryptPassword bool) error {
...
+			if !strings.HasSuffix(r, "$") {
+				r = r + "$"
+			}
+
+			p, err := regexp.Compile(r)
+			if err != nil {
+				return err
+			}
+			if p.NumSubexp() != 1 {
+				return terror.Annotate(err, "'shadow-table-rule' regex pattern isn't contains exactly one submatch")
```
lichunzhu, line 365: `err` is actually `nil` here. We should add a new `terror` and print the `ShadowTableRule` here.
Example:
```
				return terror.ErrConfigOnlineDDLInvalidRegex.Generate("'shadow-table-rule", r)
```
Defenition in [terror](https://github.com/pingcap/dm/blob/master/pkg/terror/error_list.go):
```
	ErrConfigOnlineDDLInvalidRegex   = New(codeConfigOnlineDDLInvalidRegex, ClassConfig, ScopeInternal, LevelHigh, "''%s' regex pattern %s isn't contains exactly one submatch")
```

dm/config/subtask.go
```diff
@@ -338,6 +344,54 @@ func (c *SubTaskConfig) Adjust(verifyDecryptPassword bool) error {
...
+				r = "^" + r
+			}
+
+			if !strings.HasSuffix(r, "$") {
+				r = r + "$"
+			}
+
+			p, err := regexp.Compile(r)
+			if err != nil {
+				return err
```
lichunzhu, line 362: Use a terror to annotate this

dm/config/subtask.go
```diff
@@ -338,6 +344,54 @@ func (c *SubTaskConfig) Adjust(verifyDecryptPassword bool) error {
 		log.L().Warn("'online-ddl-scheme' will be deprecated soon. Recommend that use online-ddl instead of online-ddl-scheme.")
 	}
 
+	if len(c.ShadowTableRule) == 0 {
+		c.ShadowTableRule = []string{"^_(.+)_(?:new|gho)$"}
+	} else {
+		var shadowTableRule []string
```
lichunzhu, line 350:
```suggestion
		shadowTableRule := make([]string, 0, len(c.ShadowTableRule))
```

dm/config/subtask.go
```diff
@@ -338,6 +344,54 @@ func (c *SubTaskConfig) Adjust(verifyDecryptPassword bool) error {
...
+			if !strings.HasPrefix(r, "^") {
+				r = "^" + r
+			}
+
+			if !strings.HasSuffix(r, "$") {
+				r = r + "$"
+			}
+			p, err := regexp.Compile(r)
+			if err != nil {
+				return err
```
lichunzhu, line 386: ditto

dm/config/subtask.go
```diff
@@ -338,6 +344,54 @@ func (c *SubTaskConfig) Adjust(verifyDecryptPassword bool) error {
...
+			}
+			if p.NumSubexp() != 1 {
+				return terror.Annotate(err, "'trash-table-rule' regex pattern isn't contains exactly one submatch")
```
lichunzhu, line 389: ditto

syncer/syncer.go
```diff
@@ -346,7 +347,25 @@ func (s *Syncer) Init(ctx context.Context) (err error) {
...
+			}
+			shadowRegs = append(shadowRegs, shadowReg)
+		}
+		for _, tg := range s.cfg.TrashTableRule {
+			trashReg, err := regexp.Compile(tg)
+			if err != nil {
+				return err
+			}
+			trashRegs = append(trashRegs, trashReg)
+		}
```
lichunzhu, line 367: How about putting these code in `onlineddl.NewRealOnlinePlugin`?

Ehco1996:
Please fix the integration [test error](https://ci.pingcap.net/blue/organizations/jenkins/dm_ghpr_new_test/detail/dm_ghpr_new_test/2493/pipeline) on CI as well 🤣

This is because you have changed the config of the task and can be fixed by referring to this [commit](https://github.com/pingcap/dm/pull/2055/commits/6e87cb299dacac7f03f77722c9f035aba21adfde)

Note that in your case you should not use diff -i to ignore the inconsistency, but rather set `shadow-table-rule ` and `trash-table-rule ` in task.yaml and check that they are yi'vi after export

syncer/online-ddl-tools/online_ddl.go
wentaojin: local has already build binary test

syncer/syncer.go
wentaojin: why putting in `onlineddl.NewRealOnlinePlugin` , one compile performance isn't good ?

dm/config/subtask.go
```diff
@@ -192,6 +193,11 @@ type SubTaskConfig struct {
 	IsSharding bool   `toml:"is-sharding" json:"is-sharding"`
 	ShardMode  string `toml:"shard-mode" json:"shard-mode"`
 	OnlineDDL  bool   `toml:"online-ddl" json:"online-ddl"`
+
+	// pt/gh-ost name rule,support regex
+	ShadowTableRule []string `yaml:"shadow-table-rule" toml:"shadow-table-rule" json:"shadow-table-rule"`
+	TrashTableRule  []string `yaml:"trash-table-rule" toml:"trash-table-rule" json:"trash-table-rule"`
```
glorv, line 199:
```suggestion
	ShadowTableRules []string `yaml:"shadow-table-rules" toml:"shadow-table-rules" json:"shadow-table-rules"`
	TrashTableRules  []string `yaml:"trash-table-rules" toml:"trash-table-rules" json:"trash-table-rules"`
```

dm/config/task.go
```diff
@@ -302,6 +302,10 @@ type TaskConfig struct {
 	MySQLInstances []*MySQLInstance `yaml:"mysql-instances" toml:"mysql-instances" json:"mysql-instances"`
 
 	OnlineDDL bool `yaml:"online-ddl" toml:"online-ddl" json:"online-ddl"`
+	// pt/gh-ost name rule,support regex
+	ShadowTableRule []string `yaml:"shadow-table-rule" toml:"shadow-table-rule" json:"shadow-table-rule"`
```
glorv, line 306: ditto

pkg/terror/error_list.go
```diff
@@ -877,6 +879,10 @@ var (
 	ErrConfigGenTableRouter         = New(codeConfigGenTableRouter, ClassConfig, ScopeInternal, LevelHigh, "generate table router error", "Please check the `routes` config in task configuration file.")
 	ErrConfigGenColumnMapping       = New(codeConfigGenColumnMapping, ClassConfig, ScopeInternal, LevelHigh, "generate column mapping error", "Please check the `column-mappings` config in task configuration file.")
 	ErrConfigInvalidChunkFileSize   = New(codeConfigInvalidChunkFileSize, ClassConfig, ScopeInternal, LevelHigh, "invalid `chunk-filesize` %v", "Please check the `chunk-filesize` config in task configuration file.")
+	ErrConfigOnlineDDLRegexCompile  = New(codeConfigOnlineDDLRegexCompile, ClassConfig, ScopeInternal, LevelHigh,
```
glorv, line 882: I think we can merge this two error code together, the only reason that regexp compile fail is that the syntax is not correct.

dm/config/subtask.go
```diff
@@ -343,6 +372,26 @@ func (c *SubTaskConfig) Adjust(verifyDecryptPassword bool) error {
 		log.L().Warn("'online-ddl-scheme' will be deprecated soon. Recommend that use online-ddl instead of online-ddl-scheme.")
 	}
 
+	if len(c.ShadowTableRules) == 0 {
+		c.ShadowTableRules = []string{"^_(.+)_(?:new|gho)$"}
```
Ehco1996, line 376: maybe change `^_(.+)_(?:new|gho)$` and `"^_(.+)_(?:ghc|del|old)$"` to constans

dm/config/subtask.go
```diff
@@ -317,6 +323,29 @@ func (c *SubTaskConfig) Decode(data string, verifyDecryptPassword bool) error {
 	return c.Adjust(verifyDecryptPassword)
 }
 
+func adjustOnlineTableRules(ruleType string, rules []string) ([]string, error) {
+	adjustedRules := make([]string, 0, len(rules))
+	for _, r := range rules {
+		if !strings.HasPrefix(r, "^") {
+			r = "^" + r
```
Ehco1996, line 333: perhaps it would be better to just return the error here rather than adjust re ((add "^" and "$") ) for the user

syncer/syncer_test.go
```diff
@@ -141,14 +141,16 @@ func (s *testSyncerSuite) SetUpSuite(c *C) {
...
-		Name:         "syncer_ut",
-		Mode:         config.ModeIncrement,
-		Flavor:       "mysql",
-		LoaderConfig: loaderCfg,
+		From:             config.GetDBConfigFromEnv(),
+		To:               config.GetDBConfigFromEnv(),
+		ServerID:         101,
+		MetaSchema:       "test",
+		Name:             "syncer_ut",
+		ShadowTableRules: []string{"^_(.+)_(?:new|gho)$"},
```
Ehco1996: no need to manually add these two rules here, subtask.Adjust already do this

Ehco1996 (commented):
rest LGTM

syncer/syncer_test.go
lichunzhu: We didn't adjust the whole subtask config here.

dm/config/subtask.go
lichunzhu: I think this way is simpler for users. They need to check the error message and rewrite task.yaml if we return an error here. Waiting for other reviewer' suggestions...

lichunzhu:
/run-all-tests
https://ci.pingcap.net/blue/organizations/jenkins/dm_ghpr_new_test/detail/dm_ghpr_new_test/2632/pipeline

glorv (approved):

Ehco1996 (approved):

dm/config/subtask.go
okJiang: It is ok if noted in document.

PR #2090, opened 2021-09-03
syncer: refine filter.go and ddl.go

okJiang:
### What problem does this PR solve? 

Part of https://github.com/pingcap/dm/issues/2064

### What is changed and how it works?

- refine `filter.go`
	- modify `parseDDLSQL`, `splitAndFilterDDL`, `handleOnlineDDL`;
	- rename `splitAndFilterDDL` to `preprocessDDL`;
- refine `ddl.go`
	- rename `skipQuery` to `filterQueryEvent` and `skipRowsEvent` to `filterRowsEvent`;
	- remove `filterSQL` from `filterQueryEvent`;
	- wrap a new function `filterOneEvent` from both `filterQueryEvent` and `filterRowsEvent`.
- add a new struct `queryEventContext` to wrap all variables handleQueryEvent used.

```go
// May be modified in the future
type queryEventContext struct {
	eventContext

	p         *parser.Parser // used parser
	ddlSchema string         // used schema
	originSQL string         // before split
	// split multi-schema change DDL into multiple one schema change DDL due to TiDB's limitation
	splitedDDLs []string // after split before online ddl
	appliedDDLs []string // after onlineDDL apply if onlineDDL != nil and track, before route
	routedDDLs  []string // after route

	ddlInfo             *shardingDDLInfo
	needHandleDDLs      []string
	needTrackDDLs       []trackedDDL
	sourceTbls          map[string]map[string]struct{} // db name -> tb name
	onlineDDLTableNames map[string]*filter.Table
}
```

### Check List 

Tests 

 - Unit test

PR #2147, opened 2021-09-17
syncer: refactor `filter.go`

okJiang:

part of https://github.com/pingcap/dm/issues/2064

- Extract `filterSQL` from `skipQuery`;
- Extract `filterOneEvent` from `skipQuery/skipDMLEvent`;
- Update UT.

### Check List 

Tests 

 - Unit test


syncer/filter.go
```diff
@@ -112,13 +55,50 @@ func (s *Syncer) skipDMLEvent(table *filter.Table, eventType replication.EventTy
 	case replication.DELETE_ROWS_EVENTv0, replication.DELETE_ROWS_EVENTv1, replication.DELETE_ROWS_EVENTv2:
 		et = bf.DeleteEvent
 	default:
-		return false, terror.ErrSyncerUnitInvalidReplicaEvent.Generate(eventType)
+		filtered, err := s.filterOneEvent(table, bf.NullEvent, "")
+		if err != nil {
+			return false, err
```
okJiang, line 60: Maybe should return `terror.ErrSyncerUnitInvalidReplicaEvent.Generate(eventType)`?

syncer/filter.go
```diff
@@ -112,13 +55,50 @@ func (s *Syncer) skipDMLEvent(table *filter.Table, eventType replication.EventTy
...
+	tables := s.baList.Apply([]*filter.Table{table})
+	if len(tables) == 0 {
+		return true, nil
+	}
+	if s.binlogFilter == nil {
+		return false, nil
+	}
+	action, err := s.binlogFilter.Filter(table.Schema, table.Name, et, sql)
+	if err != nil {
+		return false, terror.Annotatef(terror.ErrSyncerUnitBinlogEventFilter.New(err.Error()), "skip event %s on `%s`.`%s`", et, table.Schema, table.Name)
```
okJiang, line 101:
```suggestion
		return false, terror.Annotatef(terror.ErrSyncerUnitBinlogEventFilter.New(err.Error()), "skip event %s on %v", et, table)
```

syncer/filter_test.go
```diff
@@ -91,41 +71,131 @@ func (s *testFilterSuite) TestSkipQueryEvent(c *C) {
...
+			replication.UPDATE_ROWS_EVENTv0,
+			false,
+		}, {
+			&filter.Table{Schema: "foo", Name: "test"},
+			replication.DELETE_ROWS_EVENTv0,
 			false,
 		},
+	}
+	for _, ca := range cases {
+
```
okJiang, line 136:
```suggestion
```

syncer/filter_test.go
```diff
@@ -91,41 +71,131 @@ func (s *testFilterSuite) TestSkipQueryEvent(c *C) {
...
 		{
-			"create table foo.bar(id int)",
-			[]*filter.Table{{Schema: "foo", Name: "bar"}},
+			// system table
+			"create table mysql.test (id int)",
+			&filter.Table{Schema: "mysql", Name: "test"},
+			"",
+			true,
+		}, {
+
```
okJiang, line 177:
```suggestion
```

syncer/ddl.go
```diff
@@ -89,7 +89,7 @@ func (s *Syncer) parseDDLSQL(sql string, p *parser.Parser, schema string) (resul
 			if len(table.Schema) == 0 {
 				table.Schema = schema
 			}
-			ignore, err2 := s.skipDMLEvent(table, replication.QUERY_EVENT)
+			ignore, err2 := s.filterRowsEvent(table, replication.QUERY_EVENT)
```
Ehco1996, line 92: i think we should remove `case ast.DMLNode` from this func, beacuse we can make sure all passed in sql is DDL in row based binlog

cc @lance6716 @glorv

syncer/filter.go
```diff
@@ -24,85 +24,28 @@ import (
...
 )
 
-// skipQuery will return true when
-// - given `sql` matches builtin pattern.
-// - any schema of table names is system schema.
-// - any table name doesn't pass block-allow list.
-// - type of SQL doesn't pass binlog filter.
-func (s *Syncer) skipQuery(tables []*filter.Table, stmt ast.StmtNode, sql string) (bool, error) {
+func (s *Syncer) filterQueryEvent(tables []*filter.Table, stmt ast.StmtNode, sql string) (bool, error) {
 	if utils.IsBuildInSkipDDL(sql) {
```
Ehco1996, line 28: is this check duplicated with `filterSQL`, can we remove one  ?

syncer/filter.go
```diff
@@ -112,13 +55,50 @@ func (s *Syncer) skipDMLEvent(table *filter.Table, eventType replication.EventTy
...
+}
 
+// filterOneEvent will return true when
+// - given `sql` matches builtin pattern.
+// - any schema of table names is system schema.
+// - any table name doesn't pass block-allow list.
+// - type of SQL doesn't pass binlog filter.
+// - pattern of SQL doesn't pass binlog filter.
+func (s *Syncer) filterOneEvent(table *filter.Table, et bf.EventType, sql string) (bool, error) {
+	if filter.IsSystemSchema(table.Schema) {
```
Ehco1996, line 89: seems not contains `matches builtin pattern` check?

syncer/filter.go
okJiang: Because `filterSQL` and `filterQueryEvent` are not serial execution after the https://github.com/pingcap/dm/pull/2090. `filterSQL` will only be called when SQL parse failed. So I think needn't remove it.

syncer/filter.go
okJiang: What is builtin pattern check?

syncer/ddl.go
okJiang: About the change of `ddl.go`, I suggest discuss it in https://github.com/pingcap/dm/pull/2090. It had a huge change for `ddl.go`. You can read the draft changes and give me some suggestions for better.

lance6716: for now user may ignore precheck about ROW format binlog

syncer/filter.go
```diff
@@ -112,13 +55,50 @@ func (s *Syncer) skipDMLEvent(table *filter.Table, eventType replication.EventTy
...
+		} else {
+			return false, terror.ErrSyncerUnitInvalidReplicaEvent.Generate(eventType)
+		}
 	}
+	return s.filterOneEvent(table, et, "")
+}
 
-	action, err := s.binlogFilter.Filter(schemaName, tableName, et, "")
+// filterSQL filter unsupported sql in tidb and global sql-patterns in binlog-filter config file.
+func (s *Syncer) filterSQL(sql string) (bool, error) {
```
lance6716, line 71: maybe `filterSQLByPattern`

syncer/ddl.go
```diff
@@ -38,7 +38,7 @@ type parseDDLResult struct {
 
 func (s *Syncer) parseDDLSQL(sql string, p *parser.Parser, schema string) (result parseDDLResult, err error) {
 	// check skip before parse (used to skip some un-supported DDLs)
-	ignore, err := s.skipQuery(nil, nil, sql)
+	ignore, err := s.filterSQL(sql)
```
lance6716, line 41: unify `ignore`, `shouldSkip`, `filtered`

okJiang: `skip` used in so many place, so we unified use `skip`, and change function name `filter*` to `skip*`.

syncer/filter.go
Ehco1996: in this func's comment 
```
// skipOneEvent will return true when
// - given `sql` matches builtin pattern.
```

okJiang: > seems not contains matches builtin pattern check?

you are right.

> // - given `sql` matches builtin pattern.

this represents `utils.IsBuildInSkipDDL`. I will delete this comment next commit.

syncer/filter_test.go
```diff
@@ -91,41 +71,129 @@ func (s *testFilterSuite) TestSkipQueryEvent(c *C) {
...
+			"create table tx.test (id int)",
+			&filter.Table{Schema: "tx", Name: "test"},
+			bf.CreateTable,
+			false,
+		}, {
+			// test balist
+			"create table s1.test (id int)",
+			&filter.Table{Schema: "s1", Name: "test"},
+			bf.CreateTable,
+			false, // there should be true, so what does BAList configure?
```
okJiang, line 191: PTAL

okJiang, line 191: PTAL @lance6716

lance6716: maybe baList is not initialized from config

syncer/filter.go
Ehco1996: `needSkip, err := s.skipOneEvent(table, bf.NullEvent, "")` why we need check NullEvent here? maybe just return `terror.ErrSyncerUnitInvalidReplicaEvent.Generate(eventType)?` is enouth

okJiang: Because if the `table` can be skipped by `baList`, we needn't return error.

But it is indeed not good by call `skipOneEvent`. Let me modify it.

I also think return `terror.ErrSyncerUnitInvalidReplicaEvent.Generate(eventType)` is right.

Ehco1996 (approved):
LGTM

okJiang:
cc @lance6716 @glorv

syncer/filter_test.go
```diff
@@ -50,39 +52,17 @@ func (s *testFilterSuite) TearDownSuite(c *C) {
...
-			Action:        bf.Ignore,
-		}, {
 			SchemaPattern: "foo*",
 			TablePattern:  "",
 			Events:        []bf.EventType{bf.CreateTable},
 			SQLPattern:    []string{"^create\\s+table"},
-			Action:        bf.Do,
-		}, {
-			SchemaPattern: "foo*",
-			TablePattern:  "bar*",
```
lance6716, line 79: please keep the test about of "finer granularity has higher priority"

syncer/filter_test.go
```diff
@@ -91,41 +71,172 @@ func (s *testFilterSuite) TestSkipQueryEvent(c *C) {
...
+		},
+	}
+	syncer := NewSyncer(cfg, nil)
+	var err error
+	syncer.baList, err = filter.New(syncer.cfg.CaseSensitive, syncer.cfg.BAList)
+	c.Assert(err, IsNil)
+	// test binlog filter
+	filterRules := []*bf.BinlogEventRule{
 		{
-			"rename table foo.test to foo.test1",
```
lance6716, line 112: test a RENAME SQL because it has more than one table name

lance6716 (commented):
rest lgtm

syncer/filter.go
```diff
@@ -24,85 +24,25 @@ import (
...
 )
 
-// skipQuery will return true when
-// - given `sql` matches builtin pattern.
-// - any schema of table names is system schema.
-// - any table name doesn't pass block-allow list.
-// - type of SQL doesn't pass binlog filter.
-func (s *Syncer) skipQuery(tables []*filter.Table, stmt ast.StmtNode, sql string) (bool, error) {
+func (s *Syncer) skipQueryEvent(tables []*filter.Table, stmt ast.StmtNode, sql string) (bool, error) {
 	if utils.IsBuildInSkipDDL(sql) {
```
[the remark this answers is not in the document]
okJiang: There is a reply https://github.com/pingcap/dm/pull/2147/files#r711425725.

syncer/filter_test.go
okJiang: I agree. `RENAME` SQL has more than one table name, but L113 only has one table name...

lance6716: the logic is RENAME will be skipped when any table is skipped (`if len(tbs) != len(tables)`)

okJiang: I know the logic. I mean the input `tables` is from `FetchDDLTables`. If SQL is `RENAME`, the input `Tables` will contain [old1, new1], but the test case in L113 is only [old1]. So the former test maybe have a problem.

syncer/filter_test.go
okJiang: fixed in https://github.com/pingcap/dm/pull/2147/commits/7b53729eea240ff8b9cb6c35385fec57e9ab15ef

syncer/filter_test.go
```diff
@@ -50,82 +52,212 @@ func (s *testFilterSuite) TearDownSuite(c *C) {
...
+			true,
+		}, {
+			// test filter one event
+			"drop table foo.test",
+			[]*filter.Table{{Schema: "foo", Name: "test"}},
+			false,
+		}, {
+			"create table foo.test (id int)",
+			[]*filter.Table{{Schema: "foo", Name: "test"}},
+			true,
```
lance6716, line 87: maybe you can test RENAME in this case

okJiang: fixed in https://github.com/pingcap/dm/pull/2147/commits/4191f43ccd43e0027a5aab0d0bae64bc8ea7789e

lance6716 (approved):

PR #2153, opened 2021-09-18
test: fix integration test which may cause case fail sometimes

D3Hunter:
integration test: fix invalid pgrep expression which may cause case fail sometime; add sleep to all_mode test(may fail sometimes)

some integration test case(all_mode) may fail sometimes(when process doesn't exit fast enough or delayed)

- fix `pgrep` expression in `wait_pattern_exit`
- add `sleep 2` to integration test case `all_mode`

### Check List 

Tests 
 - No code

### DIFF
--- a/tests/_utils/test_prepare
+++ b/tests/_utils/test_prepare
@@ -32,7 +32,7 @@ function wait_pattern_exit() {
     pattern=$1
     while true
     do
-        if [ "pgrep -f $pattern" != "0" ]; then
+        if ! pgrep -f $pattern >/dev/null 2>&1; then
             echo "pattern $pattern already exit"
             return 0
         fi
--- a/tests/all_mode/run.sh
+++ b/tests/all_mode/run.sh
@@ -400,6 +400,7 @@ function run() {
 		"resume-relay -s mysql-replica-01" \
 		"\"result\": true" 2
 
+	sleep 2
 	# relay should continue pulling from syncer's checkpoint, so only pull the latest binlog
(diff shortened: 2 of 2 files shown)


okJiang:
Thanks for your first contribution. 

Does this pr resolve the unstable test case `all_mode` in https://github.com/pingcap/dm/issues/2073?

If so, you can link them😊.

D3Hunter:
> Hi, thanks for your first contribution.
> 
> Does this pr resolve the unstable test case `all_mode` in #2073?
> 
> If so, you can link them😊.

[jenkins 2161](https://ci.pingcap.net/blue/organizations/jenkins/dm_ghpr_new_test/detail/dm_ghpr_new_test/2161/pipeline/) is the case i met for adding `sleep 2`.

lance6716 (approved):

tests/_utils/test_prepare
```diff
@@ -32,7 +32,7 @@ function wait_pattern_exit() {
     pattern=$1
     while true
     do
-        if [ "pgrep -f $pattern" != "0" ]; then
+        if ! pgrep -f $pattern >/dev/null 2>&1; then
```
glorv, line 35: Could you also fix the typo in L36 `echo "pattern $pattern already exist"`

glorv (approved):

tests/_utils/test_prepare
D3Hunter: seems not a typo, it should be `process with pattern $pattern already exit`

PR #2158, opened 2021-09-23
*: update tools and fix lint

Ehco1996:

as title

* update golangci-lint and many other tools version
* remove `golint` from golangci-lint config, because this linter is deprecated in new version

`WARN [runner] The linter 'golint' is deprecated (since v1.41.0) due to: The repository of the linter has been archived by the owner.  Replaced by revive.`
* fix some code style for new lint

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

Related changes


.golangci.yml
```diff
@@ -11,7 +11,6 @@ linters:
     - exportloopref
     - gocritic
     - godot
-    - golint
```
GMHDBJD, line 14: Do we need to disable the whole golint linter？

GMHDBJD, line 14: LGTM. Maybe still remain it, though it's deprecated.

GMHDBJD (approved):

lichunzhu (approved):
LGTM

.golangci.yml
Ehco1996: i think no need to keep `golint`, becasue it's check have been replaced by `revive` (we already have this linter)

Issue #2165, opened 2021-09-24
feature: pause/resume/stop (sub)tasks related to a upstream source in batch mode


D3Hunter:
## Feature Request

**Is your feature request related to a problem? Please describe:**
customer may shutdown/restart MySQL for maintenance and want to pause/resume all related dm (sub)tasks.

although it can be done by using `query-status` to get a list of tasks related to the source and `pause-task`/`resume-task` one by one, or write a script to do the job, it would be better have a `dmctl` command do the job as it happens often in production env.

in this scenario there is no need to pause/resume the whole task(one task may have multiple subtasks, each one correspond to one upstream source), only subtasks related to the source

**Describe the feature you'd like:**
- change existing `pause-task`/`resume-task`/`stop-task` command so that it can pause/resume/stop (sub)tasks related to a upstream source in batch mode

**Describe alternatives you've considered:**
use `query-status` to get a list of tasks related to the source and `pause-task`/`resume-task` one by one, or write a script to do the job

**Teachability, Documentation, Adoption, Migration Strategy:**

PR #2166, opened 2021-09-24
[#2165] dmctl: change pause-task/resume-task/stop-task to support operate tasks bound to a source.

D3Hunter:
see the title and #2165

- change `pause-task`/`resume-task`/`stop-task` to support operate tasks bound to a source. it's enabled when `task-name/task-conf` is not specified and only on `-s source` is passed. it's a shorthand for the following steps:
  - run `query-status` with the `source-name` to get a list of subtasks bound to it
  - run operation in batch mode
- batch size can be adjusted with `--batch-size` flag, default 5

### Check List 

Tests 
- Unit test
- Integration test

Related changes


dm/ctl/master/batch_task.go
```diff
@@ -0,0 +1,174 @@
+// Copyright 2019 PingCAP, Inc.
```
lance6716, line 1:
```suggestion
// Copyright 2021 PingCAP, Inc.
```

dm/ctl/master/batch_task.go
```diff
@@ -0,0 +1,174 @@
...
+	// control workload to dm-cluster for sources with large number of tasks.
+	cmd.PersistentFlags().Int(batchSizeFlag, defaultBatchSize, "batch size")
+
+	cmd.AddCommand(
+		&cobra.Command{
+			Use:   "pause <source-name>",
+			Short: "pause subtasks bound to an upstream MySQL/MariaDB source in batch mode",
+			RunE:  batchPauseTaskFunc,
+		},
+		&cobra.Command{
```
lance6716, line 62: maybe we should also provide batch stop. A paused task still forbids its source to be removed

lichunzhu (commented):
I don't think we should add two commands for this feature. Why not directly support batch on original `resume-task` and `pause-task`?

D3Hunter:
> I don't think we should add two commands for this feature. Why not directly support batch on original `resume-task` and `pause-task`?

those commands asks user to input the specific task to operate, it's just a wrapper for the `dm-master` rpc call, although we can remove this restriction, but i think we should keep it that way.

`batch-task` is more like a syntax sugar to accomplish a task happens often, to save labor of writing script to orchestrate those `"atomic"` operation.

dm/ctl/master/batch_task.go
D3Hunter: which scenario needs to stop tasks in batch mode?

lichunzhu:
> > I don't think we should add two commands for this feature. Why not directly support batch on original `resume-task` and `pause-task`?
> 
> those commands asks user to input the specific task to operate, it's just a wrapper for the `dm-master` rpc call, although we can remove this restriction, but i think we should keep it that way.
> 
> `batch-task` is more like a syntax sugar to accomplish a task happens often, to save labor of writing script to orchestrate those `"atomic"` operation.

I think `pause-task` and `resume-task`'s syntax don't conflict with `batch-task`. Besides, I think adding this `batch-task` will confuse the users who don't know dm well. To me, if two commands are doing nearly the same things, we don't have to add a new one.
What's more, here is a command example from `kubectl` of `kubernetes`:
```
~ kubectl get --help
  ...
  # Return only the phase value of the specified pod.
  kubectl get -o template pod/web-pod-13je7 --template={{.status.phase}}

  # List all replication controllers and services together in ps output format.
  kubectl get rc,services
```

dm/ctl/master/batch_task.go
lance6716: a prerequisite of `operate-source stop` is no task is using the source.

GMHDBJD:
I think new command should be disscuss with PM now. Maybe we can add a flag for current command? e.g.
```pause-task -s source --all`

**Comment by D3Hunter** [comment]
> > > I don't think we should add two commands for this feature. Why not directly support batch on original `resume-task` and `pause-task`?
> > 
> > 
> > those commands asks user to input the specific task to operate, it's just a wrapper for the `dm-master` rpc call, although we can remove this restriction, but i think we should keep it that way.
> > `batch-task` is more like a syntax sugar to accomplish a task happens often, to save labor of writing script to orchestrate those `"atomic"` operation.
> 
> I think `pause-task` and `resume-task`'s syntax don't conflict with `batch-task`. Besides, I think adding this `batch-task` will confuse the users who don't know dm well. To me, if two commands are doing nearly the same things, we don't have to add a new one.
> What's more, here is a command example from `kubectl` of `kubernetes`:
> 
> ```
> ~ kubectl get --help
>   ...
>   # Return only the phase value of the specified pod.
>   kubectl get -o template pod/web-pod-13je7 --template={{.status.phase}}
> 
>   # List all replication controllers and services together in ps output format.
>   kubectl get rc,services
> ```

it's a point

File: dm/ctl/master/operate_task.go [hunk 3]
```diff
@@ -0,0 +1,194 @@
...
+		return nil
+	}
+
+	result := batchOperateTask(taskOp, batchSize, sources, resp.Sources[0].SubTaskStatus)
+	common.PrettyPrintInterface(result)
+
+	return nil
+}
+
+func batchOperateTask(taskOp pb.TaskOp, batchSize int, sources []string, subTaskStatus []*pb.SubTaskStatus) *batchTaskResult {
```
Ehco1996, line 117: maybe move this func to `dm/master/server.go` and make this func to public, so that other package  such as `openapi` could reuse this batch-operate-task logic?

Ehco1996: or we can implement this `batch-op` in `dm-master` and add a new rpc  instead of calling `dm-master-server` muti times ?

Ehco1996 (commented):
maybe not use `git push -f` to update your code, so other reviewers could see  code change history, all your commit will  be squashed there is no need to worry about multiple commits affecting the commit history

dm/ctl/master/operate_task.go
D3Hunter: `batch-op` is composite task which can be done with multiple `atomic` api calls(`query-status` then `xxx-task`), i think we should not add such an api unless it's very necessary(such as performance bottleneck), or else the number of api would explode

D3Hunter:
> maybe not use `git push -f` to update your code, so other reviewers could see code change history, all your commit will be squashed there is no need to worry about multiple commits affecting the commit history

i squash it locally

D3Hunter:
@lichunzhu @GMHDBJD  updated, have a look.

dm/ctl/master/operate_task.go
```diff
@@ -0,0 +1,184 @@
...
+		return err
+	}
+
+	common.PrettyPrintResponse(resp)
+	return nil
+}
+
+func addOperateSourceTaskFlags(cmd *cobra.Command) {
+	// control workload to dm-cluster for sources with large number of tasks.
+	cmd.Flags().Int(batchSizeFlag, defaultBatchSize, "batch size")
```
glorv, line 79: Shall we remove this parameter by automatically providing a proper value?🤔

dm/ctl/master/operate_task.go
```diff
@@ -0,0 +1,184 @@
...
+	common.PrettyPrintInterface(result)
+
+	return nil
+}
+
+func batchOperateTask(taskOp pb.TaskOp, batchSize int, sources []string, subTaskStatus []*pb.SubTaskStatus) *batchTaskResult {
+	result := batchTaskResult{Result: true, BatchResults: []*operateTaskResult{}}
+
+	if len(subTaskStatus) < batchSize {
+		batchSize = len(subTaskStatus)
```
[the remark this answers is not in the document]
D3Hunter: when number of tasks need to operate < batch size, there is no need to start that match go routines

dm/ctl/master/operate_task.go
D3Hunter: > Shall we remove this parameter by automatically providing a proper value?🤔
now it's a simple rule, default to 5, but can be adjusted manually
the max number of tasks related to a single source ever met on production env is 10+, so it seems enough for most cases.
maybe we can enhance it later?

lichunzhu, line 79: We'd better explain this parameter is batch size of xxx to users.

dm/ctl/master/operate_task.go
```diff
@@ -0,0 +1,184 @@
...
+			defer wg.Done()
+
+			for name := range workCh {
+				taskResult := operateTaskResult{TaskName: name, Op: taskOp.String()}
+				taskOpResp, err := common.OperateTask(taskOp, name, sources)
+				if err != nil {
+					taskResult.Result = false
+					taskResult.Msg = err.Error()
+				} else {
+					taskResult.Result = true
```
lichunzhu, line 141:
```suggestion
					taskResult.Result = taskOpResp.Result
```

dm/ctl/master/operate_task.go
```diff
@@ -0,0 +1,184 @@
...
+
+const (
+	batchSizeFlag    = "batch-size"
+	defaultBatchSize = 5
+)
+
+type batchTaskResult struct {
+	Result       bool                 `json:"result"`
+	Msg          string               `json:"msg"`
+	BatchResults []*operateTaskResult `json:"batchResults"`
```
lichunzhu, line 38:
```suggestion
	Tasks []*operateTaskResult `json:"tasks"`
```
It's similar to `sources`. `batchResults` is a little confusing to me.

dm/ctl/master/operate_task.go
```diff
@@ -0,0 +1,184 @@
...
+	if err != nil {
+		return "", 0, false
+	}
+	if len(sources) == 0 {
+		common.PrintLinesf(`must give one source-name when task-name/task-conf is not specified`)
+		return "", 0, false
+	} else if len(sources) > 1 {
+		common.PrintLinesf(`can give only one source-name when task-name/task-conf is not specified`)
+		return "", 0, false
+	}
```
lichunzhu, line 177: Is is possible for us to support `len(sources) > 1` in the future?

D3Hunter: maybe when we have some feedback from users, it's quite enough for now.

PS: with multiple `-s` parameter, the semantic is a little confusing, it can be:
- tasks with all sources
- tasks with either source

tests/dmctl_basic/run.sh
```diff
@@ -178,6 +178,41 @@ function run() {
 		"\"result\": true" 2 \
 		'"worker": "worker1"' 1
 
+	echo "############################################################"
+	echo "test case for operate tasks bound to a source"
+	echo "############################################################"
+	run_dm_ctl $WORK_DIR "127.0.0.1:$MASTER_PORT" "start-task $cur/conf/foo-task.yaml" "\"result\": true" 3
+	run_dm_ctl $WORK_DIR "127.0.0.1:$MASTER_PORT" "start-task $cur/conf/bar-task.yaml" "\"result\": true" 3
```
okJiang, line 185: Can you add a `source3` to `bar-task.yaml` not `foo-task.yaml`? And test `pause-task/resume-task/stop-task -s source3`

okJiang (dismissed):

dm/ctl/master/operate_task.go
```diff
@@ -0,0 +1,184 @@
...
+)
+
+type batchTaskResult struct {
+	Result bool                 `json:"result"`
+	Msg    string               `json:"msg"`
+	Tasks  []*operateTaskResult `json:"tasks"`
+}
+
+type operateTaskResult struct {
+	TaskName string                     `json:"taskName"`
```
Ehco1996, line 42:
```suggestion
	TaskName string                     `json:"task-name"`
```
This is more in line with the current usage of codebase

D3Hunter: renamed to task to follow current convention

dm/ctl/master/operate_task.go
```diff
@@ -0,0 +1,184 @@
...
+	}
+
+	sort.Slice(result.Tasks, func(i, j int) bool {
+		return result.Tasks[i].Task < result.Tasks[j].Task
+	})
+
+	return &result
+}
+
+func parseOperateSourceTaskParams(cmd *cobra.Command) (string, int, bool) {
```
glorv, line 166: I think is better to return a error instead of a bool so the caller can further wrap the error detail if needed

glorv (commented):
rest LGTM

lichunzhu (approved):
LGTM

tests/dmctl_basic/conf/foo-task.yaml
```diff
@@ -0,0 +1,28 @@
...
+  - source-id: "mysql-replica-01"
+    block-allow-list:  "instance"
+    route-rules: ["route-rule-1"]
+  - source-id: "mysql-replica-02"
+    block-allow-list:  "instance"
+    route-rules: ["route-rule-1"]
+
+block-allow-list:
+  instance:
+    do-dbs: ["dmctl_command_dummy"]
```
okJiang, line 28: ditto

tests/dmctl_basic/conf/foo-task.yaml
```diff
@@ -0,0 +1,28 @@
...
+    target-schema: "test"
+    target-table: "foo"
+
+mysql-instances:
+  - source-id: "mysql-replica-01"
+    block-allow-list:  "instance"
+    route-rules: ["route-rule-1"]
+  - source-id: "mysql-replica-02"
+    block-allow-list:  "instance"
+    route-rules: ["route-rule-1"]
```
okJiang, line 24: ditto

tests/dmctl_basic/conf/bar-task.yaml
```diff
@@ -0,0 +1,28 @@
...
+  port: 4000
+  user: "root"
+  password: ""
+
+routes:
+  route-rule-1:
+    schema-pattern: "test"
+    table-pattern: "bar*"
+    target-schema: "test"
+    target-table: "bar"
```
okJiang, line 16: Does it work?

tests/dmctl_basic/conf/bar-task.yaml
```diff
@@ -0,0 +1,28 @@
...
+
+mysql-instances:
+  - source-id: "mysql-replica-01"
+    block-allow-list:  "instance"
+    route-rules: ["route-rule-1"]
```
okJiang, line 21: ditto

tests/dmctl_basic/conf/bar-task.yaml
```diff
@@ -0,0 +1,28 @@
...
+  - source-id: "mysql-replica-02"
+    block-allow-list:  "instance"
+    route-rules: ["route-rule-1"]
```
okJiang, line 24: ditto

tests/dmctl_basic/conf/baz-task.yaml
```diff
@@ -0,0 +1,25 @@
...
+  port: 4000
+  user: "root"
+  password: ""
+
+routes:
+  route-rule-1:
+    schema-pattern: "test"
+    table-pattern: "baz*"
+    target-schema: "test"
+    target-table: "baz"
```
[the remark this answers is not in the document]
D3Hunter: just for test whether pause/resume/stop works as expected, just a dummy task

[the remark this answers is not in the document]
okJiang: Ok, I mean your task just pause/resume/stop, don't execute sql to load and syncer, so that you don't use the `route` module and `black-white list` as follow.

If that, maybe you can remove it.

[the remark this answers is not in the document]
D3Hunter: all integration test uses same mysql instances, `black-white list` is used to isolate cases.

`route` can be removed though.

PR #2182, opened 2021-09-28
onlineddl: report an error when online ddl only matches only one regex

lichunzhu:
When we mistakenly configure `trash-table-rules` in online ddl, downstream real table might be renamed to `fake trash table`.

Give a warning to users when they mistakenly configure only `trash-table-rules` or `shadow-table-rules` and don't execute any ddl to downstream at this time.

### Check List 

Tests 

 - Unit test
 - Integration test

Side effects

 - Increased code complexity

Related changes


syncer/util.go
```diff
@@ -118,3 +119,20 @@ func printServerVersion(tctx *tcontext.Context, db *conn.BaseDB, scope string) {
...
 }
+
+const (
+	shadowTableMatch int = iota
+	trashTableMatch
+	allTableMatch
+)
+
+func unmatchedOnlineDDLRules(match int) string {
+	switch match ^ 1 {
```
Ehco1996, line 130:
```suggestion
	switch match  {
	case shadowTableMatch
	case trashTableMatch
	case allTableMatch
	}
```
It's a little more readable that way

pkg/terror/error_list.go
```diff
@@ -880,6 +881,8 @@ var (
 	ErrConfigInvalidChunkFileSize   = New(codeConfigInvalidChunkFileSize, ClassConfig, ScopeInternal, LevelHigh, "invalid `chunk-filesize` %v", "Please check the `chunk-filesize` config in task configuration file.")
 	ErrConfigOnlineDDLInvalidRegex  = New(codeConfigOnlineDDLInvalidRegex, ClassConfig, ScopeInternal, LevelHigh,
 		"config '%s' regex pattern '%s' invalid, reason: %s", "Please check if params is correctly in the configuration file.")
+	ErrConfigOnlineDDLMistakeRegex = New(codeConfigOnlineDDLMistakeRegex, ClassConfig, ScopeInternal, LevelHigh,
+		"online ddl sql '%s' invalid, table %s fail to match '%s' online ddl regex", "Please update your shadow-table-rules or trash-table-rules in the configuration file.")
```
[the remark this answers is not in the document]
lichunzhu: Table name will be printed as `%s`.`%s` format here, so I delete quotation here.

syncer/ddl.go
```diff
@@ -138,6 +151,22 @@ func (s *Syncer) splitAndFilterDDL(
 
 		// get real tableNames before apply block-allow list
 		if s.onlineDDL != nil {
+			if onlineDDLFinish && len(tables) > 1 {
+				// record trash/shadow table to give users information to check regex
+				tableRecords[i^1] = tables[i^1].String()
```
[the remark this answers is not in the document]
lichunzhu: Must use i^1 here.
                            tables[0]       tables[1]
DDL 0: rename real table -> trash table
DDL 1: rename trash table -> real table

syncer/util.go
lichunzhu: Then `shadowTableMatch` should return `config.TrashTableRules`. I think it's more complicated...

syncer/ddl.go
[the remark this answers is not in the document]
lichunzhu: `tableRecords[i^1]` is because `trashTableMatch==1`.

syncer/ddl.go
```diff
@@ -138,6 +154,24 @@ func (s *Syncer) splitAndFilterDDL(
...
+			if onlineDDLFinish && len(tables) > 1 {
+				if i == 0 && s.onlineDDL.TableType(tables[0].Name) == onlineddl.RealTable &&
+					s.onlineDDL.TableType(tables[1].Name) == onlineddl.TrashTable {
+					onlineDDLMatched = trashTable
+					// record trash table to give users information to check regex
+					tableRecords[trashTable] = tables[1].String()
+				}
+				if i == 1 && s.onlineDDL.TableType(tables[0].Name) == onlineddl.GhostTable &&
+					s.onlineDDL.TableType(tables[1].Name) == onlineddl.RealTable {
+					if onlineDDLMatched == allTable {
```
Ehco1996, line 166: why do we need to check `if onlineDDLMatched == allTable`?

seems if it matches `i == 1 && s.onlineDDL.TableType(tables[0].Name) == onlineddl.GhostTable && s.onlineDDL.TableType(tables[1].Name) == onlineddl.RealTable` it is enough to determine that the re of the onlineDDL is a successful match.

syncer/ddl.go
```diff
@@ -165,6 +199,11 @@ func (s *Syncer) splitAndFilterDDL(
 
 		statements = append(statements, ss...)
 	}
+	if onlineDDLFinish {
+		if unmatchedOnlineDDLType := unmatchedOnlineDDLRules(onlineDDLMatched); len(unmatchedOnlineDDLType) > 0 {
```
Ehco1996, line 203: seems check `onlineDDLMatched == allTable ` is enough ?

syncer/ddl.go
lichunzhu: Because `onlineDDLMatched` only has value if we only match `1` online ddl regex.
If we match both online ddl regex or none ddl regex, we both return `allTable` and a `nil` error.

syncer/ddl.go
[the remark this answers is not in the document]
okJiang: > Must use i^1 here. The rename DDLs are in the following part.
> 
> ```
>             tables[0]       tables[1]
> 
> DDL 0      real table ───► trash table
> 
> DDL 1     shadow table ───►  real table
> ```
> 
> In DDL 0, we need to store trash table's name. In DDL 1, we need to store shadow table's name.

I suggest add this to comment.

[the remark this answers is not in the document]
lichunzhu: Already add in https://github.com/pingcap/dm/pull/2182/commits/e30eff846f696ac53f11eea10ff4ae234431d25a and write this part in a more human-readable way.

_utils/terror_gen/errors_release.txt
```diff
@@ -178,6 +178,7 @@ ErrConfigGenTableRouter,[code=20045:class=config:scope=internal:level=high], "Me
 ErrConfigGenColumnMapping,[code=20046:class=config:scope=internal:level=high], "Message: generate column mapping error, Workaround: Please check the `column-mappings` config in task configuration file."
 ErrConfigInvalidChunkFileSize,[code=20047:class=config:scope=internal:level=high], "Message: invalid `chunk-filesize` %v, Workaround: Please check the `chunk-filesize` config in task configuration file."
 ErrConfigOnlineDDLInvalidRegex,[code=20048:class=config:scope=internal:level=high], "Message: config '%s' regex pattern '%s' invalid, reason: %s, Workaround: Please check if params is correctly in the configuration file."
+ErrConfigOnlineDDLMistakeRegex,[code=20049:class=config:scope=internal:level=high], "Message: online ddl sql '%s' invalid, table %s fail to match '%s' online ddl regex, Workaround: Please update your shadow-table-rules or trash-table-rules in the configuration file."
```
GMHDBJD, line 181: ErrConfigOnlineDDLMistakeRegex,[code=20049:class=config:scope=internal:level=high], "Message: online ddl sql '%s' invalid, table %s fail to match '%s' online ddl regex, Workaround: Please check the `shadow-table-rules` or `trash-table-rules` config item in the configuration file."

syncer/ddl.go
```diff
@@ -138,6 +154,29 @@ func (s *Syncer) splitAndFilterDDL(
 
 		// get real tableNames before apply block-allow list
 		if s.onlineDDL != nil {
+			if onlineDDLFinish && len(tables) > 1 {
```
D3Hunter, line 157: can we abstract this part of logic into a named function, too much indent levels already, hard to read

GMHDBJD, line 157:
```suggestion
			if onlineDDLFinish && len(tables) = 2 {
```

D3Hunter (changes requested):

syncer/ddl.go
```diff
@@ -138,6 +154,29 @@ func (s *Syncer) splitAndFilterDDL(
...
+						onlineDDLMatched = trashTable
+					}
+				} else if i == 1 {
+					// record shadow table to give users information to check regex
+					tableRecords[shadowTable] = tables[0].String()
+					if s.onlineDDL.TableType(tables[0].Name) == onlineddl.GhostTable &&
+						s.onlineDDL.TableType(tables[1].Name) == onlineddl.RealTable {
+						// if no trash table is not matched before, we should record that shadow table is matched here
+						// if shadow table is matched before, we just return all tables are matched and a nil error
+						if onlineDDLMatched == allTable {
```
Ehco1996, line 172:
```suggestion
						if onlineDDLMatched !=  trashTable {
```

GMHDBJD (approved):
rest LGTM

syncer/ddl.go
glorv, line 157: I suggest move this local into online ddl package with a parameter of `[]*ddlInfo`. We can still first parseStatement and FetchTableNames, but then wrap the result into a `[]ddlInfo`. And the do the online ddl logic and finally other remaining logics.

okJiang: Maybe we can remove `len(tables) > 1`, because after `FetchDDLTables` `ast.RenameTableStmt`, len(tables) must equal 2 which test in https://github.com/pingcap/dm/blob/master/pkg/parser/common_test.go#L136-L142.

Ehco1996 (commented):
rest LGTM

glorv (approved):

Ehco1996 (approved):
LGTM

PR #2186, opened 2021-09-29
*: move from deprecated ioutil to os and io packages

kallydev:

Since dm is currently using Go 1.16, so it is safe to remove deprecated `ioutil` package, ref https://golang.org/doc/go1.16#ioutil.

Remove deprecated `ioutil` package.

### Check List 

Tests 

 - Unit test

Code changes
 
Side effects

Related changes


kallydev:
Hello @lance6716, do I need to change the title of the commit from `style` to `*`?

lance6716:
> Hello @lance6716, do I need to change the title of the commit from `style` to `*`?

Hi, thanks for your contribution. 

the commit will be squashed when PR merging, so the commit title is not a problem.

According to [CONTRIBUTING.md](https://github.com/pingcap/dm/blob/master/CONTRIBUTING.md#format-of-the-commit-message), we write component names like `syncer`, `worker` instead of `style`, so I made that change to the PR title. That's not a big problem, if you insist it you can change the title back 😄

lance6716 (approved):

Ehco1996 (approved):

PR #2196, opened 2021-10-08
RFC: DML Execution Optimization

GMHDBJD:
add RFC doc for DML Execution Optimization


GMHDBJD:
> does "synchronous execution of DMLs" mean "will start receiving the next batch of compacted jobs only after all of them have finished executing previous jobs", and asynchronous means current implementation?

Yes

GMHDBJD:
> maybe if the last job in buffer ?
if first jobs is safe-mode but last job is not, we still use safemode for the first job.

PR #2217, opened 2021-10-13
support explicitly set time_zone and change default time_zone to target db tz

glorv:
- Change the default `time_zone` setting from `UTC` to the global time_zone setting from target db.
- Add back the `time_zone` in TaskConfig because this is useful in some special cases.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

Related changes


lance6716:
replaced by https://github.com/pingcap/ticdc/pull/3403

Issue #2204, opened 2021-10-11
mysql source worker %s has already started with source %s, but get a request with source %s


GMHDBJD:
### What did you do?

upgrade from v2.0.6 to v2.0.7
query-status

### What did you expect to see?

task runs normal

### What did you see instead?

```
error: mysql source worker %s has already started with source %s, but get a request with source %s
workaround: Please try restart this DM-worker
```

### Versions of the cluster

DM version (run `dmctl -V` or `dm-worker -V` or `dm-master -V`):
```console
v2.0.7
```

lance6716:
as we discussed, since a "relay bound" worker should be treat as a free worker, I plan to add "relay" WorkerStage to (offline, free, bound). WorkerStage is not persisted so there should be no compatibility problem.

PR #2219, opened 2021-10-13
scheduler: add Relay as a worker stage and fix bug

lance6716:
close https://github.com/pingcap/dm/issues/2204

Now scheduler of DM master will maintain DM worker in 4 stage: Offline, Free, Relay, Bound. Relay is newly added to easily and correctly handle some stage transition.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

 - Increased code complexity
 - Breaking backward compatibility (but fix a bug)

Related changes

 - Need to update the documentation AND METRICS


lance6716:
> Can we add more comments to descript our behavior to make it easier for our collaborators to read? For example (just suggestion):
> 
> ```
> If worker fails, it fallbacks to other workers without relay. Last worker goes online and dispatches back to open relay sync.
> worker hangs with priority fallback other relay worker. If no other relay worker is found source will fallback idle worker, and with relay sync. If last worker goes online and schedules back with relay sync.
> ```

I'll add some comment of choose bound to `tryBoundForWorker` and `tryBoundForSource`. Currently we don't implement "Last worker goes online and dispatches back to open relay sync" in my memory.

PR #2234, opened 2021-10-18
sync: refactor dml.go

GMHDBJD:
refactor dml.go for later develop
- use `INSERT ON DUPLICATE KEY UPDATE` instead of `REPLACE` in safemode
- generate DML SQL string before execute rather than before addJob.

### Check List 

Tests 

 - orginal test
 - unit test

### DIFF
--- a/dm/master/openapi.go
+++ b/dm/master/openapi.go
@@ -100,6 +100,26 @@ func (s *Server) GetDocHTML(ctx echo.Context) error {
 	return ctx.HTML(http.StatusOK, html)
 }
 
+// DMAPIGetClusterMasterList get cluster master node list url is:(GET /api/v1/cluster/masters).
+func (s *Server) DMAPIGetClusterMasterList(ctx echo.Context) error {
+	return nil
+}
+
+// DMAPIOfflineMasterNode offline master node url is: (DELETE /api/v1/cluster/masters/{master-name}).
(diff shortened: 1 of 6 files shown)


lance6716:
I forget why we choose INSERT ON DUPLICATE KEY UPDATE? seems INSERT ON DUPLICATE KEY UPDATE and REPLACE can both use BatchGet when have multiple values

image
test they two :arrow_up:

GMHDBJD:
> I forget why we choose INSERT ON DUPLICATE KEY UPDATE? seems INSERT ON DUPLICATE KEY UPDATE and REPLACE can both use BatchGet when have multiple values

I check the TiDB code, `REPLACE` use get and `INSERT ON DUPLICATE KEY UPDATE` use batch get.
https://github.com/pingcap/tidb/blob/94e30df8e2d8ba2a1a26f153f40067ba3acd78eb/executor/replace.go#L130
https://github.com/pingcap/tidb/blob/94e30df8e2d8ba2a1a26f153f40067ba3acd78eb/executor/insert.go#L72

Another reason is that if update statement doesn't update index value, change it to `REPLACE` statement will delete and insert it, which will update index. While change it to INSERT ON DUPLICATE KEY UPDATE will update the value, which will not  update index.

Furthermore, if we are in safemode, `REPLACE` on same value will still delete and insert the row again, but `INSERT ON DUPLICATE UPDATE` will not update the row at all.

I tested it before and it was in line with this conclusion

Issue #2241, opened 2021-10-19
make `start-relay` easier to use


lance6716:
### Is your feature request related to a problem? Please describe

1. I want to start relay to reduce the number of upstream binlog replication slaves, not care about which worker started relay so don't want to specify worker name. Currently I must specify worker name of the source is bound or free 
2. I want to specify the position from which the relay is pulling. Currently this is specified in source configuration file, which is hard to modify

### Describe the feature you'd like

1. `start-relay -s <source-id> --on-bound`
2. `start-relay -s <source-id> <worker-name> --binlog-file <filename> --binlog-pos <position>` or `start-relay -s <source-id> <worker-name> --binlog-gtid <gtid set>`

lance6716:
PTAL @sunzy

lance6716:
and for https://github.com/pingcap/dm/issues/2205, I'm requesting

3. `start-relay --all --on-bound`, `stop-relay --all --on-bound`

lance6716:
updated the issue description

PR #2249, opened 2021-10-21
*: add etcd key for `start-relay` without worker name

lance6716:
part of #2241. The rest work will be done in another PR after #2219 is merged to reduce conflict.

as title, this key is just a marker like `enable-relay` in source config. But we'll dynamically change the value, and we should not modify the original input source config, so I persist it in etcd.

### Check List 

Tests 

 - Unit test

Code changes

 - Has persistent data change

Side effects

 - Breaking backward compatibility (this etcd key can not be recgonized when downgrade, but I think that is expected)

Related changes


lance6716:
replaced by https://github.com/pingcap/ticdc/pull/3190

PR #2270, opened 2021-10-27
relay/streamer: move binlog log reader into relay

D3Hunter:
it's first part of subtask 4 in #2214

no code logic change in this pr, to make review/merge easier 

- move binlog log reader into relay
- flat small sub packages
- rename some structs to avoid conflict

### Check List 

no code logic change


D3Hunter:
closed, move to ticdc

PR #2274, opened 2021-10-28
upgrade, scheduler: recover memory status of relay when upgrade from <2.0.2

lance6716:

need https://github.com/pingcap/dm/pull/2249 first

recover start-relay memory status when upgrading from v2.0.1, so user will not bother manually do it. the functionality will be implemented in next pr 

see above 

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes


lance6716:
changed implementation

